PR #412, opened 2019-06-27
Owasp dependency check

WMeijbeek:


fniessink:
Unfortunately, the OWASP dependency check is too slow to run on CircleCI (over 10 minutes) and also not feasible to run on Travis (since Travis doesn't save build artifacts). Closing this PR without merging.

PR #455, opened 2019-07-12
added phpldapadmin config to docker-compose.yml

Hedde:
Added ldap ui tooling for development, can be inspected at localhost:3890


Hedde:
Environment variables inside the docker-compose file where defined in two formats, picked one for consistency.

PR #606, opened 2019-09-10
To prevent reporting Checkmarx internal server errors to users when r…

### LINKED ISSUE #468: Checkmarx collector fails with Checkmark internal server error (500)
Issue #593 has additional info.

fniessink:
…eports are unexpectedly unavailable, don't immediately remove a Checkmarx report after reading it, but silently ignore a removed report and create a new one. Fixes #468.


fniessink:
@milenk, @dicksnel. I think this fix will work better: if ignores the 500 error that occurs when the report is no longer available on the Checkmarx server and forgets the id of that report. That way Quality-time will request a new report for the latest scan on the next run. This way we don't create a new report every 15 minutes, but only when really necessary.

PR #736, opened 2019-10-24
Bug 734: LDAP login problem solved for servers where user bind must b…

milenk:
…e done.


milenk:
something wrong with this PR! i'll create a new one.

PR #1454, opened 2020-09-10
Don't verify ssl for metric sources

ds-o:
Skip verification of SSL certificates for metric sources. This allows the usage of metric sources using HTTPS with self-signed certificates.


fniessink:
Closing this PR in favor of #1455.

PR #1542, opened 2020-10-08
Improve Docker instructions

Gamer1120:
It wasn't initially clear that there was a docker subfolder, so I got an error when building. It was quickly resolved, but it's better to just add it in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,11 @@ Clone this repository:
 git clone https://github.com/ICTU/quality-time.git
 ```
 
+Move into the 'quality-time/docker/' directory:
+```console
+cd quality-time/docker/
+```
+
 Build the containers:
 
 ```console


fniessink:
Are you on Windows? Because on macOs (and Linux I think, but I haven't tested) you don't need to change folders. The .env file sets the COMPOSE_FILE environment variable. We haven't found a way to specify the paths platform independently though, which would be my preferred solution. And thanks for the PR!

Gamer1120:
I'm on Windows 10, I executed the commands in PowerShell as admin. I can provide screenshots if necessary.

fniessink:
Confirmation you are on Windows was all I needed, thanks! I will do a bit of research to see if the .env file can be made platform independent, if not I'll merge your PR.

fniessink:
Adding `COMPOSE_PATH_SEPARATOR=":"` as first line of the .env file in the project root folder should fix the issue. Can you please confirm?

fniessink:
Closing this PR in favor of #1544. Thanks!

PR #1576, opened 2020-10-15
Add link to the report that generated the notification

Cas-Ros:


fniessink (changes requested):
Please use the `notification["report_name"]` in line 29 as anchor for the URL so no separate line is needed for the link to the report. 

Also, the `notification["report_name"]` isn't actually the report name, it's the report UUID. Maybe you can fix this as well?

components/notifier/src/strategies/reds_per_report.py
```diff
@@ -8,7 +8,7 @@ def reds_per_report(json) -> List:
     red_metrics = []
     for report in json["reports"]:
         red_metrics.append({
-            "report_name": report["report_uuid"],
+            "report_uuid": report["report_uuid"],
```
fniessink, line 11:
```suggestion
            "report_uuid": report["report_uuid"],
            "report_title": report["title"],
```

components/notifier/src/quality_time_notifier.py
```diff
@@ -26,11 +26,11 @@ async def notify(log_level: int = None) -> None:
         notifications = reds_per_report(json)
 
         for notification in notifications:
-            text = "\n\r" + notification["report_name"] + " contains " + str(notification["red_metrics"]) + \
-                   " red metrics" + "\n\r" + notification["url"]
+            text = "\n\r[" + notification["report_uuid"] + "](" + notification["url"] + \
```
fniessink:
```suggestion
            text = "\n\r[" + notification["report_title"] + "](" + notification["url"] + \
```

fniessink (changes requested):
You're using the report uuid as anchor for the URL, but I think the report title would be more useful. See suggested changes.

fniessink (approved):

PR #1582, opened 2020-10-16
Iks304 notify when metric changes to red

Cas-Ros:
also removed the old notification strategy


components/notifier/src/quality_time_notifier.py
```diff
@@ -23,18 +23,32 @@ async def notify(log_level: int = None) -> None:
...
+                    logging.warning("unable to send the notification for %s", notification["report_uuid"])
+        else:
+            logging.info("no new red metrics")
 
         logging.info("Sleeping %.1f seconds...", sleep_duration)
         await asyncio.sleep(sleep_duration)
 
 
+def build_notification_text(text_parameters):
+    """create and format the contents of the notification"""
```
fniessink:
```suggestion
    """Create and format the contents of the notification."""
```

components/notifier/src/quality_time_notifier.py
```diff
@@ -23,18 +23,32 @@ async def notify(log_level: int = None) -> None:
             response = await session.get(f"http://{server_host}:{server_port}/api/v3/reports")
             json = await response.json()
 
-        notifications = reds_per_report(json)
+        notifications = reds_that_are_new(json)
 
+        total_new_red_metrics = 0
         for notification in notifications:
```
fniessink:
```suggestion
        total_new_red_metrics = sum(notification["new_red_metrics"] for notification in notifications)
```
And line 30 can be removed.

components/notifier/src/quality_time_notifier.py
```diff
@@ -23,18 +23,32 @@ async def notify(log_level: int = None) -> None:
...
-            text = "\n\r[" + notification["report_title"] + "](" + notification["url"] + \
-                   ") contains " + str(notification["red_metrics"]) + " red metrics"
-            send = send_notification_to_teams(notification["teams_webhook"], text)
-            if not send:
-                logging.warning("unable to send the notification for %s", notification["report_uuid"])
+            total_new_red_metrics += notification["new_red_metrics"]
+        if total_new_red_metrics > 0:
+            for notification in notifications:
+                text = build_notification_text(notifications)
```
fniessink: Why are you passing all notifications to create the notification for a specific report? If report A has a web hook it means it also receives the red metrics for report B, right? That doesn't seem correct.

components/notifier/src/strategies/reds_that_are_new.py
```diff
@@ -0,0 +1,28 @@
...
+                if metric["status"] == "target_not_met" and used_to_be_not_red(metric):
+                    new_red_metrics[-1]["new_red_metrics"] += 1
+
+    return new_red_metrics
+
+
+def used_to_be_not_red(metric):
+    """determine if a metric used to be something other than red"""
+    if "recent_measurements" in metric:
+        return metric["recent_measurements"][0]["count"]["status"] != "target_not_met"
```
fniessink: This isn't correct: you're looking at the first recent measurement, which is the oldest recent measurement. You have to look at the last recent measurement, i.e. `metric["recent_measurements"][-1]` and check whether it is red **and** has a start and end time that are equal. You'll have to look at your unit tests, if they are passing they are probably incorrect as well...

fniessink (changes requested):

components/notifier/src/strategies/reds_that_are_new.py
Cas-Ros: replaced the comparison between the status of the metric and the most recent measurement with a comarison between the start and end of the most recent measurement

components/notifier/src/quality_time_notifier.py
Cas-Ros: corrected the message to only contain 1 report, and removed the leading sentence as it's now unnecessary

fniessink (approved):

PR #1607, opened 2020-10-24
Move metric source tab to the first position

ds-o:
When a metric is expanded to view the details, the tab with findings (such as security warnings) is now at the first position (instead of the metric and source configuration tabs).


fniessink:
We won't merge this because the detail information is retrieved using a separate API call whenever the user expands a metric. Putting this tab at the first position would mean that the tabs move after the user expands the metric. This will mean people will easily click the wrong tab if the detail information takes more than a few milliseconds to load.

PR #1613, opened 2020-10-26
Iks305 notify when metric status becomes white

Cas-Ros:


components/notifier/src/strategies/reds_that_are_new.py
```diff
@@ -53,3 +54,16 @@ def turned_red(metric, most_recent_measurement_seen: str) -> bool:
...
+    """Determine if a metric turned white after the timestamp of the most recent measurement seen."""
+    scale = metric["scale"]
+    metric_is_white = metric["status"] == "unknown"
+    recent_measurements = metric.get("recent_measurements")
+    if recent_measurements and len(recent_measurements) > 1:
+        metric_was_not_white = recent_measurements[-2][scale]["status"] != "unknown"
+        return bool(
+            metric_is_white and metric_was_not_white and
+            recent_measurements[-1]["start"] > most_recent_measurement_seen)
+    return False
```
fniessink: There's quite a bit of duplication with `turned_red`. How about we generalize `turned_white` and use it for both statuses?
```suggestion
def has_new_status(metric, new_status: str, most_recent_measurement_seen: str) -> bool:
    """Determine if a metric got a new status after the timestamp of the most recent measurement seen."""
    recent_measurements = metric.get("recent_measurements") or []
    if len(recent_measurements) < 2:
        return False  # If there are fewer than two measurements, the metric didn't recently change status
    metric_has_status = metric["status"] == new_status
    scale = metric["scale"]
    metric_had_other_status = recent_measurements[-2][scale]["status"] != new_status
    change_was_recent = recent_measurements[-1]["start"] > most_recent_measurement_seen
    return bool(metric_has_status and metric_had_other_status and change_was_recent)
```

fniessink (changes requested):
If we remove the duplication between `turned_red` and `turned_white` (see suggestion below), adding other status changes will be easy and I think we will also fix that bug we found.

fniessink (approved):

PR #1733, opened 2020-12-07
Iks315 notify on all status changes

Cas-Ros:


components/notifier/tests/strategies/test_changed_status.py
```diff
@@ -155,50 +155,46 @@ def test_new_red_metric_without_count_scale(self):
...
-        )
-        self.assertTrue(has_new_status(metric, self.most_recent_measurement_seen, self.white_metric_status))
-        self.assertFalse(has_new_status(metric, self.most_recent_measurement_seen, self.red_metric_status))
-        self.assertTrue(
-            has_new_status(metric, self.most_recent_measurement_seen, self.white_metric_status, self.red_metric_status)
-        )
+                dict(start=self.new_timestamp, count=dict(status="target_met")),
+                dict(start=self.old_timestamp, count=dict(status=self.white_metric_status))])
+        result = status_changed(metric, self.most_recent_measurement_seen)
+        self.assertTrue(result)
```
fniessink:
```suggestion
        self.assertTrue(status_changed(metric, self.most_recent_measurement_seen))
```

components/notifier/tests/test_notification.py
```diff
@@ -0,0 +1,27 @@
+""""Unit tests for the notification."""
+
+import unittest
+
+from notification import Notification
+
+
+class StrategiesTestCase(unittest.TestCase):
```
fniessink: Why is this called "StrategiesTestCase", it's just testing the Notification class, right?

```suggestion
class NotificationTestCase(unittest.TestCase):
```

docs/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Accepting the technical debt of a metric with one or more unreachable sources would set the metric status to 'technical debt target met' instead of 'unknown'. Fixes [#1636](https://github.com/ICTU/quality-time/issues/1636).
 - Collapsing an expanded metric would sometimes result in a crash of the frontend. Fixes [#1717](https://github.com/ICTU/quality-time/issues/1717).
+- Notifications can now be send for all status changes. Previously notifications where only sent out when a metric either turned red or white.
```
fniessink: Status changes aren't configurable, right? So shouldn't this read "will be sent" instead of "can be sent"?

```suggestion
- Notifications will now be sent for all status changes. Previously, notifications were only sent when a metric either turned red or white.
```

fniessink (changes requested):

PR #1738, opened 2020-12-10
Issue 1649 trend table per subject

emdv88:


components/server/src/database/measurements.py
```diff
@@ -39,9 +39,18 @@ def recent_measurements_by_metric_uuid(database: Database, max_iso_timestamp: st
     return measurements_by_metric_uuid
 
 
-def all_measurements(database: Database, metric_uuid: MetricId, max_iso_timestamp: str = ""):
-    """Return all measurements, without the entities, except for the most recent one."""
-    measurement_filter: Dict[str, Union[str, Dict[str, str]]] = {"metric_uuid": metric_uuid}
+def measurements_by_metric(
+    database: Database,
+    metric_uuids: Union[MetricId, List[MetricId]],
```
fniessink: Why not *metric_uuids? That would make the isinstance check unnecessary.

fniessink (changes requested):

components/server/src/database/measurements.py
emdv88: This is personal preference. Because there is an endpoint that will always only ask measurements for one metric, I prefer to have the typecheck here instead of creating a list with one metric_uuid in it in the calling function.

fniessink: But that's not needed:

```
>>> def i_take_one_or_more_metric_ids(*metric_ids):
...   print(metric_ids)
...
>>> i_take_one_or_more_metric_ids("a")
('a',)
>>> i_take_one_or_more_metric_ids("a", "b")
('a', 'b')
>>> i_take_one_or_more_metric_ids(*["a", "b"])
('a', 'b')
```

emdv88: That's a nice solution aswell. Gotta be careful with the variable order, but in this case it does make sense. I'll change it right away.

fniessink (approved):
LGTM. Only thing left to do before merge is a rebase on master. I've tried to rebase, but it seems you have merged from master to your branch which makes the rebase quite hard. Can you give it a go?

PR #1740, opened 2020-12-11
Iks366 refactor metric dict in notifier to class

Cas-Ros:


fniessink (approved):

components/notifier/src/models/metric_notification_data.py
```diff
@@ -0,0 +1,27 @@
+"""Metric data needed for notifications."""
+
+
+class MetricNotificationData:
```
fniessink:
```suggestion
class MetricNotificationData:  # pylint: disable=too-few-public-methods
```

fniessink (changes requested):

PR #1772, opened 2020-12-17
Iks311 notify when a metric has had the same status for 3 weeks

Cas-Ros:


components/notifier/tests/strategies/test_notification_strategy.py
```diff
@@ -5,7 +5,7 @@
 import pathlib
 import unittest
 
-from strategies.changed_status import get_notable_metrics_from_json, status_changed
+from strategies.notification_strategy import NotableEvents
 
 
 class StrategiesTestCase(unittest.TestCase):
```
Cas-Ros: Pylint says there are to many public methods. Should this be ignored or should this file be split?

components/notifier/src/models/metric_notification_data.py
```diff
@@ -4,7 +4,7 @@
 class MetricNotificationData:  # pylint: disable=too-few-public-methods
     """Handle metric data needed for notifications."""
 
-    def __init__(self, metric, data_model):
+    def __init__(self, metric, data_model, reason):
```
fniessink: Reason is a string I assume?
```suggestion
    def __init__(self, metric, data_model, reason: str) -> None:
```

components/notifier/src/quality_time_notifier.py
```diff
@@ -24,8 +24,9 @@ async def notify(log_level: int = None) -> NoReturn:
         f"{os.environ.get('SERVER_PORT', '5001')}/api/{api_version}/reports"
     )
     data_model = await retrieve_data_model(api_version)
-    most_recent_measurement_seen = datetime.max.isoformat()
+    most_recent_measurement_seen = datetime.max.isoformat() + "+00:00"
```
fniessink, line 28: Please add a comment explaining why the "+00:00" needs to be added.

components/notifier/src/quality_time_notifier.py
```diff
@@ -24,8 +24,9 @@ async def notify(log_level: int = None) -> NoReturn:
...
     outbox = Outbox()
+    notable = NotableEvents(data_model)
```
fniessink: "notable" is an adjective. Class instance names should be a noun. It's a bit hard in this case, but maybe "notification_finder"?

components/notifier/src/quality_time_notifier.py
```diff
@@ -36,7 +37,7 @@ async def notify(log_level: int = None) -> NoReturn:
         except Exception as reason:  # pylint: disable=broad-except
             logging.error("Could not get reports from %s: %s", reports_url, reason)
             json = dict(reports=[])
-        notifications = get_notable_metrics_from_json(data_model, json, most_recent_measurement_seen)
+        notifications = notable.get_notable_metrics_from_json(json, most_recent_measurement_seen)
```
fniessink: The method is called "get_notable_metrics_from_json" but the result is named "notifications". They can't both be correct. Please rename either the method or the variable

components/notifier/src/strategies/notification_strategy.py
```diff
@@ -0,0 +1,67 @@
+"""Strategies for notifying users about metrics."""
+
+from datetime import datetime, timedelta
+from typing import List, Optional
+from models.notification import Notification
+from models.metric_notification_data import MetricNotificationData
+
+
+class NotableEvents:
```
fniessink: Are instance of this class notable events? I don't think so. The responsibility of this class is to determine notifications, so maybe call it something like "NotificationFinder"?

components/notifier/src/strategies/notification_strategy.py
```diff
@@ -0,0 +1,67 @@
...
+    """Handle notification contents and status."""
+
+    def __init__(self, data_model):
+        """Initialise already notified list and store data model."""
+        self.already_notified = []
+        self.data_model = data_model
+
+    def get_notable_metrics_from_json(self, json, most_recent_measurement_seen: str) -> List[Notification]:
```
fniessink: As mentioned where this method is called: the method name says it returns notable metrics, but it really returns instances of the Notification class so a better name would be get_notifications.

components/notifier/src/strategies/notification_strategy.py
```diff
@@ -0,0 +1,67 @@
...
+        recent_measurements = metric.get("recent_measurements") or []
+        if len(recent_measurements) < 2:
+            return False  # If there are fewer than two measurements, the metric couldn't have recently changed status
+        scale = metric["scale"]
+        metric_had_other_status = recent_measurements[-2][scale]["status"] != recent_measurements[-1][scale]["status"]
+        change_was_recent = recent_measurements[-1]["start"] > most_recent_measurement_seen
+        return bool(metric_had_other_status and change_was_recent)
+
+    def long_unchanged_status(self, metric, metric_uuid, most_recent_measurement_seen: str) -> bool:
+        """Determine if a metric has been red for 3 weeks."""
```
fniessink: Not just red.

components/notifier/src/strategies/notification_strategy.py
```diff
@@ -0,0 +1,67 @@
...
+        status_start = datetime.fromisoformat(metric["status_start"])
+        datetime_most_recent_measurement_seen = datetime.fromisoformat(most_recent_measurement_seen)
+        difference = datetime_most_recent_measurement_seen - status_start
+        if timedelta(days=21).total_seconds() < difference.total_seconds() < timedelta(
+            days=22
+        ).total_seconds() and not self.already_notified.__contains__(metric_uuid):
```
fniessink: Timedelta's can be compared directly. Also, don't use __contains__ directly, use "in":
```suggestion
        if timedelta(days=21) < difference < timedelta(days=22) and metric_uuid not in self.already_notified:
```

fniessink (changes requested):

components/notifier/tests/strategies/test_notification_strategy.py
fniessink: The class "StrategiesTestCase" needs to be split in two classes. I'd suggest one class for testing the notifications for changed status and one for notifications for metrics that have the same status for a long period.

fniessink (approved):

PR #1903, opened 2021-02-08
Issue 1644 merge requests

### LINKED ISSUE #1644: Add metric for merge requests, particularly without review
Add a new metric "merge requests".
Add GitLab and Azure Devops as source.
Parameters depend on the source, but preferably state (closed, open, ...) and nr approvals (integer).
Entities = merged requests, columns = created date time, closed date time, approvers, issue title (links to issue), branch name (links to branch)

fniessink:
Closes #1644.

### DIFF
--- a/components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
+++ b/components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
@@ -15,7 +15,7 @@ class AzureDevopsMergeRequests(AzureDevopsRepositoryBase):
     async def _api_url(self) -> URL:
         """Extend to add the pull requests API path."""
         api_url = str(await super()._api_url())
-        return URL(f"{api_url}/pullrequests?api-version=4.1")
+        return URL(f"{api_url}/pullrequests?api-version=4.1&searchCriteria.status=all")
 
     async def _landing_url(self, responses: SourceResponses) -> URL:
         """Extend to add the pull requests path."""
--- a/components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
+++ b/components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
@@ -27,12 +27,13 @@ class AzureDevopsMergeRequests(AzureDevopsRepositoryBase):
         merge_requests = []
         for response in responses:
             merge_requests.extend((await response.json())["value"])
+        landing_url = (await self._landing_url(responses)).rstrip("s")
         entities = [
             Entity(
                 key=merge_request["pullRequestId"],
                 title=merge_request["title"],
                 target_branch=merge_request["targetRefName"],
-                url=merge_request["url"],
+                url=f"{landing_url}/{merge_request['pullRequestId']}",
                 state=merge_request["status"],
                 created=merge_request.get("creationDate"),
                 closed=merge_request.get("closedDate"),
--- a/components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
+++ b/components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
@@ -12,15 +12,26 @@ from .base import AzureDevopsRepositoryBase
 class AzureDevopsMergeRequests(AzureDevopsRepositoryBase):
     """Collector for merge requests (pull requests in Azure Devops)."""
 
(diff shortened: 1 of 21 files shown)


components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+            for merge_request in merge_requests
+            if self._include_merge_request(merge_request)
+        ]
+        return SourceMeasurement(entities=entities, total=str(len(merge_requests)))
+
+    def _include_merge_request(self, merge_request) -> bool:
+        """Return whether the merge request should be counted."""
+        min_upvotes = int(cast(str, self._parameter("upvotes")))
+        request_has_fewer_than_min_upvotes = min_upvotes == 0 or self._upvotes(merge_request) < min_upvotes
```
emdv88: Is there a reason that this will always be true if the minimum amount of upvotes is 0? Shouldn't it always be false in that case?

components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
```diff
@@ -0,0 +1,66 @@
...
+                url=f"{landing_url}/{merge_request['pullRequestId']}",
+                state=merge_request["status"],
+                created=merge_request.get("creationDate"),
+                closed=merge_request.get("closedDate"),
+                downvotes=str(self._downvotes(merge_request)),
+                upvotes=str(self._upvotes(merge_request)),
+            )
+            for merge_request in merge_requests
+            if self._include_merge_request(merge_request)
+        ]
```
emdv88: For readability maybe use a simple for loop instead of list comprehension? Might be just personal preference though

components/collector/src/source_collectors/api_source_collectors/gitlab/merge_requests.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+            for merge_request in merge_requests
+            if self._include_merge_request(merge_request)
+        ]
+        return SourceMeasurement(entities=entities, total=str(len(merge_requests)))
+
+    def _include_merge_request(self, merge_request) -> bool:
+        """Return whether the merge request should be counted."""
+        min_upvotes = int(cast(str, self._parameter("upvotes")))
+        request_has_fewer_than_min_upvotes = min_upvotes == 0 or int(merge_request["upvotes"]) < min_upvotes
```
emdv88: same question as for the azure function.

components/collector/src/source_collectors/api_source_collectors/sonarqube.py
```diff
@@ -27,7 +23,7 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
         response = (await super()._get_source_responses(show_component_url))[0]
         json = await response.json()
         if "errors" in json:
-            raise SonarQubeException(json["errors"][0]["msg"])
+            raise SourceCollectorException(json["errors"][0]["msg"])
```
emdv88, line 26: Do the errormessages that we get from SonarQube show that they are SonarQube errors? Otherwise, with this change we lost the information that it was the SonarQube api that threw the error.

components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
fniessink: It's a bit counterintuitive so I'll add a comment, but I believe it's correct as currently implemented: if the minimum number of upvotes is zero, merge requests are included regardless of how many upvotes they actually have. If the minimum number of upvotes is more than zero then only merge requests that have less than the minimum number of upvotes are included in the count.

components/collector/src/source_collectors/api_source_collectors/azure_devops/merge_requests.py
fniessink: I've extracted the creation of the entity to a separate method so that the list comprehension fits on one line.

components/collector/src/source_collectors/api_source_collectors/sonarqube.py
fniessink: In the UI, the error is shown with the source so it's always clear which source caused the error. In the logging, the URL is included so there it's clear too.

PR #1921, opened 2021-02-18
Issue 1693 export endpoint

emdv88:


emdv88:
snyk tests fail, but I cannot access it because "i'm not a member of this organization". Can you provide me access?

components/server/src/initialization/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.primitives import serialization
+
+
+EXPORT_FIELDS_KEYS_NAME = "export_fields_keys"
+EXPORT_FIELDS_USAGE_DESCRIPTION = """
+Use this public/private key fields when exporting sensitive
+content (like api credentials) from the Quality-Time database.
```
fniessink:
```suggestion
content (like api credentials) from the Quality-time database.
```

components/server/src/initialization/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def initialize_secrets(database):
+    """Make sure public and private keys exist."""
+    # do nothing if it already exists
+    if database.secrets.find_one({"name": EXPORT_FIELDS_KEYS_NAME}):
+        return
+
+    # else: create new public/private key pair and insert it
+    # private key
```
fniessink:
```suggestion
```

Comment seems unnecessary as the variable name is pretty clear.

components/server/src/initialization/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+    private_key = rsa.generate_private_key(public_exponent=65537, key_size=4096, backend=default_backend())
+    privkey = private_key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+
+    # public key
```
fniessink:
```suggestion
```

Comment seems unnecessary as the variable name is pretty clear.

components/server/src/initialization/secrets.py
```diff
@@ -0,0 +1,45 @@
...
+    public_key = private_key.public_key()
+    pubkey = public_key.public_bytes(
+        encoding=serialization.Encoding.PEM, format=serialization.PublicFormat.SubjectPublicKeyInfo
+    )
+
+    # store everything in the database
```
fniessink:
```suggestion
```

Comment seems unnecessary as `database.secrets.insert` is quite descriptive.

components/server/src/model/transformations.py
```diff
@@ -18,6 +18,21 @@ def hide_credentials(data_model, *reports) -> None:
...
+    """Encrypt all credentials in the reports."""
+    for source in iter_sources(reports):
+        for parameter_key, parameter_value in source.get("parameters", {}).items():
+            if parameter_value and is_password_parameter(data_model, source["type"], parameter_key):
+                password = source["parameters"][parameter_key]
+                if isinstance(password, (dict, list)):
+                    password = json.dumps(password)
+
+                password_bytes = password.encode()
+                key_encrypted_value = asymmetric_encrypt(public_key.encode(), password_bytes)
```
fniessink: Why is this variable called "key_encrypted_value" and not simply "encrypted_value"?

components/server/src/routes/report.py
```diff
@@ -87,6 +89,29 @@ def export_report_as_pdf(report_uuid: ReportId):
     return response.content
 
 
+@bottle.get("/api/v3/report/<report_uuid>/json")
+def export_report_as_json(database: Database, report_uuid: ReportId):
+    """Return the quality report, including information about other reports needed for move/copy actions."""
```
fniessink: Is this part of the comment correct: "including  information about ..."?

components/server/src/server_utilities/functions.py
```diff
@@ -68,3 +74,34 @@ def percentage(numerator: int, denominator: int, direction: Direction) -> int:
     if denominator == 0:
         return 0 if direction == "<" else 100
     return int((100 * Decimal(numerator) / Decimal(denominator)).to_integral_value(ROUND_HALF_UP))
+
+
+def symmetric_encrypt(message: bytes) -> Tuple[bytes, bytes]:
+    """
+    Encrypt the given value using Fernet 32 byte key.
+    Returns the generated key and the encrypted message.
+    """
```
fniessink: DeepSource will complain about this docstring format I think.

components/server/tests/server_utilities/test_functions.py
```diff
@@ -21,6 +27,42 @@ def test_uuid(self):
...
+        key, encrypted_message = symmetric_encrypt(test_message)
+        self.assertNotEqual(test_message, encrypted_message)
+
+        # decryption
+        fernet = Fernet(key)
+        decrypted_message = fernet.decrypt(encrypted_message)
+        self.assertEqual(decrypted_message, test_message)
+
+    def test_asymmetric_encrypt(self):
+        """Test wether message is encripted using public/private keys."""
```
fniessink:
```suggestion
        """Test wether message is encrypted using public/private keys."""
```

tests/feature_tests/steps/report.py
```diff
@@ -47,6 +68,20 @@ def check_pdf(context):
     assert_equal("application/pdf", context.response.headers["Content-Type"])
 
 
+@then("the client receives the json")
+def check_json(context):
+    """Check the pdf."""
```
fniessink:
```suggestion
    """Check the json."""
```

tests/feature_tests/steps/report.py
```diff
@@ -47,6 +68,20 @@ def check_pdf(context):
...
+    assert_equal(200, context.response.status_code)
+    assert_equal("application/json", context.response.headers["Content-Type"])
+
+
+@then("the client receives a json error")
+def check_json_error(context):
+    """Check the pdf."""
```
fniessink:
```suggestion
    """Check the json error."""
```

fniessink (changes requested):

components/server/src/model/transformations.py
emdv88: Because it's only the key of the symmetric encryption that is encrypted, not the value of the message itself. 
But indeed, maybe something like encrypted_fernet_key would be more descriptive.

components/server/src/routes/report.py
emdv88: Mh, I actually don't think it is. Don't remember why I put it there though.

components/server/src/server_utilities/functions.py
emdv88: It seems to be ok with it for now. That being said, I think there could be a cleaner style to specify the returned value.

components/server/src/model/transformations.py
emdv88: disregard my previous comment, it's called like this because this function returns a tuple with the rsa encrypted fernet key and the fernet (symmetric) encrypted value of the password.

emdv88:
Changed the code on all remaining comments. Should be ready for merging I think.

One note would be that apparently sometimes the feature tests don't start from an empty database which means that the rsa encryption secrets don't need to be generated, leaving that code uncovered. I have no real clue when or why this happens, but for now it was fixed by simply rerunning the feature tests in circleci.

PR #1966, opened 2021-03-01
When measuring security warnings with OWASP ZAP as source, allow for …

### LINKED ISSUE #1902: Add option  to choose between ZAP alerts or instances
Quality time now only shows ZAP instances and I would like to choose between instances or alerts (one lever higher).

fniessink:
…counting alert types as security warnings as opposed to alert instances. Closes #1902.

### DIFF
--- a/components/collector/src/source_collectors/owasp_zap/security_warnings.py
+++ b/components/collector/src/source_collectors/owasp_zap/security_warnings.py
@@ -29,16 +29,24 @@ class OWASPZAPSecurityWarnings(XMLFileSourceCollector):
             entity_kwargs = dict(name=name, description=description, risk=risk)
             if count_alert_instances:
                 for alert_instance in alert.findall("./instances/instance"):
-                    method = alert_instance.findtext("method", default="")
-                    uri = self.__stable(hashless(URL(alert_instance.findtext("uri", default=""))))
-                    key = md5_hash(f"{':'.join(ids)}:{method}:{uri}")
-                    old_key = md5_hash(f"{':'.join(ids[1:])}:{method}:{uri}")
-                    location = f"{method} {uri}"
-                    entities[key] = Entity(key=key, old_key=old_key, uri=uri, location=location, **entity_kwargs)
+                    entity = self.__alert_entity(ids, entity_kwargs, alert_instance)
+                    entities[entity["key"]] = entity
             else:
-                key = md5_hash(f"{':'.join(ids)}")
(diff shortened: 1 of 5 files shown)


components/collector/src/source_collectors/owasp_zap/security_warnings.py
```diff
@@ -14,32 +14,40 @@ class OWASPZAPSecurityWarnings(XMLFileSourceCollector):
...
+            if count_alert_instances:
+                for alert_instance in alert.findall("./instances/instance"):
+                    entity = self.__alert_entity(ids, entity_kwargs, alert_instance)
+                    entities[entity["key"]] = entity
+            else:
+                entity = self.__alert_entity(ids, entity_kwargs)
+                entities[entity["key"]] = entity
         return list(entities.values())
 
+    def __alert_entity(self, ids, entity_kwargs, alert_instance=None) -> Entity:
```
emdv88: maybe make this two functions? one alert_entity_type and one alert_entity, In the end these are very different entities.

components/collector/src/source_collectors/owasp_zap/security_warnings.py
```diff
@@ -14,32 +14,40 @@ class OWASPZAPSecurityWarnings(XMLFileSourceCollector):
     """Collector to get security warnings from OWASP ZAP."""
 
     async def _parse_entities(self, responses: SourceResponses) -> Entities:
-        """Override to parse the securty warnings from the XML."""
+        """Override to parse the security warnings from the XML."""
         entities: dict[str, Entity] = {}
         tag_re = re.compile(r"<[^>]*>")
         risks = cast(list[str], self._parameter("risks"))
+        count_alert_instances = cast(str, self._parameter("alerts")) == "alert instances"
```
emdv88, line 21: I might be misunderstanding this, but does this default to "alert types" right now? Maybe it would be better to default to the old behaviour, so "alert instances".

fniessink: It does default to alert instances. See the default value specified in the data model.

components/collector/src/source_collectors/owasp_zap/security_warnings.py
fniessink: You're right. I will change this.

components/collector/src/source_collectors/owasp_zap/security_warnings.py
emdv88: ah true

PR #1987, opened 2021-03-11
1693 report import

emdv88:


emdv88:
There seems to be something that makes Bettercodehub unhappy, but I can't figure what it was. It does not seem to be something that I changed in the code.

emdv88:
Same goes for deepsource. The error is raised on a piece of code that I never touched. Did the settings change?

emdv88:
This should be ready for merging now. The changelog has been updated.

The better Code hub seems to still fail, but indeed I think a bigger refactor might be the best solution to fix this. An alternative idea for this refactor might also be to make the Database a singleton class. That way, it does not need to be injected into every endpoint anymore.

Alternatively indeed creating classes for reports and/or subjects could help aswell.

emdv88:
Edit: I have no idea why there is suddenly one line in the datamodels.py that is not covered anymore.

components/server/src/data/datamodel.json
```diff
@@ -4586,6 +4586,17 @@
                         "security_warnings"
                     ]
                 },
+                "landing_url": {
+                    "name": "URL to a Safety report in a human readable format",
+                    "short_name": "URL",
+                    "help": "If provided, users clicking the source URL will visit this URL instead of the Robot Framework report in XML format.",
```
fniessink:
```suggestion
                    "help": "If provided, users clicking the source URL will visit this URL instead of the Safety report in JSON format.",
```

components/server/src/data/example-reports/example-report-hq.json
```diff
@@ -1,144 +1,315 @@
...
     "report_uuid": "example-report-hq",
-    "subjects": [
-        {
-            "name": "HQ",
+    "subjects": {
+        "dc8c9e76-123f-4acc-a30b-f0b02753e908": {
             "type": "software",
-            "metrics": [
-                {
+            "name": "HQ",
```
fniessink: We should really replace this example with something else because HQ doesn't exist in Sonarcloud.io anymore.

components/server/src/routes/auth.py
```diff
@@ -134,3 +135,11 @@ def logout(database: Database) -> dict[str, bool]:
     """Log the user out."""
     delete_session(database)
     return dict(ok=True)
+
+
+@bottle.get("/api/v3/public_key")
+def get_public_key(database: Database) -> dict:
+    """Return a serialized version of the public key."""
+    public_key = database.secrets.find_one({"name": EXPORT_FIELDS_KEYS_NAME}, {"public_key": True, "_id": False})
```
fniessink, line 143: Shouldn't we add `"private_key": False` so the private key isn't needlessly passed to the server?

components/server/src/routes/report.py
```diff
@@ -40,7 +44,22 @@ def post_report_import(database: Database):
...
+
+    secret = database.secrets.find_one({"name": EXPORT_FIELDS_KEYS_NAME}, {"private_key": True, "_id": False})
+    private_key = secret["private_key"]
+
+    try:
+        decrypt_credentials(data_model, private_key, report)
+    except DecryptionError:
+        bottle.response.status = 400
+        bottle.response.content_type = "application/json"
+        return {"error": "Decryption of source credentials failed. Did you use the correct public key for encryption?"}
```
fniessink: I find the question "Did you use the correct public key for encryption?" a bit abstract. Suggestion: "Did you use the public key of this Quality-time instance to encrypt this report?"

components/server/tests/server_utilities/test_functions.py
```diff
@@ -39,6 +47,16 @@ def test_symmetric_encryption(self):
         decrypted_message = fernet.decrypt(encrypted_message)
         self.assertEqual(decrypted_message, test_message)
 
+    def test_symmetric_decryption(self):
+        """Test wether message that has been encrypted using symmetric_encrypt ca be decrypted."""
```
fniessink:
```suggestion
        """Test whether the message that has been encrypted using symmetric_encrypt can be decrypted."""
```

components/server/tests/server_utilities/test_functions.py
```diff
@@ -39,6 +47,16 @@ def test_symmetric_encryption(self):
...
+        # encryption
+        test_message = b"this is a test message"
+        key, encrypted_message = symmetric_encrypt(test_message)
+
+        # decryption
+        message = symmetric_decrypt(key, encrypted_message)
+        self.assertEqual(message, test_message)
+
     def test_asymmetric_encrypt(self):
         """Test wether message is encrypted using public/private keys."""
```
fniessink:
```suggestion
        """Test whether the message is encrypted using public/private keys."""
```

components/server/tests/server_utilities/test_functions.py
```diff
@@ -59,10 +77,34 @@ def test_asymmetric_encrypt(self):
             padding.OAEP(mgf=padding.MGF1(algorithm=hashes.SHA256()), algorithm=hashes.SHA256(), label=None),
         )
         fernet = Fernet(decrypted_key)
-        decrypted_message = fernet.decrypt(encrypted_message.encode())
+        decrypted_message = fernet.decrypt(encrypted_message.encode()).decode()
 
         self.assertEqual(decrypted_message, test_message)
 
+    def test_asymmetric_decrypt(self):
+        """Test wether message that has been encrypted using asymmetric_encrypt ca be decrypted."""
```
fniessink:
```suggestion
        """Test whether the message that has been encrypted using asymmetric_encrypt can be decrypted."""
```

fniessink (changes requested):
Documentation in the user guide is missing. We document the pdf export API, I think we should also document this API in the user guide.

components/server/src/routes/auth.py
emdv88: That's not necessary. From the mongo docs I read that if a field is explicitly included, al non-mentioned field will be automatically excluded. With the exception of _id which needs to be excluded explicitly.

https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/

components/server/src/data/example-reports/example-report-hq.json
emdv88: Maybe simply leave away this example report? I think 3 example reports should principally be sufficient, especially if we are gonna have to spend time to maintain them.

components/server/src/routes/auth.py
fniessink: Ah, didn't know that. Good to know.

components/server/src/routes/report.py
emdv88: changed according to your suggestion.

docs/USAGE.md
```diff
@@ -318,6 +318,25 @@ To change the number of dates and the time between dates shown in the trend tabl
 
 To export an older version of a report, add the `report_date` parameter with a date value in ISO-format (YYYY-MM-DD), for example `http://www.quality-time.example.org/api/v3/report/<report_uuid>/pdf?report_date=2020-09-01`.
 
+## JSON report import/export
+
+*Quality-time* provides functionality for importing and exporting reports in JSON format. This functionality can be used for backing up reports or for transferring reports from one *Quality-time* instance to another one. Currently this functionality is only available via the api, with one endpoint for importing and one for exporting the JSON reports.
```
fniessink:
```suggestion
*Quality-time* provides functionality for importing and exporting reports in JSON format. This functionality can be used for backing up reports or for transferring reports from one *Quality-time* instance to another one. Currently this functionality is only available via the API, with one endpoint for importing and one for exporting the JSON reports.
```

docs/USAGE.md
```diff
@@ -318,6 +318,25 @@ To change the number of dates and the time between dates shown in the trend tabl
...
+
+A *Quality-time* report JSON contains the configuration of the report, with all it's subjects, metrics and sources. It does not contain any actual measurements. 
+The credentials of configured sources will be encrypted to prevent the loss of sensitive data.
+
+### Export API
+
+The exporting endpoint is available under `http://www.quality-time.example.org/api/v3/report/<report-uuid>/json?public_key=<public-key>`. The public key argument is optional, if no public key is provided, the public key of the exporting *Quality-time* instance is used for encrypting the source credentials. If the report needs to be imported in a different *Quality-time* instance, the public key of that instance should be provided. It can be opbtained at `www.quality-time.example.org/public_key`. The exported JSON report can only be imported into the *Quality-time* who's public key has been used for the credential encryption during the export.
```
fniessink: Is that really the correct URL for the public key?

```suggestion
The exporting endpoint is available under `http://www.quality-time.example.org/api/v3/report/<report-uuid>/json?public_key=<public-key>`. The public key argument is optional. If no public key is provided, the public key of the exporting *Quality-time* instance is used for encrypting the source credentials. If the report needs to be imported in a different *Quality-time* instance, the public key of that instance should be provided. It can be obtained at `www.quality-time.example.org/api/v3/public_key`. The exported JSON report can only be imported into the *Quality-time* whose public key has been used for the encryption of credentials during the export.
```

docs/USAGE.md
```diff
@@ -318,6 +318,25 @@ To change the number of dates and the time between dates shown in the trend tabl
...
+
+### Import API
+
+Exported reports can be imported back into *Quality-time*. The endpoint for this is `http://www.quality-time.example.org/api/v3/report/import`. The import accepts JSON content only.
+
+Only exports that have been exported using the public key of the imported *Quality-time* instance can be imported. The Source credentials will be decrypted on import and the sources should be ready to collect measurements automatically.
```
fniessink:
```suggestion
Only exports that have been exported using the public key of the destination *Quality-time* instance can be imported. The source credentials will be decrypted on import and the sources should be ready to collect measurements automatically.
```

fniessink (changes requested):
I think the documentation needs a few improvements, especially the public key that is mentioned doesn't look correct to me.

fniessink (approved):

PR #2008, opened 2021-03-22
Refactor scale handling

fniessink:


components/server/src/database/measurements.py
```diff
@@ -1,27 +1,29 @@
...
 from model.metric import Metric
-from model.queries import get_attribute_type, get_measured_attribute
-from server_utilities.functions import iso_timestamp, percentage
-from server_utilities.type import MeasurementId, MetricId, Scale, Status, TargetType
+from server_utilities.functions import iso_timestamp
+from server_utilities.type import MeasurementId, MetricId
 
 
-def latest_measurement(database: Database, metric_uuid: MetricId):
+def latest_measurement(database: Database, metric_uuid: MetricId, metric: Metric) -> Optional[Measurement]:
```
emdv88: Edited: shouldn't this function (and all other functions that are in this file and that apply to one specific metric) become methods of the Metric class?

In combination with a global Database singleton that method would not need any input arguments.

components/server/src/database/measurements.py
```diff
@@ -72,86 +74,16 @@ def update_measurement_end(database: Database, measurement_id: MeasurementId):
     return database.measurements.update_one(filter={"_id": measurement_id}, update={"$set": {"end": iso_timestamp()}})
 
 
-def insert_new_measurement(
-    database: Database, data_model, metric_data: dict, measurement: dict, previous_measurement: dict
-) -> dict:
+def insert_new_measurement(database: Database, measurement: Measurement) -> Measurement:
```
emdv88, line 77: This could be a class method on the Measurement class:

Measurement.insert(metric: Metric, measurement_data: dict)

components/server/src/model/measurement.py
```diff
@@ -0,0 +1,87 @@
...
+from collections.abc import Sequence
+from typing import Optional, cast
+
+from model.metric import Metric
+from model.source import Source
+from server_utilities.functions import iso_timestamp, percentage
+from server_utilities.type import Scale, Status, TargetType
+
+
+class Measurement(dict):  # lgtm [py/missing-equals]
```
emdv88, line 122: I like the idea of subclassing dict!

components/server/src/model/source.py
```diff
@@ -0,0 +1,49 @@
+"""Source model class."""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
emdv88, line 12: This is an interesting solution... Up until now I used quotes if a real import wasn't possible due to circular imports

components/server/src/model/source.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, metric: Metric, *args, **kwargs):
+        self.__metric = metric
+        super().__init__(*args, **kwargs)
+
+    def total(self) -> int:
+        """Return the measurement total of the source."""
+        return int(self["total"] or "100")
+
+    def value(self) -> int:
```
emdv88, line 31: What is the value (and the total) of a source? Isn't it the Metric (or a measurement) that has a value?

components/server/src/database/measurements.py
fniessink: Nice idea: Metric.latest_measurement() and Metric.latest_successful_measurement(). Disadvantage is that it would mix the model layer (Metric) with the database layer.

components/server/src/model/source.py
fniessink: Yeah, must love the `__future__` :-)

components/server/src/model/source.py
fniessink: A measurement can have multiple sources. The values of the sources are combined (by taking the sum, min or max) to arrive at the measurement value. Similar for the total.

components/server/src/database/measurements.py
fniessink: Same comment as above: looks nice, but mixes model and database layer. Let's discuss.

components/server/src/database/measurements.py
emdv88: What is exactly the thought behind the database layer? Looking through it, my intuition would indeed be to move basically everything that is currently in there onto the different models. But that might just be me being too used to orm patterns :D

fniessink: I prefer model classes to have no knowledge of storage or presentation, cf. "clean architecture". See e.g. https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html

PR #2015, opened 2021-03-26
1842 permissions

fniessink:


components/frontend/src/report/ReportsTitle.js
```diff
@@ -31,10 +32,22 @@ export function ReportsTitle(props) {
...
+                            set_value={(value) => set_reports_attribute("permissions", {edit_reports: value}, props.reload)}
+                            value={props.permissions.edit_reports || []}
+                        />
+                    </Grid.Column>
+                </Grid.Row>
+                <Grid.Row columns={1}>
+                    <Grid.Column>
+                        <MultipleChoiceInput
+                            allowAdditions
+                            label="Users allowed to edit measured entities (user name or email address)"
```
fniessink, line 61: I don't think we use the term "entity" in the user interface at all, and I'm not sure we should start using it because it is such an abstract word. Maybe say: "Users allowed to manage false positives"?

components/server/src/routes/reports.py
```diff
@@ -35,8 +36,20 @@ def post_reports_attribute(reports_attribute: str, database: Database):
...
+            report_editors = value[EDIT_REPORT_PERMISSION]
+            if len(report_editors) > 0 and user["user"] not in report_editors and user["email"] not in report_editors:
+                report_editors.append(user["user"])  # Make sure users don't remove themselves as editor by accident
+            permissions[EDIT_REPORT_PERMISSION] = report_editors
+
+        if EDIT_ENTITY_PERMISSION in value:
+            permissions[EDIT_ENTITY_PERMISSION] = value[EDIT_ENTITY_PERMISSION]
+
+        value = permissions
+
```
fniessink: This needs some thorough unit testing 😁

components/server/src/routes/report.py
```diff
@@ -111,7 +112,7 @@ def export_report_as_pdf(report_uuid: ReportId):
     return response.content
 
 
-@bottle.get("/api/v3/report/<report_uuid>/json")
+@bottle.get("/api/v3/report/<report_uuid>/json", authentication_required=True)
 def export_report_as_json(database: Database, report_uuid: ReportId):
     """Return the quality-time report, including iencrypted credentials for api access to the sources."""
```
fniessink: Caught an unrelated typo:
```suggestion
    """Return the quality-time report, including encrypted credentials for api access to the sources."""
```

components/server/src/routes/documentation.py
```diff
@@ -5,13 +5,18 @@
...
-                url=f'http://www.quality-time.example.org:{port}{route.rule}', method=route.method,
-                description=route.callback.__doc__) for route in routes}
+    return {
+        route.rule: dict(
+            url=f"http://www.quality-time.example.org:{port}{route.rule}",
+            method=route.method,
+            description=route.callback.__doc__,
+        )
+        for route in routes
+    }
```
fniessink: Would be nice if the new route keyword arguments were included here too, e.g:

```suggestion
    return {
        route.rule: dict(
            url=f"http://www.quality-time.example.org:{port}{route.rule}",
            method=route.method,
            description=route.callback.__doc__,
            authentication_needed=route...,
            required_permissions=route...,
        )
        for route in routes
    }
```

fniessink (commented):
Looking good already!

components/frontend/src/report/ReportsTitle.js
emdv88: I was doubting about this as well. But it's not only false positives, isn't it? the user is given a number of other options aswell as far as I could tell.

Ideally the name would be based on the type of entity that is being measured I think, but that would require more work.

fniessink: Let's just bite the bullet and introduce the term here, and add it to the user docs as well?

emdv88: Yeah, i'd say that would be a good idea.

PR #2045, opened 2021-04-13
2037 timeout

### LINKED ISSUE #2037: TimeoutError on the ‘Failed CI-Jobs’ Metric in combination with Gitlab
The ‘Failed CI-Jobs’ Metric in combination with Gitlab causes/receives a ‘TimeoutError’ when there are many old/previous pipelines (and so also jobs). This is probably caused by the implementation of #1813 .

QT version: v3.19.1

Error:

```
Connection error
  File "/home/collector/base_collectors/source_collector.py", line 109, in __safely_get_source_responses
    responses = await self._get_source_responses(api_url)
  File "/home/collector/source_collectors/gitlab/base.py", line 31, in _get_source_responses
    all_responses.extend(responses := await super()._get_source_responses(*next_urls))
  File "/home/collector/base_collectors/source_collector.py", line 137, in _get_source_responses
    raise response
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 1104, in throw
    self._coro.throw(arg)
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client.py", line 544, in _request
    await resp.start(conn)
  File "/usr/local/lib/python3.9/site-packages/aiohttp/client_reqrep.py", line 905, in start
    self._continue = None
  File "/usr/local/lib/python3.9/site-packages/aiohttp/helpers.py", line 656, in __exit__
    raise asyncio.TimeoutError from None
asyncio.exceptions.TimeoutError
```

fniessink:
Fixes #2037.

### DIFF
--- a/components/collector/src/source_collectors/gitlab/failed_jobs.py
+++ b/components/collector/src/source_collectors/gitlab/failed_jobs.py
@@ -1,16 +1,12 @@
 """GitLab failed jobs collector."""
 
-from collector_utilities.type import URL, Job
+from collector_utilities.type import Job
 
 from .base import GitLabJobsBase
 
 
 class GitLabFailedJobs(GitLabJobsBase):
     """Collector class to get failed job counts from GitLab."""
-
-    async def _api_url(self) -> URL:
-        """Extend to return only failed jobs."""
-        return URL(str(await super()._api_url()) + "&scope=failed")
 
     def _count_job(self, job: Job) -> bool:
         """Return whether the job has failed."""
--- a/components/collector/src/base_collectors/metrics_collector.py
+++ b/components/collector/src/base_collectors/metrics_collector.py
@@ -3,7 +3,6 @@
 import asyncio
 import logging
 import os
-import random
 import traceback
 from datetime import datetime, timedelta
 from typing import Any, Final, NoReturn, cast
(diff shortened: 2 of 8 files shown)


emdv88:
I think the unit test fails because it will now randomly collect a measurement even if no 15 minutes have passed. The random method should probably be patched for this test.

components/collector/src/source_collectors/gitlab/failed_jobs.py
```diff
@@ -1,17 +1,13 @@
...
-from collector_utilities.type import URL, Job
+from collector_utilities.type import Job
 
 from .base import GitLabJobsBase
 
 
 class GitLabFailedJobs(GitLabJobsBase):
     """Collector class to get failed job counts from GitLab."""
 
-    async def _api_url(self) -> URL:
```
emdv88, line 11: Don't we need to replace this by a filter function to send only the currently failing jobs to the server? If I understand all of this correctly, now we changed 'failed jobs' to 'all jobs'

emdv88: Of course, for that we would need some way to identify jobs over multiple 'runs'. Don't know if that is possible with the gitlab api.

fniessink:
> I think the unit test fails because it will now randomly collect a measurement even if no 15 minutes have passed. The random method should probably be patched for this test.

You're correct. I've removed the randomness for now. Want to think about a better implementation.

components/collector/src/source_collectors/gitlab/failed_jobs.py
fniessink: The filter function is part of the subclass:

```python
class GitLabFailedJobs(GitLabJobsBase):
    """Collector class to get failed job counts from GitLab."""

    def _count_job(self, job: Job) -> bool:
        """Return whether the job has failed."""
        failure_types = list(self._parameter("failure_type"))
        return super()._count_job(job) and job["status"] in failure_types
```

PR #2157, opened 2021-05-11
Better measurements collection index

fniessink:


components/server/src/initialization/database.py
```diff
@@ -45,7 +45,7 @@ def create_indexes(database: Database) -> None:
     """Create any indexes."""
     database.datamodels.create_index("timestamp")
     database.reports.create_index("timestamp")
-    database.measurements.create_index("start")
```
fniessink, line 48: Should this index be dropped?

PR #2175, opened 2021-05-16
Remove individual metric trend table 

### LINKED ISSUE #2174: Remove individual metric trend table

fniessink:
Since subjects have a trend table view, there is little added value in also having a trend table per metric. Remove the trend table view for individual metrics. Closes #2174.


emdv88 (approved):
Ruimt lekker op dit!

PR #2211, opened 2021-05-27
Use Pydantic for generating the data model

fniessink:


components/server/README.md
```diff
@@ -10,315 +10,189 @@ API documentation can be retrieved via http://www.quality-time.example.org/api (
 
 ## Data model
 
-The [`datamodel.json`](src/data/datamodel.json) describes the domain model used by the application. It allows for a frontend that doesn't need to know about specific metrics and sources. Every time the server is started, the latest data model is imported into the database. The server [`data` package](src/data/datamodel.json) contains the `datamodel.json`.
+The data model package describes the domain model used by the application. It allows for a frontend that doesn't need to know about specific metrics and sources. On server start up, it checks whether the data model has changed, and if so, imports it into the database.
```
emdv88, line 13: The package is currently called "data" instead of "data model" since it also still contains the example reports. Maybe it would be a good idea to give the data model it's own package and make call the data package example_data?

components/server/README.md
```diff
@@ -10,315 +10,189 @@ API documentation can be retrieved via http://www.quality-time.example.org/api (
...
-           "description": "..."
-       },
-       "percentage": {
-           "name": "Percentage",
-           "description": "..."
-       }
-    }
-}
-```
+The `scales` part of the data model defines the scales uses. At the time of writing these include an absolute count scale, a percentage scale, and a version number scale.
```
emdv88:
```suggestion
The `scales` part of the data model defines the scales that can be used to measure metrics. At the time of writing these include an absolute count scale, a percentage scale, and a version number scale.
```

components/server/README.md
```diff
@@ -10,315 +10,189 @@ API documentation can be retrieved via http://www.quality-time.example.org/api (
...
-                "random"
-            ],
-            "tags": [
-                "maintainability",
-                "testability"
-            ]
-        }
-    }
-}
+The `metrics` part of the data model is an object where the keys are the metric types and the values are objects describing the metric type. Metrics have the following meta model:
```
emdv88:
```suggestion
The `metrics` part of the data model is a dictionary with all supported metric types. The keys are the metric type names and the values are objects describing the metric type. All metrics types have the following meta model:
```

components/server/README.md
```diff
@@ -10,315 +10,189 @@ API documentation can be retrieved via http://www.quality-time.example.org/api (
...
-                            "key": "last_update",
-                            "type": "datetime"
-                        }
-                    ]
-                }
-            }
-        }
-    }
-}
+The `sources` part of the data model is an object where the keys are the source types and the values are objects describing the source. Sources have the following meta model:
```
emdv88:
```suggestion
The `sources` part of the data model is a dictionary that describes all supported source types. The keys are the source type names and the values are objects describing the source type. All source types have the following meta model:
```

PR #2283, opened 2021-06-17
Check the up-to-dateness of the generated documentation as part of the pipeline

### LINKED ISSUE #2210: Check documentation completeness in github ci
When running the release script, the completeness of the docs/METRICS_AND_SOURCES.md is automatically checked. This check should be moved to be part of the github ci and be run on push.

fniessink:
Also run the quality checks on on the code and lint the Markdown files. Closes #2210.


docs/.pylintrc
```diff
@@ -0,0 +1,539 @@
+[MASTER]
```
emdv88: do we really need a separate pylintrc for each component (and for the docs)? Is there any difference between them?

docs/tests/test_generated_docs.py
```diff
@@ -0,0 +1,20 @@
...
+class GeneratedDocumentationTest(unittest.TestCase):
+    """Tests for the generated documentation."""
+
+    def test_up_to_dateness(self):
+        """Tests that the workspace is clean after generting the docs."""
+        docs = pathlib.Path(__file__).resolve().parent.parent
+        repo = git.Repo(docs.parent)
+        create_metrics_and_sources = docs / "src" / "create_metrics_and_sources_md.py"
+        subprocess.run(["python3", create_metrics_and_sources], check=True)  # skipcq: BAN-B603,BAN-B607
+        metrics_and_sources = docs / "METRICS_AND_SOURCES.md"
```
emdv88, line 19: up until here this isn't really a test, right? it's actually generating the docs. Wouldn't it be nicer to just call the create_metrics_and_sources_md.py from a github action?

docs/.pylintrc
fniessink: They're all the same I think. I'll replace them by one root .pylintrc.

docs/tests/test_generated_docs.py
fniessink: And automatically commit the changed docs? I think I would prefer a pre-commit hook so that the updated docs are part of the same commit as the changes that caused it.

fniessink: I've added updating the generated documentation to the pre-commit hook. You'll need to copy git-hooks/pre-commit to .git/hooks to effectuate it.

PR #2374, opened 2021-07-15
always fetch reports_overview

### LINKED ISSUE #2373: After reload report page, form fields are writable without permissions
Steps to reproduce:
 - log in as a user without edit report permission
 - navigate to a report
 - reload page
 - All form fields are now writable, but the changes will have no effect because the api correctly returns a 403

This bug occurs because in App.js the reload_reports function is never called when the page is loaded directly on a specific report.
Possible solution: in reload_report, also make a call to the reports_overview api endpoint. Possibly store that data in a context provider, along with the datamodel and the permissions.

emdv88:
Closes #2373.

### DIFF
--- a/components/frontend/src/App.js
+++ b/components/frontend/src/App.js
@@ -77,8 +77,8 @@ class App extends Component {
   }
 
   reload_report(report_date, show_error) {
-    Promise.all([get_datamodel(report_date), get_report(this.state.report_uuid, report_date)]).then(
-      ([data_model, reports]) => {
+    Promise.all([get_datamodel(report_date), get_reports(report_date), get_report(this.state.report_uuid, report_date)]).then(
+      ([data_model, reports_overview, reports]) => {
         if (data_model.ok === false || reports.ok === false) {
           show_error();
         } else {
@@ -86,6 +86,12 @@ class App extends Component {
           this.setState({
             loading: false,
             datamodel: data_model,
+            reports_overview: {
+              layout: reports_overview.layout,
+              subtitle: reports_overview.subtitle,
+              title: reports_overview.title,
+              permissions: reports_overview.permissions
+            },
             reports: reports.reports || [],
             last_update: now
           });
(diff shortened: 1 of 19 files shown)


components/frontend/src/App.js
```diff
@@ -77,15 +77,21 @@ class App extends Component {
   }
 
   reload_report(report_date, show_error) {
-    Promise.all([get_datamodel(report_date), get_report(this.state.report_uuid, report_date)]).then(
-      ([data_model, reports]) => {
+    Promise.all([get_datamodel(report_date), get_reports(report_date), get_report(this.state.report_uuid, report_date)]).then(
```
fniessink: get_reports() can be expensive because it summarizes all the reports in the Quality-time instance for which it needs to gather recent measurements from the database. It seems we're only interested in the permissions here. Isn't it better to create a separate endpoint for the permissions and use that here?

components/frontend/src/App.js
```diff
@@ -95,15 +101,21 @@ class App extends Component {
 
   reload_tag_report(report_date, show_error) {
     const tag = this.state.report_uuid.slice(4);
-    Promise.all([get_datamodel(report_date), get_tag_report(tag, report_date)]).then(
-      ([data_model, report]) => {
+    Promise.all([get_datamodel(report_date), get_reports(report_date), get_tag_report(tag, report_date)]).then(
```
fniessink: See line 80 above.

components/frontend/src/App.js
emdv88: mh... that might be an option. Then we will also get a seperate endpoint for the configured issue trackers

emdv88: I think I'd make a new endpoint "reports_overview" that will return anything that is in the document of the current reports_overview. That should not become too expensive to load, so we can always call that on a reload and then depending on what is needed we call the get_report, get_tag_report or get_reports for the rest of the info. Maybe even make that a local state in the Report or Reports component.

PR #2389, opened 2021-07-20
 Include the tests from TestNG XML reports in the measurement details of the tests metric

### LINKED ISSUE #2388: Add entities to TestNG
Needed for #2139.

fniessink:
Closes #2388.


components/collector/src/source_collectors/testng/tests.py
```diff
@@ -1,23 +1,46 @@
...
+        return SourceMeasurement(value=str(test_count), total=str(total), entities=entities)
+
+    @classmethod
+    def __entities(cls, tree: Element, test_results_to_count: list[str]) -> Entities:
+        """Transform the test methods into entities."""
+        entities = Entities()
+        for test_class in tree.findall(".//class"):
+            class_name = test_class.get("name", "")
+            for test_method in test_class.findall(".//test-method"):
+                test_result = cls.TEST_RESULT.get(test_method.get("status", ""))
```
emdv88: Is the default empty string required useful here? If I understand correctly, if it would fallback to the empty string, the TEST_RESULT.get() would raise a keyError anyway.

emdv88: Personally, in this case I would prefer the TEST_RESULT[test_method["status"]] syntax, or maybe even cleaner to define a status variable before

components/collector/src/source_collectors/testng/tests.py
```diff
@@ -1,23 +1,46 @@
...
+                if test_method.get("is-config", "false") == "true" or test_result not in test_results_to_count:
+                    continue  # Skip config (beforeClass/afterClass) methods and test results the user wants to ignore
+                name = test_method.get("name", "")
+                description = test_method.get("description", "")
+                key = f"{class_name}_{name}"
```
emdv88, line 47: Does this key have to be unique? because this could theoretically lead to a key "_" for all entities without class_name and name. In the end I'm wondering if default to empty string makes sense for class_name and name.

components/collector/src/source_collectors/testng/tests.py
fniessink: I added "" as second argument to `Element.get()` because otherwise it returns None as default value which makes the type of the `Element.get()` result `Optional[str]` instead of `str`. But I'll see if I can refactor it so that it doesn't raise this question.

components/collector/src/source_collectors/testng/tests.py
fniessink: It does have to be unique. So far from looking a testng-result.xml files it seems test class name and test method name should always be present and the combination should be unique. Unfortunately, there's no DTD for testng-result.xml, so I'm not 100% sure. I'll add some comments.

PR #2417, opened 2021-07-23
Add a test case metric 

### LINKED ISSUE #2139: Add metric(s) to relate logical test cases in Jira/Azure DevOps to executed test cases in test reports
Metric: "Test cases"

Entities are (logical) test cases, with status (based on the test report).

Parameter: statuses to include, values depend on the test report type, but probably include passed/failed/skipped/missing

This metric is different than existing metrics because it needs sources of two different types.

The assumption is that the physical test cases in the test reports will contain a reference to the logical test cases in Jira/Azure DevOps, e.g. by means of an id in the test case name.

UI:
- The metric details get multiple source tabs, one for test cases and one for test reports.

Out of scope:
- Adding the issue number to the test report. We assume that can be done through the test case name.

Later:
- Number of physical test cases without logical test case
- Extending the data model to specify what combination of sources is needed for a metric (e.g. "one_of(Jira, Azure) and one_of(Junit, Robot, etc.)").

fniessink:
The test case metric matches test cases from Jira to tests from a TestNG report, provided the test methods in the TestNG refer to the keys of Jira issues. Closes #2139.


components/collector/src/metric_collectors/test_cases.py
```diff
@@ -0,0 +1,100 @@
...
+        ("errored", "skipped"): "errored",
+        ("errored", "failed"): "errored",
+        ("errored", "errored"): "errored",
+    }
+    TEST_CASE_KEY_RE = re.compile(r"\w+\-\d+")
+    TEST_CASE_SOURCE_TYPES = ["jira"]
+    TEST_REPORT_SOURCE_TYPES = ["junit", "testng"]
+
+    async def collect(self) -> Optional[MetricMeasurement]:
+        """Override to add the rest results from the test report(s) to the test cases."""
```
emdv88:
```suggestion
        """Override to add the test results from the test report(s) to the test cases."""
```

components/collector/src/metric_collectors/test_cases.py
```diff
@@ -0,0 +1,100 @@
...
+        test_cases = self.test_cases(measurement.sources)
+        test_case_keys = set(test_cases.keys())
+        for entity in self.test_entities(measurement.sources):
+            for test_case_key in self.referenced_test_cases(entity) & test_case_keys:
+                test_result_so_far = test_cases[test_case_key]["test_result"]
+                test_result = entity["test_result"]
+                test_cases[test_case_key]["test_result"] = self.TEST_RESULT_STATE[(test_result_so_far, test_result)]
+        # Set the value of the test sources to zero as this metric only counts test cases
+        for test in self.test_sources(measurement.sources):
+            test.value = "0"
```
emdv88: Wouldn't it be more correct to remove the key value as a whole, since it is meaningless in this context. For a next refactor, I would suggest to let the server determine the value (and total) of a source in order to improve separation of concerns and let the collector only send lists of entities.

components/collector/src/metric_collectors/test_cases.py
```diff
@@ -0,0 +1,100 @@
...
+        # Filter the test cases by test result
+        for source in self.test_case_sources(measurement.sources):
```
emdv88, line 61: maybe the filtering could be incorporated in the loop s of lines 51-55? This might safe an extra loop an potentially performance.

components/collector/src/metric_collectors/test_cases.py
fniessink: Maybe in the future, but for now I want to see both the Jira logical test cases and the physical test cases from JUnit/TestNG in the frontend to make it easier to inspect the result. By setting the value of the Unit/TestNG to zero we make sure the number of physical test cases is not counted (only the logical test cases from Jira) while the physical test cases are still visible as entities.

components/collector/src/metric_collectors/test_cases.py
fniessink: There will be typically only be a handful of sources, so iterating through these twice (once to get the test case sources, i.e. Jira, and once to get the test report sources, i.e. JUnit/TestNG) won't be be a problem for performance I think.

PR #2465, opened 2021-08-02
Set the feature compatibility version to the current MongoDB version …

### LINKED ISSUE #2027: Prepare upgrade of MongoDB
Upgrade from 4.2.13 to 4.4.5 (or newer).

Check https://docs.mongodb.com/manual/release-notes/4.4-upgrade-standalone/

Note: if we need to change Quality-time and because of this change it wouldn't work with Mongodb 4.2, then this would lead to a major version update of Quality-time, because Quality-time instances without the docker-compose changes introduced in version 3.17.0 will not work anymore.

fniessink:
…to prepare for upgrade to the next version. Partially fixes #2027.

### DIFF
--- a/components/server/src/initialization/database.py
+++ b/components/server/src/initialization/database.py
@@ -21,7 +21,9 @@ from .report import import_example_reports, initialize_reports_overview
 def init_database() -> Database:  # pragma: no cover-behave
     """Initialize the database connection and contents."""
     database_url = os.environ.get("DATABASE_URL", "mongodb://root:root@localhost:27017")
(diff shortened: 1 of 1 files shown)


components/server/src/initialization/database.py
```diff
@@ -44,6 +46,14 @@ def init_database() -> Database:  # pragma: no cover-behave
     return database
 
 
+def set_feature_compatibility_version(database: Database) -> None:
```
emdv88:
```suggestion
def set_feature_compatibility_version(admin_database: Database) -> None:
```

To indicate that this not the same database as all the other database arguments that we have...

PR #2505, opened 2021-08-16
Readthedocs

fniessink:
See https://quality-time.readthedocs.io/en/latest/intro.html.


components/collector/README.md
```diff
@@ -1,23 +1,3 @@
...
-The collector uses the following environment variables:
-
-| Name | Default value | Description |
-| :--- | :---------- | :------------ |
-| SERVER_HOST | server | Hostname of the server. The collector uses this to get the metrics and post the measurements. |
-| SERVER_PORT | 5001 | Port of the server. The collector uses this to get the metrics and post the measurements. |
-| COLLECTOR_SLEEP_DURATION | 20 | The maximum amount of time (in seconds) that the collector sleeps between collecting measurements. |
-| COLLECTOR_MEASUREMENT_LIMIT | 30 | The maximum number of metrics that the collector measures each time it wakes up. If more metrics need to be measured, they will be measured the next time the collector wakes up. |
-| COLLECTOR_MEASUREMENT_FREQUENCY | 900 | The amount of time (in seconds) after which a metric should be measured again. |
+See the [software documentation](https://quality-time.readthedocs.io/en/latest/software.html#collector).
```
emdv88, line 3: Is it (without too much hassle) possible too keep some README's or developers documentation in the repository itself? Just because that is usually the first place for me to look, so it would save me clicking on one link. Of course this should only be done if we can use the same text in the new readthedocs documentation aswell.

components/proxy/README.md
```diff
@@ -1,3 +1,3 @@
 # *Quality-time* proxy
 
-Wrap the Caddy image in a *Quality-time* image so the Caddy version number can be changed when needed.
\ No newline at end of file
+See the [software documentation](https://quality-time.readthedocs.io/en/latest/software.html#proxy).
```
emdv88, line 3: On the other hand, I think this README, as well as the one for the renderer might aswell be deleted... The linked information is only a single line.

docs/src/changelog.md
```diff
@@ -215,7 +216,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 ### Added
 
 - When using Jira as source for the 'issues' and the 'user story points' metric, show the issue type in the metric details. Closes [#1674](https://github.com/ICTU/quality-time/issues/1674).
-- Allow for limiting editing rights to specific people. Grant editing rights to people by adding their username or email address to the editors field on the homepage. Expand the overview title to access the editors field. Also see the [user manual](USAGE.md#limiting-editing-rights). Closes [#294](https://github.com/ICTU/quality-time/issues/294).
+- Allow for limiting editing rights to specific people. Grant editing rights to people by adding their username or email address to the editors field on the homepage. Expand the overview title to access the editors field. Also see [](usage.md#configuring-permissions). Closes [#294](https://github.com/ICTU/quality-time/issues/294).
```
emdv88: a link without text, is it still clickable?

components/collector/README.md
fniessink: All the source docs are still there, but moved to docs/src. 

It is possible to include Markdown files from outside the docs folder, but the disadvantage is that you still need a file in the docs folder that does an include, for example `{include} ../../components/collector/README.md`. So I thought would be easier to move everything in the docs folder. I left pointers to the relevant docs in each of the component's README.md to make it relatively easy to navigate to the appropriate spot in the documentation.

components/proxy/README.md
fniessink: Yeah, but it is a deep link to the documentation about the component, in this case the proxy, so that if you're browsing the source code it's easy to open the accompanying docs. Admittedly, not as easy as having the docs right there...

docs/src/changelog.md
fniessink: It works in VS Code, not on GitHub unfortunately. The advantage of this way of linking is that Sphinx uses the section title from the linked section to create the link in the document; less duplication. But seeing it doesn't work on GitHub, I'll add a link text.

PR #2508, opened 2021-08-18
The security warnings in OWASP ZAP reports do not have unique keys. H…

### LINKED ISSUE #2429: Security warning status and Rationale disappear from LRK QT
Security warning status and Rationale seems to disappear. On 22-07-2021 both fields were filled in. For example at 22-07-2021 with subject KOT&B app => OWASP ZAP warnings => Warning: CSP: style-src unsafe-inline things were looking like this:
image
And at 26-07-2021 the same subject (after new scans at 24-07-2021 with the same results) it looks like this:
image

Example with more detail of what's missing:
On 22-07-2021:
image
On 26-07-2021:
image

Can these fields made persistent with recurring warnings?
http://www.quality-time.lrk.ictu/34d9b1f0-c8b3-4689-94d8-23f2ec20dc6b

fniessink:
…owever, *Quality-time* needs security warnings to be uniquely identifiable to detect whether the list of warnings changes between measurements. Therefore, *Quality-time* generates keys for OWASP ZAP security warnings itself. Unfortunately, the key that *Quality-time* generated, was not guaranteed to be unique. Fixes #2429.

### DIFF
--- a/components/collector/src/source_collectors/owasp_zap/security_warnings.py
+++ b/components/collector/src/source_collectors/owasp_zap/security_warnings.py
@@ -40,8 +40,8 @@ class OWASPZAPSecurityWarnings(XMLFileSourceCollector):
         """Create an alert instance entity."""
         method = alert_instance.findtext("method", default="")
         uri = self.__stable_url(hashless(URL(alert_instance.findtext("uri", default=""))))
-        key = md5_hash(f"{':'.join(ids)}:{method}:{uri}")
-        old_key = md5_hash(f"{':'.join(ids[1:])}:{method}:{uri}")
+        key = md5_hash(f"{':'.join(ids[:-1])}:{method}:{uri}")
+        old_key = md5_hash(f"{':'.join(ids)}:{method}:{uri}")
         location = f"{method} {uri}"
         return Entity(key=key, old_key=old_key, uri=uri, location=location, **entity_kwargs)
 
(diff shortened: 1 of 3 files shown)


components/collector/src/source_collectors/owasp_zap/security_warnings.py
```diff
@@ -40,8 +40,8 @@ def __alert_instance_entity(self, ids, entity_kwargs, alert_instance) -> Entity:
         """Create an alert instance entity."""
         method = alert_instance.findtext("method", default="")
         uri = self.__stable_url(hashless(URL(alert_instance.findtext("uri", default=""))))
-        key = md5_hash(f"{':'.join(ids)}:{method}:{uri}")
-        old_key = md5_hash(f"{':'.join(ids[1:])}:{method}:{uri}")
+        key = md5_hash(f"{':'.join(ids[:-1])}:{method}:{uri}")
+        old_key = md5_hash(f"{':'.join(ids)}:{method}:{uri}")
```
emdv88, line 44: Why are you adding the alert to the old key?

fniessink: This allows Quality-time to keep track of the entity's status even though its key changed when a new measurement is added. https://github.com/ICTU/quality-time/blob/d5f528f06680f06e898265be909073ff1bb4d280/components/server/src/model/source.py#L35-L42

PR #2697, opened 2021-10-07
1872 datamodel context

emdv88:


components/frontend/src/App.js
```diff
@@ -19,6 +19,7 @@ import { nr_measurements_api } from './api/measurement';
 import { login } from './api/auth';
 import { show_message, show_connection_messages } from './widgets/toast';
 import { getUserPermissions, isValidDate_YYYYMMDD } from './utils'
+import { DataModel } from './context/Contexts';
```
fniessink: Why is the file called `context/Contexts.js` when only the data model is in there and we also have a `context/Permissions.js`? Doesn't it make more sense to call the file `context/DataModel.js`?

components/frontend/src/metric/Metric.js
```diff
@@ -10,9 +10,12 @@ import { SourceStatus } from './SourceStatus';
 import { StatusIcon } from './StatusIcon';
 import { TrendSparkline } from './TrendSparkline';
 import "./Metric.css";
+import { DataModel } from '../context/Contexts';
 
-function MeasurementValue({ metric, metric_type, metric_unit, latest_measurement }) {
-    const value = metric.value && metric_type.unit === "minutes" && metric.scale !== "percentage" ? format_minutes(metric.value) : metric.value || "?";
+function MeasurementValue({ metric, metric_unit, latest_measurement }) {
+    const dataModel = useContext(DataModel)
```
fniessink, line 16: No ";" at the end of the line. We should really introduce something like black for JS :-)

components/frontend/src/metric/Metric.js
```diff
@@ -25,31 +28,34 @@ function MeasurementValue({ metric, metric_type, metric_unit, latest_measurement
...
 
-function MeasurementSources({ datamodel, metric, latest_measurement }) {
+function MeasurementSources({ metric, latest_measurement }) {
     const sources = latest_measurement?.sources ?? [];
-    return sources.map((source, index) => [index > 0 && ", ", <SourceStatus key={source.source_uuid} source_uuid={source.source_uuid}
-        metric={metric} source={source} datamodel={datamodel} />])
+    const source_statusses = sources.map(
+        (source, index) => [index > 0 && ", ", <SourceStatus key={source.source_uuid} source_uuid={source.source_uuid} metric={metric} measurement_source={source} />]
+    );
+    return source_statusses
```
fniessink:
```suggestion
    const sourceStatuses = sources.map(
        (source, index) => [index > 0 && ", ", <SourceStatus key={source.source_uuid} source_uuid={source.source_uuid} metric={metric} measurement_source={source} />]
    );
    return sourceStatuses
```

components/frontend/src/context/Contexts.js
```diff
@@ -0,0 +1,3 @@
+import React from "react";
+
+export const DataModel = React.createContext(null);
```
fniessink:
```suggestion
export const DataModel = React.createContext(null);

```

components/frontend/src/trend_table/MeasurementsRow.js
```diff
@@ -3,11 +3,12 @@ import { TableRowWithDetails } from '../widgets/TableRowWithDetails';
 import { MetricDetails } from '../metric/MetricDetails';
 import { formatMetricScale, formatMetricUnit, formatMetricScaleAndUnit, format_minutes, get_metric_name } from "../utils";
 import './TrendTable.css';
+import { useContext } from "react";
```
fniessink, line 6: Hmm, interesting that we don't import React here and it still works.

fniessink (changes requested):

components/frontend/src/metric/Metric.js
fniessink: But not as part of this PR of course.

components/frontend/src/App.js
emdv88: True. I left the Permissions.js separate because it has it's own consumer component that is specified there. 
But the DataModel context is literally just a single line of code, so I think it doesn't make sense to create a file only for the DataModel. As it is currently I would define the DataModel Context in the App.js above the main class definition.

But finally, I think that it would be good to start using more context providers, for example for report, subject, metric, source, maybe more? so then it could make sense to put them all together in this Contexts.js file.

components/frontend/src/metric/Metric.js
emdv88: I'm sure this is not the only spot. I think eslint can do something like that...

components/frontend/src/trend_table/MeasurementsRow.js
emdv88: indeed it's kinda weird. I always thought that importing react was necessary if JSX is used in the file...

components/frontend/src/App.js
fniessink: Clear. Since we don't have any other context providers (yet), I've renamed Contexts.js to DataModel.js. We can always change it back.

components/frontend/src/metric/Metric.js
fniessink: I'll take a look.

fniessink (approved):

PR #2732, opened 2021-10-14
Give metric.get('sources') {} as default return value 

fniessink:
This ensures that calling values() on the return value of get() can't fail.

### DIFF
--- a/components/collector/src/base_collectors/collector.py
+++ b/components/collector/src/base_collectors/collector.py
@@ -143,7 +143,7 @@ class Collector:
 
     def __can_collect(self, metric) -> bool:
         """Return whether the user has specified all mandatory parameters for all sources."""
-        sources = metric.get("sources")
+        sources = metric.get("sources", {})
         for source in sources.values():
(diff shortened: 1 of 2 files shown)


components/collector/src/base_collectors/collector.py
```diff
@@ -143,7 +143,7 @@ def __can_and_should_collect(self, metric_uuid: str, metric) -> bool:
 
     def __can_collect(self, metric) -> bool:
```
emdv88: as discussed yesterday, adding typehints to this function could help mypy find these kind of potential bugs

fniessink: You're right. If I remove the default return value of `{}` and I tell mypy that metric has type `dict`:

```python
    def __can_collect(self, metric: dict) -> bool:
        """Return whether the user has specified all mandatory parameters for all sources."""
        sources = metric.get("sources")
        for source in sources.values():
            ...
```

I get an error:

```
src/base_collectors/collector.py:147: error: Item "None" of "Optional[Any]" has no attribute "values"
```

Hooray for type hints.

PR #2736, opened 2021-10-16
UI improvements

fniessink:


components/frontend/src/subject/Subjects.test.js
```diff
@@ -177,8 +177,9 @@ describe("<Subjects />", () => {
     });
     it('toggles subject trend table', () => {
         const wrapper = subjects();
+        console.log(wrapper.debug());
```
emdv88: remove console log?

emdv88 (changes requested):
Only the console log in the unittest

components/frontend/src/subject/Subjects.test.js
fniessink: Oopsie

PR #2747, opened 2021-10-21
2692 latest measurements

emdv88:


emdv88:
You'll see in the changes that I had to add a whole lot of wait seconds to the feature tests. This is because the get_latest_measurements_by_uuid will return the wrong latest measurement if the start time is the same (within the one second accuracy). This happened to accidentally go the right way in the old get_recent_measurements_by_uuid as long as the order in which they were inserted into the database was chronological. Maybe we should fix this "fallback" that the old method contained keeps working?

components/server/src/database/measurements.py
```diff
@@ -1,6 +1,7 @@
 """Measurements collection."""
 
 from datetime import datetime, timedelta
+from typing import Dict, List
```
fniessink: This isn't necessary anymore. You can use dict and list.
```suggestion
```

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,36 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
     return None if latest is None else Measurement(metric, latest)
 
 
+def latest_measurements_by_metric_uuid(database: Database, metric_uuids: List[str]) -> Dict[str, Measurement] | None:
```
fniessink:
```suggestion
def latest_measurements_by_metric_uuid(database: Database, metric_uuids: list[str]) -> dict[str, Measurement] | None:
```

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,36 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
...
+    """Return the latest measurements in a dict with metric_uuids as keys."""
+    latest_measurements = database.measurements.find(
+        {"metric_uuid": {"$in": metric_uuids}},
+        sort=[("start", pymongo.ASCENDING)],
+        projection={"_id": False, "sources.entities": False, "entity_user_data": False},
+    )
+    latest_measurements_by_uuid = {measurement["metric_uuid"]: measurement for measurement in latest_measurements}
+    return latest_measurements_by_uuid
```
fniessink:
```suggestion
    return {measurement["metric_uuid"]: measurement for measurement in latest_measurements}
```

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,36 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
     return None if latest is None else Measurement(metric, latest)
 
 
+def latest_measurements_by_metric_uuid(database: Database, metric_uuids: List[str]) -> Dict[str, Measurement] | None:
+    """Return the latest measurements in a dict with metric_uuids as keys."""
+    latest_measurements = database.measurements.find(
+        {"metric_uuid": {"$in": metric_uuids}},
+        sort=[("start", pymongo.ASCENDING)],
+        projection={"_id": False, "sources.entities": False, "entity_user_data": False},
+    )
```
fniessink: This retrieves *all* measurements for the selected metric_uuids and then the dict comprehension on the next line keeps only the most recent one for each metric_uuid. That's not immediately obvious. And it retrieves much more data from the database than we need. Can't we make MongoDB do that work by passing a more restrictive filter?

For reports, we use a last flag. That's one way to solve this I guess. In that case, don't forget to update the index.

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,36 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
...
-    """Return all recent measurements."""
+def recent_measurements_by_metric_uuid(database: Database, max_iso_timestamp: str = "", days=7, metric_uuids=None):
+    """Return all recent measurements, or only those of the specified metrics."""
     max_iso_timestamp = max_iso_timestamp or iso_timestamp()
     min_iso_timestamp = (datetime.fromisoformat(max_iso_timestamp) - timedelta(days=days)).isoformat()
+    filter_dict = {"end": {"$gte": min_iso_timestamp}, "start": {"$lte": max_iso_timestamp}}
+    # metric_uuids needs to be optional as long as the /reports endpoint is supported for backwards compatibility
+    # however, there is no test anymore covering that endpoint, which means incomplete coverage on this if statement
+    if metric_uuids is not None:  # pragma: no cover
+        filter_dict["metric_uuid"] = {"$in": metric_uuids}
```
fniessink:
```suggestion
    measurement_filter = {"end": {"$gte": min_iso_timestamp}, "start": {"$lte": max_iso_timestamp}}
    # metric_uuids needs to be optional as long as the /reports endpoint is supported for backwards compatibility
    # however, there is no test anymore covering that endpoint, which means incomplete coverage on this if statement
    if metric_uuids is not None:  # pragma: no cover
        measurement_filter["metric_uuid"] = {"$in": metric_uuids}
```

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,36 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
...
     recent_measurements = database.measurements.find(
-        filter={"end": {"$gte": min_iso_timestamp}, "start": {"$lte": max_iso_timestamp}},
+        filter_dict,
```
fniessink:
```suggestion
        measurement_filter,
```

components/server/src/model/transformations.py
```diff
@@ -6,8 +6,16 @@
 from datetime import date
 from json.decoder import JSONDecodeError
 from typing import cast
-
-from server_utilities.functions import DecryptionError, asymmetric_decrypt, asymmetric_encrypt, unique, uuid
+from database.measurements import latest_measurements_by_metric_uuid, recent_measurements_by_metric_uuid
+
```
fniessink:
```suggestion
```

components/server/src/server_utilities/functions.py
```diff
@@ -183,3 +183,11 @@ def asymmetric_decrypt(private_key: str, fernet_key_message: Tuple[str, str]) ->
     message = symmetric_decrypt(decrypted_key, message_bytes)
 
     return message.decode()
+
+
+def report_metrics_uuids(report: Dict) -> List[str]:
```
fniessink:
```suggestion
def report_metrics_uuids(report: dict) -> list[str]:
```

components/server/src/server_utilities/functions.py
```diff
@@ -7,7 +7,7 @@
 from collections.abc import Callable, Hashable, Iterable, Iterator
 from datetime import datetime, timezone
 from decimal import ROUND_HALF_UP, Decimal
-from typing import Tuple, TypeVar, cast
+from typing import Dict, List, Tuple, TypeVar, cast
```
fniessink:
```suggestion
from typing import Tuple, TypeVar, cast
```

(Tuple can be removed as well...)

components/server/tests/database/test_measurements.py
```diff
@@ -26,19 +30,19 @@ def setUp(self):
             {"start": "8", "end": "9", "metric_uuid": METRIC_ID3},
         ]
 
-        def find_one_side_effect(query, projection, sort=None):
+        def find_one_side_effect(query, projection=None, sort=None):  # pylint: disable=unused-argument
```
fniessink: Why is the disable unused-argument needed? It looks like all three arguments are used.

components/server/tests/database/test_measurements.py
```diff
@@ -67,3 +71,32 @@ def test_get_timestamp_restriction(self):
         for measurement in measurements:
             self.assertEqual(measurement["metric_uuid"], METRIC_ID)
             self.assertIn(measurement["start"], ["0", "3"])
+
+    def test_recent_measurements_by_uuid(self):
+        """Test that we get all measurements with all metric id's."""
```
fniessink:
```suggestion
        """Test that we get all measurements with all metric ids."""
```

components/server/tests/database/test_measurements.py
```diff
@@ -67,3 +71,32 @@ def test_get_timestamp_restriction(self):
...
+        self.assertEqual(len(recent_measurements), 3)
+        self.assertIn(METRIC_ID, recent_measurements)
+        self.assertEqual(len(recent_measurements[METRIC_ID]), 3)
+        self.assertIn(METRIC_ID2, recent_measurements)
+        self.assertEqual(len(recent_measurements[METRIC_ID2]), 3)
+        self.assertIn(METRIC_ID3, recent_measurements)
+        self.assertEqual(len(recent_measurements[METRIC_ID3]), 3)
+
+    def test_recent_measurements_by_uuid_uuid_filter(self):
+        """Test that we get all measurements with all metric id's."""
```
fniessink:
```suggestion
        """Test that we get all measurements with all metric ids."""
```

components/server/tests/database/test_measurements.py
```diff
@@ -67,3 +71,32 @@ def test_get_timestamp_restriction(self):
...
+        recent_measurements = recent_measurements_by_metric_uuid(self.database, metric_uuids=[METRIC_ID, METRIC_ID2])
+        self.assertEqual(len(recent_measurements), 2)
+        self.assertIn(METRIC_ID, recent_measurements)
+        self.assertEqual(len(recent_measurements[METRIC_ID]), 3)
+        self.assertIn(METRIC_ID2, recent_measurements)
+        self.assertEqual(len(recent_measurements[METRIC_ID2]), 3)
+
+    def test_latest_measurements_by_uuid_uuid_filter(self):
+        """Test that we get all measurements with all metric id's."""
```
fniessink:
```suggestion
        """Test that we get all measurements with all metric ids."""
```

docs/src/changelog.md
```diff
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Changed
 
+- Add a separate "latest_measurent" key to each metric in the /report endpoint. Closes [#2692](https://github.com/ICTU/quality-time/issues/2692).
+- Add a parameter metric_uuids to recent_measurements_by_metric_uuid to minimize the database load. Closes [#2695](https://github.com/ICTU/quality-time/issues/2695).
```
fniessink: As the changelog is mainly aimed at end users, I'm not sure it's useful to mention technical details like that here. On the other hand, it does make it easier for users to report bugs if they can relate buggy behavior to changes. So maybe:
```suggestion
- Performance improvements. Closes [#2692](https://github.com/ICTU/quality-time/issues/2692) and [#2695](https://github.com/ICTU/quality-time/issues/2695).
```

fniessink (changes requested):

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,36 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
...
 
 
+def latest_measurements_by_metric_uuid(database: Database, metric_uuids: List[str]) -> Dict[str, Measurement] | None:
+    """Return the latest measurements in a dict with metric_uuids as keys."""
+    latest_measurements = database.measurements.find(
+        {"metric_uuid": {"$in": metric_uuids}},
+        sort=[("start", pymongo.ASCENDING)],
+        projection={"_id": False, "sources.entities": False, "entity_user_data": False},
+    )
+    latest_measurements_by_uuid = {measurement["metric_uuid"]: measurement for measurement in latest_measurements}
```
fniessink: If multiple measurements for the same metric are within one second of each other, this does not return the most recent measurement.

components/server/src/database/measurements.py
```diff
@@ -17,20 +18,43 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
     return None if latest is None else Measurement(metric, latest)
 
 
+def latest_measurements_by_metric_uuid(database: Database, metric_uuids: List[str]) -> Dict[str, Measurement] | None:
+    """Return the latest measurements in a dict with metric_uuids as keys."""
+    latest_measurement_ids = database.measurements.aggregate(
+        [
+            {"$match": {"metric_uuid": {"$in": metric_uuids}}},
+            {"$sort": {"metric_uuid": 1, "start": -1}},
+            {"$group": {"_id": "$metric_uuid", "measurement_id": {"$first": "$_id"}}},
```
fniessink, line 26: Is an alternative to copy all measurement fields (instead of just the measurement_id) as part of the `$group` step? Something like:
```suggestion
            {"$group": {"_id": "$metric_uuid", "metric_uuid": {"$first": "$metric_uuid"}, "sources": {"$first": "$sources"}, "start": {"$first": "$start"}, "end": {"$first": "$end"}, etc...},
```

That way the separate find isn't necessary and the result of the aggregation would basically be a list of measurements. A disadvantage of this approach is that the measurement fields need to be listed explicitly.

PR #2767, opened 2021-10-27
Refactor frontend

fniessink:
Several smallish refactorings, mainly to reduce the number of props passed around.

### DIFF
--- a/components/frontend/src/metric/MetricDetails.js
+++ b/components/frontend/src/metric/MetricDetails.js
@@ -116,10 +116,9 @@ export function MetricDetails({
                     <Sources
                         reports={reports}
                         report={report}
+                        metric={metric}
                         metric_uuid={metric_uuid}
-                        metric_type={metric.type}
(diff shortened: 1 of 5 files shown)


components/frontend/src/metric/MetricDetails.js
```diff
@@ -116,10 +116,9 @@ export function MetricDetails({
                     <Sources
                         reports={reports}
                         report={report}
+                        metric={metric}
                         metric_uuid={metric_uuid}
-                        metric_type={metric.type}
                         metric_unit={metric_unit}
```
emdv88, line 121: Are the metric_uuid and metric_unit here still necessary as separate input arguments next to the metric?

fniessink: The `metric_uuid` is needed for the add/copy/move source buttons. The `metric_unit` probably not, but because it is passed across quite a few components I want to tackle that one in a different PR.

PR #2776, opened 2021-10-28
this does not seem to break anything?

emdv88:


components/server/src/database/measurements.py
```diff
@@ -51,10 +53,12 @@ def recent_measurements_by_metric_uuid(database: Database, max_iso_timestamp: st
     # however, there is no test anymore covering that endpoint, which means incomplete coverage on this if statement
     if metric_uuids is not None:  # pragma: no cover
         measurement_filter["metric_uuid"] = {"$in": metric_uuids}
+    projection = {"_id": False, "metric_uuid": True, "start": True, "end": True}
+    projection.update({scale + ".value": True for scale in data_model.get("scales", {}).keys()})
```
fniessink, line 57: Why is this needed?

components/server/src/routes/external/reports_overview.py
```diff
@@ -28,7 +28,7 @@ def get_reports(database: Database):  # pragma: no cover
     data_model = latest_datamodel(database, date_time)
     overview = latest_reports_overview(database, date_time)
     overview["reports"] = []
-    recent_measurements = recent_measurements_by_metric_uuid(database, date_time)
+    recent_measurements = recent_measurements_by_metric_uuid(latest_datamodel(database), database, date_time)
```
fniessink: Why not: 
```suggestion
    recent_measurements = recent_measurements_by_metric_uuid(data_model, database, date_time)
```

fniessink (changes requested):

components/server/src/database/measurements.py
emdv88: Because the actual measured value is stored inside the dictionary with the scale name as key:
```
{
    ...
    <scale>: {
        "value":  5
    }
    ...
}
```

PR #2814, opened 2021-11-03
Clean up the measurements

### LINKED ISSUE #2639: Analyse /api/v3/report endpoint for unneeded data
Don't send information to the frontend that isn't even used:

- [ ] `/api/v3/reports`:
  - [x] #2692
  - [x] #2693
  - [x] #2694
  - [x] #2695

Cleanup:
- [ ] Make database functions return model class instances (Metric, Measurement)
- [ ] Fix typing
- [ ] Remove quasi duplicate database functions

fniessink:
Closes #2639.


emdv88:
Note to self: the measurement.summarize function seems to not work yet.

emdv88:
About performance: for the same reports overview performance seems to have improved from 3 seconds to 1.1 seconds for the reports endpoint. This is however with the perliminary, since not all functionality is present yet.

PR #2959, opened 2021-12-01
Time travel was broken

### LINKED ISSUE #2958: Time travel seems broken
When entering a past date, the measurement values shown are still the current values. The trend table seems to work ok.

fniessink:
Quality-time would show the current measurement value regardless of the date selected. Fixes #2958.

### DIFF
--- a/components/server/src/database/measurements.py
+++ b/components/server/src/database/measurements.py
@@ -1,6 +1,7 @@
 """Measurements collection."""
 
 from datetime import datetime, timedelta
+from typing import Union
 
 import pymongo
 from pymongo.database import Database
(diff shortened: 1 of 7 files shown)


components/server/src/database/measurements.py
```diff
@@ -17,11 +18,18 @@ def latest_measurement(database: Database, metric: Metric) -> Measurement | None
...
+def latest_measurements_by_metric_uuid(
+    database: Database, date_time: str, metric_uuids: list[str]
+) -> dict[str, Measurement] | None:
     """Return the latest measurements in a dict with metric_uuids as keys."""
+    metric_uuid_match: MatchType = {"metric_uuid": {"$in": metric_uuids}}
+    date_time_match: MatchType = {"start": {"$lte": date_time}} if date_time else {}
     latest_measurement_ids = database.measurements.aggregate(
         [
-            {"$match": {"metric_uuid": {"$in": metric_uuids}}},
+            {"$match": metric_uuid_match | date_time_match},
```
emdv88: Does this do the same as the dict.update() function?

fniessink: Yup, new Python 3.9 syntax: https://www.python.org/dev/peps/pep-0584/

PR #3064, opened 2021-12-16
2926 subject comments

### LINKED ISSUE #2926: Allow for adding a comment to a subject in a quality report
if would be great to be able to be able to toggle the visibility on or off.

fniessink:
Closes #2926.


components/frontend/src/subject/SubjectTitle.js
```diff
@@ -45,6 +46,15 @@ function SubjectTypeAndName({ subject, subject_uuid, subject_name, reload }) {
                     />
```
emdv88, line 46: Maybe rename this component to "configuration" instead "SubjectTypeAndName" as it is now more than just the type and the name? (on line 27 of this file, where, weirdly, I can not place a comment)

fniessink: Yes, good suggestion.

PR #3177, opened 2022-01-06
add report_uuid to internal_metrics endpoint

### LINKED ISSUE #2972: Include report_uuid in /internal-api/metrics endpoint
This allows the collector to include the report_uuid in the post to /internal-api/measurements, which in turn makes it unnecessary for the server to iterate over all reports to find the report that has the metric_uuid for which the measurement is posted, rendering database/reports.py:latest_metric obsolete.

emdv88:
Closes #2972.

### DIFF
--- a/components/collector/src/base_collectors/collector.py
+++ b/components/collector/src/base_collectors/collector.py
@@ -132,7 +132,7 @@ class Collector:
         metric_collector = metric_collector_class(session, metric, self.data_model)
         if measurement := await metric_collector.collect():
             measurement.metric_uuid = metric_uuid
-            measurement.report_uuid = metric.get("report_uuid", "")
+            measurement.report_uuid = metric["report_uuid"]
             api_url = URL(f"{self.server_url}/internal-api/{self.API_VERSION}/measurements")
             await post(session, api_url, measurement.as_dict())
 
--- a/components/collector/src/base_collectors/collector.py
+++ b/components/collector/src/base_collectors/collector.py
@@ -132,6 +132,7 @@ class Collector:
         metric_collector = metric_collector_class(session, metric, self.data_model)
         if measurement := await metric_collector.collect():
             measurement.metric_uuid = metric_uuid
+            measurement.report_uuid = metric["report_uuid"]
             api_url = URL(f"{self.server_url}/internal-api/{self.API_VERSION}/measurements")
             await post(session, api_url, measurement.as_dict())
 
--- a/components/collector/src/base_collectors/collector.py
+++ b/components/collector/src/base_collectors/collector.py
@@ -132,7 +132,7 @@ class Collector:
         metric_collector = metric_collector_class(session, metric, self.data_model)
         if measurement := await metric_collector.collect():
             measurement.metric_uuid = metric_uuid
-            measurement.report_uuid = metric["report_uuid"]
+            measurement.report_uuid = metric.get("report_uuid", "")
             api_url = URL(f"{self.server_url}/internal-api/{self.API_VERSION}/measurements")
             await post(session, api_url, measurement.as_dict())
 
--- a/components/collector/tests/base_collectors/test_collector.py
+++ b/components/collector/tests/base_collectors/test_collector.py
@@ -111,7 +111,7 @@ class CollectorTest(unittest.IsolatedAsyncioTestCase):
             await self._fetch_measurements(mock_async_get_request)
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid"),
+            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid", report_uuid=""),
         )
 
     async def test_fetch_with_empty_sources(self):
@@ -164,7 +164,12 @@ class CollectorTest(unittest.IsolatedAsyncioTestCase):
             )
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=True, sources=[self._source(connection_error="error")], metric_uuid="metric_uuid"),
+            json=dict(
+                has_error=True,
+                sources=[self._source(connection_error="error")],
+                metric_uuid="metric_uuid",
+                report_uuid="",
+            ),
         )
 
     async def test_fetch_with_empty_client_error(self):
@@ -180,7 +185,10 @@ class CollectorTest(unittest.IsolatedAsyncioTestCase):
         post.assert_called_once_with(
             self.measurement_api_url,
             json=dict(
-                has_error=True, sources=[self._source(connection_error="ClientPayloadError")], metric_uuid="metric_uuid"
+                has_error=True,
+                sources=[self._source(connection_error="ClientPayloadError")],
+                metric_uuid="metric_uuid",
+                report_uuid="",
             ),
         )
 
@@ -192,7 +200,7 @@ class CollectorTest(unittest.IsolatedAsyncioTestCase):
             await self._fetch_measurements(mock_async_get_request)
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid"),
+            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid", report_uuid=""),
         )
 
     @patch("asyncio.sleep", Mock(side_effect=RuntimeError))
@@ -205,7 +213,7 @@ class CollectorTest(unittest.IsolatedAsyncioTestCase):
             await quality_time_collector.collect()
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid"),
+            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid", report_uuid=""),
         )
 
     async def test_fetch_twice(self):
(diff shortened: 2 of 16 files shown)


components/collector/src/base_collectors/collector.py
```diff
@@ -132,6 +132,7 @@ async def collect_metric(
         metric_collector = metric_collector_class(session, metric, self.data_model)
         if measurement := await metric_collector.collect():
             measurement.metric_uuid = metric_uuid
+            measurement.report_uuid = metric.get("report_uuid", "")
```
fniessink: Why the `get("report_uuid)`? The collector gets the metrics from the Internal endpoint `get_metrics` so the metrics are guaranteed to have a report_uuid attribute?

components/collector/tests/base_collectors/test_collector.py
```diff
@@ -111,7 +111,7 @@ async def _parse_source_responses(self, responses: SourceResponses) -> SourceMea
             await self._fetch_measurements(mock_async_get_request)
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid"),
+            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid", report_uuid=""),
```
fniessink: Why doesn't the json contain a report_uuid?

components/collector/tests/base_collectors/test_collector.py
```diff
@@ -164,7 +164,12 @@ async def test_fetch_with_client_error(self):
             )
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=True, sources=[self._source(connection_error="error")], metric_uuid="metric_uuid"),
+            json=dict(
+                has_error=True,
+                sources=[self._source(connection_error="error")],
+                metric_uuid="metric_uuid",
+                report_uuid="",
```
fniessink: Why doesn't the json contain a report_uuid?

components/collector/tests/base_collectors/test_collector.py
```diff
@@ -180,7 +185,10 @@ async def test_fetch_with_empty_client_error(self):
         post.assert_called_once_with(
             self.measurement_api_url,
             json=dict(
-                has_error=True, sources=[self._source(connection_error="ClientPayloadError")], metric_uuid="metric_uuid"
+                has_error=True,
+                sources=[self._source(connection_error="ClientPayloadError")],
+                metric_uuid="metric_uuid",
+                report_uuid="",
```
fniessink: Why doesn't the json contain a report_uuid?

components/collector/tests/base_collectors/test_collector.py
```diff
@@ -192,7 +200,7 @@ async def test_fetch_with_post_error(self):
             await self._fetch_measurements(mock_async_get_request)
         post.assert_called_once_with(
             self.measurement_api_url,
-            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid"),
+            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid", report_uuid=""),
```
fniessink: Why doesn't the json contain a report_uuid? Same for all the test cases below without a report_uuid.

components/server/src/external/database/reports.py
```diff
@@ -102,11 +102,12 @@ def _prepare_documents_for_insertion(
...
-def latest_report(database: Database, data_model, report_uuid: str) -> Report:
+def latest_report(database: Database, data_model, report_uuid: str) -> Report | None:
     """Get latest report with this uuid."""
-    return Report(
-        data_model, database.reports.find_one({"report_uuid": report_uuid, "last": True, "deleted": DOES_NOT_EXIST})
-    )
+    report_dict = database.reports.find_one({"report_uuid": report_uuid, "last": True, "deleted": DOES_NOT_EXIST})
+    if report_dict:
+        return Report(data_model, report_dict)
+    return None
```
fniessink:
```suggestion
    return Report(data_model, report_dict) if report_dict else None
```

components/server/src/internal/database/reports.py
```diff
@@ -16,10 +16,18 @@ def latest_reports(database: Database, data_model: dict) -> list[Report]:
     return [Report(data_model, report_dict) for report_dict in report_dicts]
 
 
-def latest_metric(database: Database, metric_uuid: MetricId) -> Metric | None:
+def latest_report(database: Database, data_model, report_uuid: str) -> Report | None:
+    """Get latest report with this uuid."""
+    report_dict = database.reports.find_one({"report_uuid": report_uuid, "last": True, "deleted": DOES_NOT_EXIST})
+    if report_dict:
+        return Report(data_model, report_dict)
+    return None
```
fniessink:
```suggestion
    return Report(data_model, report_dict) if report_dict else None
```

components/server/src/internal/database/reports.py
```diff
@@ -16,10 +16,18 @@ def latest_reports(database: Database, data_model: dict) -> list[Report]:
...
+def latest_metric(database: Database, report_uuid: ReportId, metric_uuid: MetricId) -> Metric | None:
     """Return the latest metric with the specified metric uuid."""
     data_model = latest_datamodel(database)
-    for report in latest_reports(database, data_model):
-        if metric_uuid in report.metric_uuids:
-            return report.metrics_dict[metric_uuid]
+    report = latest_report(database, data_model, report_uuid)
+    if report and metric_uuid in report.metric_uuids:
+        return report.metrics_dict[metric_uuid]
     return None
```
fniessink:
```suggestion
    return report.metrics_dict.get(metric_uuid) if report else None
```

components/server/src/internal/routes/measurement.py
```diff
@@ -15,7 +15,8 @@ def post_measurement(database: Database) -> None:
     """Put the measurement in the database."""
     measurement_data = dict(bottle.request.json)
     metric_uuid = measurement_data["metric_uuid"]
-    if (metric := latest_metric(database, metric_uuid)) is None:
+    report_uuid = measurement_data.get("report_uuid", "")
```
fniessink: Shouldn't the metric be guaranteed to have a report_uuid, making the get unnecessary?

components/server/tests/internal/database/test_reports.py
```diff
@@ -37,10 +37,14 @@ def test_latest_metrics(self):
         ]
         self.assertEqual(
             Metric(self.data_model, dict(tags=[], type="metric_type"), METRIC_ID),
-            latest_metric(self.database, METRIC_ID),
+            latest_metric(self.database, "report_uuid", METRIC_ID),
```
fniessink:
```suggestion
            latest_metric(self.database, REPORT_ID, METRIC_ID),
```

components/server/tests/internal/database/test_reports.py
```diff
@@ -37,10 +37,14 @@ def test_latest_metrics(self):
...
     def test_no_latest_metrics(self):
         """Test that None is returned for missing metrics."""
-        self.database.measurements.find.return_value = []
-        self.assertEqual(None, latest_metric(self.database, MetricId("non-existing")))
+        self.assertIsNone(latest_metric(self.database, "report_uuid", MetricId("non-existing")))
+
+    def test_no_latest_report(self):
+        """Test that None is returned for missing metrics."""
+        self.database.reports.find_one.return_value = None
+        self.assertIsNone(latest_metric(self.database, "report_uuid", METRIC_ID))
```
fniessink:
```suggestion
        self.assertIsNone(latest_metric(self.database, REPORT_ID, METRIC_ID))
```

fniessink (approved):

components/collector/src/base_collectors/collector.py
emdv88: Unless this update was for some reason only applied to the collector and not to the server.

Also, there are a lot of (unit)tests that will fail because there is no report_uuid is not present in the mocks. Requiring the report_uuid as a mandatory property would add complexity to all those tests without actually improving their value. So using get() improves the testability of the code.

All in all, using get() felt like a more fail-safe solution, even though I do agree that technically the report_uuid should always be filled.

components/collector/tests/base_collectors/test_collector.py
emdv88: I didn't add the report_uuid to the fixtures for thse tests since it is not necessary to test what these test tests. 

I do agree however that a testcase should be added to check that if a report_uuid is present, it will be forwarded.

components/server/src/internal/routes/measurement.py
emdv88: As in the collector, yes it should. But requiring less mandatory properties makes the code easier to test. I do feel that this is an interesting discussion though, might be worth having a talk about.

components/collector/src/base_collectors/collector.py
fniessink: Changing the production code to cater for the unit tests is not a good argument I think. And I think in this case the get() communicates the wrong message to the reader of the code: it suggests the report_uuid is optional while in fact it is not. The code should crash, violently and loudly, if there is no report_uuid, not swallow the issue and then fail somewhere else in the code later on, making it hard to debug.

And I don't agree it adds complexity to the tests. It adds duplication, yes, but that can easily be solved by refactoring the tests.

components/collector/tests/base_collectors/test_collector.py
fniessink: Well, you did add a report_uuid, but gave it the value of an empty string. My suggestion is to give it a proper value just like the metric_uuid argument: 
```suggestion
            json=dict(has_error=False, sources=[self._source()], metric_uuid="metric_uuid", report_uuid="report_uuid"),
```

By the way, to prevent SonarQube from complaining about duplicating "metric_uuid" and "report_uuid", there are constants defined in the fixtures module.

components/server/src/internal/routes/measurement.py
fniessink: See my comments above. Writing those comments convinced me even more that we shouldn't use a get() here ;-) 

I'm curious to hear what you think.

PR #3179, opened 2022-01-06
change popup text

### LINKED ISSUE #3171: Improve the status text in the tooltip
It would be convenient for the end-user if QT uses unambiguous, self-explanatory texts.
The text in the tooltips of each metric can be improved.

1. The text "Last measured" can be confused with a source measuring something. In case QT could not fetch anything, like in the attached image, this should be "last data retrieval attempt".
2. The text "First measured" could be interpreted as the moment when QT started measuring this, where it means to say: "last change of value".

image

emdv88:
Closes #3171.

### DIFF
--- a/components/frontend/src/metric/Metric.test.js
+++ b/components/frontend/src/metric/Metric.test.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Table } from 'semantic-ui-react';
-import { render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { Metric } from './Metric';
 import { DataModel } from '../context/DataModel';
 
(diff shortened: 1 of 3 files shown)


components/frontend/src/metric/Metric.test.js
```diff
@@ -84,3 +84,22 @@ it('renders the minutes as percentage', () => {
     expect(screen.getAllByText(/50% minutes/).length).toBe(1);
     expect(screen.getAllByText(/≦ 0% minutes/).length).toBe(1);
 });
+
+it('renders correct popups with status', async () => {
+    render_metric("violations");
+    fireEvent.mouseOver(screen.getByText(/50 violations/));
+
+    await waitFor(() => screen.getByTestId('value-popup'))
```
fniessink: I'm not a fan of `data-testid`. Can't we wait for `screen.getByText(/Metric was last measured/)`?

components/frontend/src/metric/Metric.test.js
```diff
@@ -84,3 +84,22 @@ it('renders the minutes as percentage', () => {
...
+    expect(screen.getAllByText(/Metric was last measured/).length).toBe(1);
+    expect(screen.getAllByText(/Value was first measured/).length).toBe(1);
+});
+
+it('renders correct popups without status', async () => {
+    report.subjects.subject_uuid.metrics.violations.status = null;
+    render_metric("violations");
+    fireEvent.mouseOver(screen.getByText(/50 violations/));
+
+    await waitFor(() => screen.getByTestId('value-popup'))
```
fniessink: I'm not a fan of data-testid. Can't we wait for screen.getByText(/Last measurement attempt/)?

fniessink (changes requested):
Please add an entry to the changelog.

fniessink (approved):

PR #3292, opened 2022-01-21
3269 shared component

### LINKED ISSUE #3269: Move shared code to new component

emdv88:
Closes #3269.


.circleci/config.yml
```diff
@@ -7,20 +7,24 @@ jobs:
...
+            3) component=shared_python;;
           esac
           cd components/$component
           python -m venv venv
           . venv/bin/activate
           pip install wheel
           pip install --progress-bar off -r requirements.txt -r requirements-dev.txt
+          if [ -f "requirements-internal.txt" ]; then
+            pip install --progress-bar off -r requirements-internal.txt
+          fi
```
fniessink, line 27:
```suggestion
          pip install --progress-bar off -r <(cat requirements*.txt)
```

components/server/Dockerfile
```diff
@@ -2,14 +2,18 @@ FROM python:3.10.2-buster AS compile-image
...
-USER server
 
 RUN python -m venv venv
 ENV PATH="/home/server/venv/bin:$PATH"
 RUN python -m pip install --upgrade pip
 RUN python -m pip install wheel
-COPY requirements.txt /requirements.txt
+COPY server/requirements.txt /requirements.txt
+COPY server/requirements-internal.txt /requirements-internal.txt
+COPY shared_python /home/shared_python/
```
fniessink, line 12: Is this step really necessary? Nothing is done with this /home/shared_python in the compile-image stage. And in the final stage, the code is copied again.

fniessink (approved):
Looks good to me. A few files were reformatted by black. That makes reviewing them hard. Next time it would be nice to do the formatting on a separate branch and then rebase the feature branch on top of it.

components/server/Dockerfile
emdv88: It's necessary to do the pip install the requirements-internal.txt below on line 14. Maybe it's a better solution to keep the requirements-internal out of the first stage completely.

.circleci/config.yml
emdv88: Turns out this doesn't work because pip will require hashes for all requirements files if it requires them for one. So the internal requirements have to be installed in a separate command.

components/server/Dockerfile
emdv88: Actually, on closer inspection, I think the best is to leave it like this for now. In the future we should look for a solution to not install the internal dependencies with the -e argument. When that is achieved, we don't need to copy the shared_python code in the final stage anymore.

PR #3430, opened 2022-02-16
3191 instantiate sources

### LINKED ISSUE #3191: Instantiate sources when instantiating a metric
Currently, when a report is instantiated (for example via a call to latest_reports()), it also instantiates it's subjects, which in turn instantiates it's metrics. However, the metric does not yet instantiate it's sources.

This issue is about instantiating all sources on metric instantiation. This is also a first step in a refactor to remove the Data classes (external/model/data.py) and use model classes instead.

fniessink:
Closes #3191.


fniessink:
Closing this one in favor of #3444.

PR #3467, opened 2022-02-22
Dark mode

### LINKED ISSUE #3159: Allow for user to change to dark mode layout
Open issues:
- [x] Listen for changes to preferred mode (https://stackoverflow.com/questions/56393880/how-do-i-detect-dark-mode-using-javascript)
- [x] Date picker dark mode => won't do for now.

fniessink:
Closes #3159.


components/frontend/src/AppUI.js
```diff
@@ -61,67 +63,73 @@ export function AppUI({
...
+                            open_report={open_report}
+                            reload={reload}
+                            report_date={report_date}
+                            report_uuid={report_uuid}
+                            reports={reports}
+                            reports_overview={reports_overview}
+                            sortColumn={sortColumn}
+                            sortDirection={sortDirection}
+                            toggleVisibleDetailsTab={toggleVisibleDetailsTab}
+                            visibleDetailsTabs={visibleDetailsTabs}
```
emdv88, line 127: I think we should really consider using contextproviders a little more to limit the amount of properties. In that sense a good start with the darkmode!

components/frontend/src/semantic_ui_react_wrappers/Card.js
```diff
@@ -0,0 +1,34 @@
...
+import { DarkMode } from '../context/DarkMode';
+
+export function Card(props) {
+    const darkMode = useContext(DarkMode)
+    let { style, ...otherProps} = props
+    if (darkMode) {
+        if (!style) {
+            style = {}
+        }
+        style.backgroundColor = "black"
```
emdv88: One could argue that it's not consistent to override styles in dark mode but not in light mode. Maybe override the backgroundcolor in light mode aswell? Or even fancier: come up with an algorithm that inverts the provided color no matter what!

components/frontend/src/semantic_ui_react_wrappers/Card.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function Header(props) {
+    const darkMode = useContext(DarkMode)
+    let { style, ...otherProps} = props
+    if (darkMode) {
+        if (!style) {
+            style = {}
+        }
+        style.color = "white"
```
emdv88: same as above

components/frontend/src/semantic_ui_react_wrappers/Feed.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function Date(props) {
+    const darkMode = useContext(DarkMode)
+    let { style, ...otherProps } = props;
+    if (darkMode) {
+        if (!style) {
+            style = {}
+        }
+        style.color = "grey"
```
emdv88: same as above

components/frontend/src/semantic_ui_react_wrappers/Feed.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function Summary(props) {
+    const darkMode = useContext(DarkMode)
+    let { style, ...otherProps } = props;
+    if (darkMode) {
+        if (!style) {
+            style = {}
+        }
+        style.color = "white"
```
emdv88: same as above

components/frontend/src/semantic_ui_react_wrappers/Table.js
```diff
@@ -0,0 +1,15 @@
...
+    return (
+        <SemanticUITable inverted={useContext(DarkMode)} {...props} />
+    )
+}
+
+Table.Body = SemanticUITable.Body
+Table.Cell = SemanticUITable.Cell
+Table.Header = SemanticUITable.Header
+Table.HeaderCell = SemanticUITable.HeaderCell
+Table.Row = SemanticUITable.Row
```
emdv88, line 15: Is there a way to do this automatically for all props of Table? This way it might be tricky if some developer in the future wants to use a property that we have not used yet.

components/frontend/src/AppUI.js
fniessink: Agreed.

components/frontend/src/semantic_ui_react_wrappers/Card.js
fniessink: Well, the Semantic UI default is really a light mode theme, so... But in this instance I think the solution should be to add an "inverted" class to the Card and change the color in a CSS file.

components/frontend/src/semantic_ui_react_wrappers/Table.js
fniessink: I don't know. It is possible to iterate over the properties of an object, but I don't want to run the risk of unexpectedly copying more than necessary.

components/frontend/src/semantic_ui_react_wrappers/Card.js
emdv88: agreed, that would be even better. Maybe an option with less work involved would be to change the default behavior to black, but keep any specified color?

components/frontend/src/semantic_ui_react_wrappers/Table.js
emdv88: we could loop over all properties, see if they start with a capital. According to convention then we should only get the actual react components :D

emdv88: I think the '...' syntax could be applied here too actually, but indeed it's not obvious what gets copied in any of those cases

PR #3648, opened 2022-03-31
3192 remove data classes

emdv88:


components/server/src/external/database/reports.py
```diff
@@ -140,3 +140,18 @@ def latest_reports_overview(database: Database, max_iso_timestamp: str = "") ->
 def report_exists(database: Database, report_uuid: ReportId):
     """Return whether a report with the specified report uuid exists."""
     return report_uuid in database.reports.distinct("report_uuid")
+
+
+def latest_report_for_uuids(all_reports: list[Report], *uuids: list) -> list[Report]:
```
fniessink: `*uuids: list` means that every uuid is a list I think? Also, there is a type for the union of ReportId, etc. so you can write:
```suggestion
def latest_report_for_uuids(all_reports: list[Report], *uuids: ItemId) -> list[Report]:
```

components/server/src/external/database/reports.py
```diff
@@ -140,3 +140,18 @@ def latest_reports_overview(database: Database, max_iso_timestamp: str = "") ->
...
+    """Return the report for given child entity uuids.
+    The uuids can be of a report, subject, metric or source.
+    The returned results will be in order of provided uuids.
+    """
+    reports = []
+    for uuid in uuids:
+        for report in all_reports:  # pragma: no cover-behave
+            child_uuids = set().union(report.subject_uuids, report.metric_uuids, report.source_uuids)
+            if uuid in child_uuids or report.uuid == uuid:
```
fniessink:
```suggestion
            uuids = {report.uuid}.union(report.subject_uuids, report.metric_uuids, report.source_uuids)
            if uuid in uuids:
```

components/server/src/external/model/actions.py
```diff
@@ -54,27 +59,38 @@ def copy_report(report, data_model):
...
-    nr_items = len(items)
-    item_to_move = getattr(data, item_type)
-    item_to_move_id = getattr(data, f"{item_type}_uuid")
-    old_index = list(items.keys()).index(item_to_move_id)
+    if isinstance(container, Report):
+        items_dict = container.subjects_dict
+    elif isinstance(container, Subject):
+        items_dict = container.metrics_dict
+    else:
+        items_dict = container.sources_dict
```
fniessink, line 71: If the `subjects_dict`, `metrics_dict`, and `sources_dict` would have the same name (`items_dict`?), this if-elif-else wouldn't be necessary. But I'm not sure `items_dict` is a good name, so maybe keep it like this for now.

components/server/src/external/model/actions.py
```diff
@@ -54,27 +59,38 @@ def copy_report(report, data_model):
...
-        reordered_items[item_to_move_id] = item_to_move
-    container[item_type + "s"] = reordered_items
+        reordered_items[item_to_move.uuid] = item_to_move
+
+    if isinstance(container, Report):
+        container["subjects"] = reordered_items
+    elif isinstance(container, Subject):
+        container["metrics"] = reordered_items
+    else:
+        container["sources"] = reordered_items
```
fniessink, line 94: If `Report`, `Subject`, and `Metric` would have a `set_items` method, this would be shorter. But, same doubts as above: `set_items` is maybe too abstract a name.

components/shared_python/src/shared/model/report.py
```diff
@@ -51,35 +80,82 @@ def _subjects(self, subject_data) -> dict[str, Subject]:
...
+                )[color] += 1
 
         return summary
+
+    def instance_and_parents_for_uuid(
+        self,
+        metric_uuid: MetricId = None,
+        source_uuid: SourceId = None,
+    ) -> tuple | None:
+        """Find an instance and it's parents.
```
fniessink:
```suggestion
        """Find an instance and its parents.
```

components/shared_python/src/shared/model/report.py
```diff
@@ -51,35 +80,82 @@ def _subjects(self, subject_data) -> dict[str, Subject]:
...
+        example:
+        If a metric_uuid is provided,
+        this function will return the metric, it's subject and it's report in that order:
```
fniessink:
```suggestion
        this function will return the metric, its subject and its report in that order:
```

fniessink (approved):

PR #3713, opened 2022-04-14
3192 remove subject and metric data classes

emdv88:


components/server/src/external/database/reports.py
```diff
@@ -102,7 +102,7 @@ def _prepare_documents_for_insertion(
             document[key] = value
 
 
-def latest_report(database: Database, data_model, report_uuid: str) -> Report | None:
+def latest_report(database: Database, data_model, report_uuid: str):
```
fniessink, line 105: Why did you remove the return type?

components/server/tests/external/routes/test_notification.py
```diff
@@ -22,6 +22,7 @@ def setUp(self):
         self.database = Mock()
         self.report = Report({}, create_report())
         self.database.reports.find.return_value = [self.report]
```
fniessink: Can this line be removed?

fniessink (approved):

components/server/src/external/database/reports.py
emdv88: Because otherwise mypy starts complaining about bad typing in all endpoints where this function is used without covering the case that the report is None (report_uuid doesnt exist, these are about 15 different endpoints for reports and notifications).

The solution is of course to implement a 404 error because the specified report has not been found. However, I think this should fall outside of the scope of this story. Instead I made story #3717 to solve this.

components/server/tests/external/routes/test_notification.py
emdv88: removed, good find!

components/server/src/external/database/reports.py
fniessink: Ah, clear.

fniessink (approved):
Don't forget to fix the SonarQube code smell.

emdv88:
Rebased and fixed the code smell. I'll merge as soon as the pipeline passed

PR #3769, opened 2022-04-29
2961 split server in two

emdv88:


emdv88:
Not all tests pass yet, but we're getting there. So I think a review might already be helpful. Some things that I still don't understand:

- building the docs fails because for some reason it doesn't find the shared_python directory
- deepsource requires an audit, but I don't know how to do that
- collecting coverage report over both internal and external server for the feature tests

Other than that, the application tests still fail so i guess something is still wrong. And some unittests need to be added to the shared_python module for complete coverage.

emdv88:
Ah and it seems that sounarcloud forgot all issues that were ignored or marked as false positive, so that will still need to be fixed aswell.

fniessink:
This seems relevant: https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/?utm_campaign=2022-05-03-brnd-dockerfilemultiplebuildcontexts&utm_medium=social&utm_source=twitter

fniessink:
Implemented via other PRs.

PR #3922, opened 2022-05-27
Internal reports endpoint and internal datamodel endpoint

emdv88:


fniessink (changes requested):
Code looks fine, but given that Sonar reports <100% coverage it seems you didn't move all the test code?

emdv88:
I don't understand where the sonar coverage comes from... in the unittests and in the feature tests we have 100% coverage everywhere. Any idea?

fniessink:
It seems coverage.py is not finding all the sources. But I've been reading https://coverage.readthedocs.io/en/6.4/source.html and I can't see what we're doing wrong.

emdv88:
But all other pipelines use coverage aswell and there it measures 100% coverage. I ran into the same problem in the original 2961 branch.

I guess this has something to do with the sonar configuration, or which coverage files it is looking at. But I couldn't find any sonar configuration in the repo.

PR #3983, opened 2022-06-07
Move initialization

fniessink:


components/shared_python/tests/shared/utils/test_functions.py
```diff
@@ -45,3 +45,13 @@ def test_future_report_date_time(self, request):
...
+
+
+class MD5HashTest(unittest.TestCase):
+    """Test md5 hash function."""
+
+    def test_hash(self):
+        """Test that hash is correct."""
+        test_string = "test"
+        test_hash = md5_hash(test_string)
+        self.assertEqual(test_hash, "098f6bcd4621d373cade4e832627b4f6")
```
fniessink: This test is the same as the one on lines 20-22. One of them can be removed.

emdv88: True, removed

PR #4003, opened 2022-06-09
remove old dependencies from external-server

emdv88:


emdv88:
incorporated in https://github.com/ICTU/quality-time/pull/4004

PR #4019, opened 2022-06-10
contact internal server in feature tests

emdv88:


tests/feature_tests/ci/test.sh
```diff
@@ -13,7 +13,15 @@ python3 -m venv venv
...
-RENDERER_HOST=localhost python tests/quality_time_server_under_coverage.py &> ../../build/quality_time_server.log &
+RENDERER_HOST=localhost python tests/quality_time_server_under_coverage.py &> ../../build/quality_time_external_server.log &
+deactivate
+cd ../..
+cd components/internal_server || exit
+python3 -m venv venv
+. venv/bin/activate
+ci/pip-install.sh
+coverage erase
+python tests/quality_time_server_under_coverage.py &> ../../build/quality_time_internal_server.log &
```
fniessink, line 25: Below (I can't give comments on lines outside the changed lines apparently) you also need to:
- combine the coverage of the internal server with the external server, I think?
- stop the internal server under coverage (I think you'll have to change the name of the two `quality_time_server_under_coverage.py`'s to `quality_time_internal_server_under_coverage.py` and `quality_time_external_server_under_coverage.py` to be able to kill them both).

fniessink (approved):

tests/feature_tests/ci/test.sh
emdv88: i tried combining the internal coverage aswell, but if i do that it reports the entire external server as uncovered. More over, it turns out that if i don't do that, I retain the 100% coverage. 
I don't fully understand why this is.

I think the kill command should kill both processes exactly because the files have the same name. Maybe not the nicest solution, but it works.

PR #4068, opened 2022-06-21
Allow for creating (Jira) issues from Quality-time.

### LINKED ISSUE #2931: Allow for creating a Jira ticket from Quality-time
# User story

As a quality manager I want to create an issue in the configured issue tracker and have it automatically linked from Quality-time so that I can create and link issues in one step.

## Acceptance criteria

The story is considered complete if:
- [x] The user can create an issue for a metric in the configured issue tracker from Quality-time, and
- [x] The issue is linked from the metric, and
- [x] The metric is linked from the issue

## Assumptions 

- An issue tracker is configured.
- The configured issue tracker user has the necessary rights to create an issue in the issue tracker.
- If the Quality-time user is a different user than is configured for the issue tracker, the Quality-time user can still edit the newly created issue in the issue tracker.

## Out of scope

- Quality-time doesn't decide on its own to create tickets.
- Automatically accept the metric as technical debt (if red or yellow)
- Automatically getting the available project keys and issue types from Jira.

# Tasks

### Frontend

- [x] Add a field for a default project for new issues to the issue tracker configuration
- [x] Add a field for a default issue type for new issues to the issue tracker configuration
- [x] Add a button next to the issue tracker id field "Create new issue for this metric"
- [x] Button should be inactive if there's no issue tracker configured.
  - [x] Button should be inactive if the user removes the issue tracker. 
- [x] Clicking the button invokes the external server endpoint for creating issues. 
- [x] The frontend uses the issue id returned by the external server to open a new browser tab for the user to change the issue with extra information.
- [x] If the configured issue tracker user has no rights to create issues in the issue tracker or something else goes wrong, show an appropriate error message using the toaster.

### External server

- [x] Add endpoint to create an issue for a specific metric. The endpoint uses the issue tracker (e.g. Jira) API to create a new issue (with the metric name as title), inserts the issue id into the issue id field, and returns the issue id to the frontend.

## Feature tests

- [x] User tries to create a new issue (really creating an issue would require a fake issue tracker)

## Documentation

- [x] Update changelog
  - [x] Mention that new tab may be blocked by the browser and user needs to allow Quality-time to open a new tab. 
- [x] Update screenshots

fniessink:
Closes #2931.


components/external_server/src/model/issue_tracker.py
```diff
@@ -58,3 +94,16 @@ def _basic_auth_credentials(self) -> tuple[str, str] | None:
     def _auth_headers(self) -> dict[str, str]:
         """Return the authorization headers, if any."""
         return dict(Authorization=f"Bearer {self.private_token}") if self.private_token else {}
+
+
+def instantiate_issue_tracker(report: Report) -> IssueTracker:
```
emdv88: shouldn't this be a classmethod?
```suggestion
    @classmethod
    def from_report(cls, report: Report) -> IssueTracker:
        ...
        return cls(url, project_key, issue_type, username, password, private_token)
```

components/external_server/src/routes/metric.py
```diff
@@ -138,3 +139,31 @@ def post_metric_attribute(metric_uuid: MetricId, metric_attribute: str, database
...
+    """Add a new issue to the metric using the configured issue tracker."""
+    data_model = latest_datamodel(database)
+    reports = latest_reports(database, data_model)
+    report = latest_report_for_uuids(reports, metric_uuid)[0]
+    issue_tracker = instantiate_issue_tracker(report)
+    metric, subject = report.instance_and_parents_for_uuid(metric_uuid=metric_uuid)
+    summary = f"Quality-time metric '{metric.name}'"
+    metric_url = dict(bottle.request.json)["metric_url"]
+    description = (
+        f"Please address metric '[{metric.name}|{metric_url}]' of subject '{subject.name}' "
```
emdv88: would it be useful to get target and current value (plus their units) in the description?

components/external_server/src/model/issue_tracker.py
emdv88: or just use the __init__ method if this is the only way we initialize a issue_tracker. According the the dataclass docs you can write:
`@dataclass(init=False)`
to prevent it from creating an init method

components/external_server/src/routes/metric.py
fniessink: Yes, the description needs improvement.

components/external_server/src/model/issue_tracker.py
fniessink: I was trying to prevent the issue tracker class from having knowledge about the internal structure of the report class. I think it makes more sense to ask a report for its issue tracker than passing a report to the issue tracker to have the issue tracker grab all the parameters from the report dict. On the other hand, I don't want to move the issue tracker class to the shared python code it it's only used in the external server. I'll subclass the report class in external server and add a get_issue_tracker() method there.

components/external_server/src/routes/metric.py
fniessink: On second thought, copying the measurement value to the issue only leads to outdated information since its not updated.

PR #4168, opened 2022-07-10
3935 automatically remove accepted technical debt when all tracked issues for a metric are resolved

### LINKED ISSUE #3935: Automatically remove accepted technical debt when all tracked issues for a metric are resolved
# User story

As a quality manager I want Quality-time to ignore accepted technical debt if the metric has linked issues that are all resolved so that the metric turns red if the issues did not fix the metric.

This works similar to the technical debt end date: as soon as the technical debt end date is in the past, the fact that technical debt is accepted is ignored and the metric is no longer grey if it doesn't meet its target. Likewise, as soon as all linked issues are resolved, the fact that technical debt is accepted will be ignored and the metric will no longer be grey if it doesn't meet its target

Note: This requires Quality-time to understand the status/resolution of issues. In case of Jira, use the builtin issue resolution field, not the status field that can be (and is) freely changed.

## Acceptance criteria

- [x] A metric with accepted technical debt that has issues that are all resolved ignores the technical debt status.

## Assumptions

- An issue tracker has been configured
 
## Open issues

- Is the text in the UI (see below) clear enough?

## Out of scope

- Changing the technical debt status (we don't do that when the technical debt end date passes either).
- Making this configurable. Since we moved the issue fields to a technical debt tab in #4165 it is clear issues and technical debt are related.

# Tasks

### Frontend

- [x] Show the issue status category (todo, doing, done) as label color.
- [x] If the metric has accepted technical debt, the target column includes a popup with the debt target. 
- [x] If the metric has accepted technical debt that is being ignored, the target column popup shows why it is ignored.

###  Shared python

- [x] Change `def accept_debt_expired(self) -> bool` in  `components/shared_python/src/shared/model/metric.py` to also return true if all metric issues have been resolved.
 
### Collector

- [x] Change `collector` to also collect issue status (open/closed) in `components/collector/src/source_collectors/jira/issue_status.py`.

## Feature tests

- [x] Create a metric that doesn't meet its target, has accepted technical debt and resolved issues, and check that its status is `target_not_met`.

## Documentation

- [x] Update changelog
  - [x] Mention technical debt status under [Editing metrics](https://quality-time.readthedocs.io/en/latest/usage.html#editing-metrics)

fniessink:
Closes #3935.

### DIFF
--- a/components/collector/src/model/issue_status.py
+++ b/components/collector/src/model/issue_status.py
@@ -1,6 +1,7 @@
 """Issue status model class."""
 
 from dataclasses import dataclass
+from typing import Literal
 
 from collector_utilities.type import ErrorMessage, URL
 
@@ -10,12 +11,19 @@ class Issue:
     """Class representing issues."""
 
     issue_id: str
+    name: str | None = None
+    summary: str | None = None
     created: str | None = None
     updated: str | None = None
 
     def as_dict(self) -> dict:
         """Return the issue as dict."""
-        return dict(issue_id=self.issue_id, created=self.created, updated=self.updated)
+        return dict(
+            issue_id=self.issue_id, name=self.name, summary=self.summary, created=self.created, updated=self.updated
+        )
+
+
+IssueStatusCategory = Literal["todo", "doing", "done"]
 
 
 class IssueStatus:  # pylint: disable=too-few-public-methods
@@ -26,15 +34,15 @@ class IssueStatus:  # pylint: disable=too-few-public-methods
         issue_id: str,
         *,
         name: str = None,
+        status_category: IssueStatusCategory = None,
         created: str = None,
         updated: str = None,
         summary: str = None,
         connection_error: ErrorMessage = None,
         parse_error: ErrorMessage = None
     ) -> None:
-        self.issue = Issue(issue_id, created, updated)
-        self.name = name
-        self.summary = summary
+        self.issue = Issue(issue_id, name, summary, created, updated)
+        self.status_category = status_category
         self.parse_error = parse_error
         self.connection_error = connection_error
         self.api_url: URL | None = None
@@ -43,12 +51,11 @@ class IssueStatus:  # pylint: disable=too-few-public-methods
     def as_dict(self) -> dict:
         """Return the issue status as dict."""
         status = dict(
-            name=self.name,
-            summary=self.summary,
+            status_category=self.status_category,
             parse_error=self.parse_error,
             connection_error=self.connection_error,
             api_url=self.api_url,
             landing_url=self.landing_url,
             **self.issue.as_dict(),
         )
-        return {key: value for key, value in status.items() if value}
+        return {key: value for key, value in status.items() if value is not None}
--- a/components/collector/src/source_collectors/jira/issue_status.py
+++ b/components/collector/src/source_collectors/jira/issue_status.py
@@ -1,13 +1,17 @@
 """Jira issue status collector."""
 
+from typing import cast
+
 from collector_utilities.type import URL
-from model import IssueStatus, SourceResponses
+from model import IssueStatus, IssueStatusCategory, SourceResponses
 
 from .base import JiraBase
 
 
 class JiraIssueStatus(JiraBase):
     """Jira issue status collector."""
+
+    STATUS_CATEGORY_MAPPING = dict(done="done", indeterminate="doing", new="todo")
 
     async def _api_url(self) -> URL:
         """Override to get the issue, including the status field, from Jira."""
@@ -23,7 +27,16 @@ class JiraIssueStatus(JiraBase):
         """Override to get the issue status from the responses."""
         json = await responses[0].json()
         name = json["fields"]["status"]["name"]
+        jira_status_category = json["fields"]["status"]["statusCategory"]["key"]
+        status_category = cast(IssueStatusCategory, self.STATUS_CATEGORY_MAPPING.get(jira_status_category, "todo"))
         created = json["fields"]["created"]
         updated = json["fields"].get("updated")
         summary = json["fields"].get("summary")
-        return IssueStatus(self._issue_id, name=name, created=created, updated=updated, summary=summary)
+        return IssueStatus(
+            self._issue_id,
+            name=name,
+            status_category=status_category,
+            created=created,
+            updated=updated,
+            summary=summary,
+        )
--- a/components/collector/src/source_collectors/jira/issue_status.py
+++ b/components/collector/src/source_collectors/jira/issue_status.py
@@ -11,7 +11,13 @@ from .base import JiraBase
 class JiraIssueStatus(JiraBase):
     """Jira issue status collector."""
 
+    # Map the Jira status categories, the keys, to the Quality-time status categories, the values:
     STATUS_CATEGORY_MAPPING = dict(done="done", indeterminate="doing", new="todo")
+
+    # Note, Jira distinguishes statuses and status categories. Statuses can be added, but the list of status
+    # categories is fixed, per https://jira.atlassian.com/browse/JRASERVER-36241. As Quality-time only needs to know
+    # whether a issue is done or not, we use status categories to keep track of that and to color the issues in the
+    # UI. We do add the status name to the issue and display it in the UI, but the status is not interpreted.
 
     async def _api_url(self) -> URL:
         """Override to get the issue, including the status field, from Jira."""
--- a/components/collector/tests/source_collectors/jira/test_issue_status.py
+++ b/components/collector/tests/source_collectors/jira/test_issue_status.py
@@ -1,5 +1,7 @@
 """Unit tests for the Jira issue status collector."""
 
+from model.issue_status import IssueStatusCategory
+
 from .base import JiraTestCase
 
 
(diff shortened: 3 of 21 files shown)


components/collector/src/model/issue_status.py
```diff
@@ -43,12 +51,11 @@ def __init__(
...
-            summary=self.summary,
+            status_category=self.status_category,
             parse_error=self.parse_error,
             connection_error=self.connection_error,
             api_url=self.api_url,
             landing_url=self.landing_url,
             **self.issue.as_dict(),
         )
-        return {key: value for key, value in status.items() if value}
+        return {key: value for key, value in status.items() if value is not None}
```
wkoot, line 61: Is it a problem that the nested Issue object is not displayed? Name and summary are now missing

components/collector/src/source_collectors/jira/issue_status.py
```diff
@@ -1,14 +1,18 @@
...
-from model import IssueStatus, SourceResponses
+from model import IssueStatus, IssueStatusCategory, SourceResponses
 
 from .base import JiraBase
 
 
 class JiraIssueStatus(JiraBase):
     """Jira issue status collector."""
 
+    STATUS_CATEGORY_MAPPING = dict(done="done", indeterminate="doing", new="todo")
```
wkoot, line 15: magic values, could need a comment

components/collector/src/source_collectors/jira/issue_status.py
```diff
@@ -23,7 +27,16 @@ async def _parse_issue_status(self, responses: SourceResponses) -> IssueStatus:
         """Override to get the issue status from the responses."""
         json = await responses[0].json()
         name = json["fields"]["status"]["name"]
+        jira_status_category = json["fields"]["status"]["statusCategory"]["key"]
```
wkoot, line 36: can the existence of fields always be assumed, or perhaps safer to use `.get('',{})`?

components/collector/tests/source_collectors/jira/test_issue_status.py
```diff
@@ -42,14 +53,38 @@ def assert_issue_status(self, response, summary: str = None, connection_error: s
 
     async def test_issue_status(self):
         """Test that the issue status is returned."""
-        issue_status_json = dict(fields=dict(status=dict(name=self.ISSUE_NAME), created=self.CREATED))
+        issue_status_json = dict(
+            fields=dict(status=dict(name=self.ISSUE_NAME, statusCategory=dict(key="new")), created=self.CREATED)
```
wkoot, line 57: Status should no longer contain name?

wkoot: (applies to multiple cases below as well)

components/frontend/src/measurement/IssueStatus.js
```diff
@@ -37,7 +37,7 @@ function IssueWithTracker({ issueStatus, showIssueCreationDate, showIssueSummary
     let popupHeader = "";
     if (issueStatus.connection_error) { popupHeader = "Connection error"; popupContent = "Quality-time could not retrieve data from the issue tracker." }
     if (issueStatus.parse_error) { popupHeader = "Parse error"; popupContent = "Quality-time could not parse the data received from the issue tracker." }
-    const color = popupContent ? "red" : "blue";
+    const color = popupContent ? "red" : {todo: "grey", doing: "blue", done: "green"}[issueStatus.status_category ?? "todo"];
```
wkoot, line 40: Is this the only change? I would expect the other changes in py code also reflected here, i.e. moving `name` and `summmary` to Issue

components/shared_python/src/shared/model/measurement.py
```diff
@@ -221,10 +227,22 @@ def equals(self, other: Measurement) -> bool:
     def debt_target_expired(self) -> bool:
         """Return whether the technical debt target is expired.
 
-        Technical debt can expire because it was turned off or because the end date passed.
+        Technical debt is considered expired when it was turned off, when the end date passed or when the metric has
+        issues that all have been done.
```
wkoot, line 231: perhaps more clear to write "when the metric has issues that have all been resolved"

components/shared_python/src/shared/model/measurement.py
```diff
@@ -221,10 +227,22 @@ def equals(self, other: Measurement) -> bool:
...
         """
         any_debt_target = any(self[scale].get("debt_target") is not None for scale in self.metric.scales())
-        return self.metric.accept_debt_expired() if any_debt_target else False
+        accept_debt_expired = not self.metric.accept_debt() or date.today().isoformat() > self.metric.debt_end_date()
```
wkoot: why is this line duplicate with 103 above?

components/shared_python/tests/shared/model/test_measurement.py
```diff
@@ -158,10 +155,10 @@ def test_calculate_status_near_target(self):
             previous_scale_measurement=None,
             measurement=measurement,
             target=1,
-            near_target=2,
-            debt_target=3,
+            debt_target=2,
```
wkoot, line 158: do these values make sense? Why is it target < debt < near ?

components/shared_python/tests/shared/model/test_measurement.py
```diff
@@ -255,6 +252,31 @@ def test_status_informative(self):
...
+        """Test that the debt target is considered to be expired when all issues have been done."""
+        measurement = self.measurement(
+            self.metric(accept_debt=True, debt_target="100", issue_ids=["FOO-40"]),
+            count=dict(debt_target="100"),
+            issue_status=[dict(status_category="done", issue_id="FOO-40")],
+        )
+        self.assertTrue(measurement.debt_target_expired())
+
+    def test_debt_target_not_expired(self):
+        """Test that the debt target is not considered to be expired without issues."""
```
wkoot, line 265: Is this the same as testing when there are issues without status, for which there was an assumption that they are not done?

tests/feature_tests/steps/measurement.py
```diff
@@ -53,6 +53,7 @@ def measure(context, number, total="100"):
 @when("the collector measures issue '{issue_id}' status '{name}'")
 def measure_issue_status(context, issue_id, name):
     """Post the measurement with the issue status."""
+    status_category = "done" if name == "Completed" else "todo"
```
wkoot, line 56: case 'doing' does not get tested?

docs/src/usage.md
```diff
@@ -238,7 +238,22 @@ If a metric doesn't meet the target value, but your team isn't able to fix the s
...
+
+```{note}
+Metrics with accepted technical debt are displayed with a money icon and grey background as long as their measurement value is worse than the metric's target and equal to or better than the metric's technical debt target.
+
+However, measurement values of a metric with accepted technical debt will *not* be evaluated against the technical debt target when:
+
+- the metric has a technical debt end date that is in the past, or
+- the metric has issues associated with it, and the issue tracker reports that all these issues have been done.
+
+If one or both of these situations apply, the technical debt target is ignored and the measurement value is evaluated against the target values. Depending on the evaluation, the metric is shown as green, yellow, or red, as usual.
```
wkoot: perhaps more clear "if any of these situations apply"

components/collector/src/source_collectors/jira/issue_status.py
fniessink: The status category should always there. But, if it's not, the user will get an error message in the UI. That's because the collector does all of its work within a broad exception handler that makes sure it doesn't crash. If an exception happens, the error is added to the `IssueStatus` instance and reported to the user.

components/collector/tests/source_collectors/jira/test_issue_status.py
fniessink: No, IssueStatuses have both a status name and a status category. The status category is one of three status categories that never change (Todo, In Progress, Done). The status (name) can be anything a Jira administrator adds to Jira, e.g. the status category In Progress can have statuses Waiting, Doing, Blocked, etc. We use status category to decide whether an issues has been done, but we also store the status name in the IssueStatus so it can be displayed in the UI.

components/frontend/src/measurement/IssueStatus.js
fniessink: Name and summary were moved, but the `IssueStatus.as_dict` method makes sure the JSON that goes over the line is not changed:

```python
class IssueStatus:
    ...
    def as_dict(self) -> dict:
        """Return the issue status as dict."""
        status = dict(
            status_category=self.status_category,
            parse_error=self.parse_error,
            connection_error=self.connection_error,
            api_url=self.api_url,
            landing_url=self.landing_url,
            **self.issue.as_dict(),  # <= includes the issue attributes in the dict (that will become the JSON sent to the frontend)
        )
        return {key: value for key, value in status.items() if value is not None}
```

components/shared_python/src/shared/model/measurement.py
fniessink: Yeah, I was trying to stay close to the Jira terminology for the status category "done". But resolved may be more recognizable for users.

components/shared_python/src/shared/model/measurement.py
fniessink: Good question 👍🏻 . I have made `date.today().isoformat() > self.metric.debt_end_date()` a method on the Metric class to reduce the duplication: 

```python
    def debt_end_date_passed(self) -> bool:
        """Return whether the end date of the accepted technical debt has passed."""
        return date.today().isoformat() > self.debt_end_date()
```

components/shared_python/tests/shared/model/test_measurement.py
fniessink: Yeah, I don't like the fact that we have unit tests on this class in the first place. The Measurement class uses  auxiliary classes to implement comparisons for different scales. These sort of "inner" classes should really not be tested directly, but only via the Measurement class I think. But that's a bit too big a change to include here.

components/shared_python/tests/shared/model/test_measurement.py
fniessink: No, here we test that when all issues are done (line 260) that the debt target should be ignored.

tests/feature_tests/steps/measurement.py
fniessink: Correct. There's no functional difference between todo and doing. Only the color in the UI is different.

components/collector/src/model/issue_status.py
fniessink: The nested issue object is included via the as_dict method.

PR #4201, opened 2022-07-17
Create a shared data model component.

### LINKED ISSUE #4199: Create separate shared python component with only the data model
This will allow the notifier and collector component to have the data model installed as python package instead of having to download it from the internal server.

# Tasks

- [x] Create shared data model component by copying shared python 
  - [x] Update Circle CI config
  - [x] Update GitHub Actions config
  - [x] Remove all non-data model code from shared data model
- [x] Update shared python 
  - [x] Use shared data model in shared python code
  - [x] Remove data model from original shared python
  - [x]  Move utility functions now only used by the external server to the external server
- [x] Update external server
  - [x] Use shared data model in external server python code
  - [x] Include shared data model in external server docker image
  - [x] Move utility functions now only used by the external server to the external server
- [x] Update internal server
  - [x] Use shared data model in internal server python code
  - [x] Include shared data model in internal server docker image
  - [x] Remove internal server data model endpoint
- [x] Update collector
  - [x] Import shared data model in collector instead of downloading it
- [x] Update notifier
  - [x] Import shared data model in notifier instead of downloading it
- [x] Update docs
  - [x] Update generation of reference docs to use shared data model instead of shared python 
  - [x] Update software documentation
  - [x] Update developer documentation

# Out of scope/future work

- Using the Pedantic version of the data model in the collector and notifier instead of loading the JSON-version
- Using the imported version of the data model in the internal server instead of retrieving it from the database

fniessink:
Closes #4199.


components/shared_python/tests/shared/initialization/test_secrets.py
```diff
@@ -12,7 +12,6 @@ class TestSecrets(unittest.TestCase):
     def setUp(self):
         """Set up database mocks."""
         self.database = Mock()
-        self.database.secrets.insert = Mock()
```
wkoot, line 15: Unused?

components/external_server/tests/routes/test_datamodel.py
```diff
@@ -13,6 +17,23 @@ def setUp(self):
...
+        """Test that the data model is None if it's not there."""
+        self.database.datamodels.find_one.return_value = None
+        self.assertEqual({}, get_data_model(self.database))
+
+    @patch("bottle.request")
+    def test_get_data_model_unchanged(self, mocked_request):
+        """Test that a 304 is returned when the data model is unchanged."""
+        mocked_request.headers = {"If-None-Match": "W/" + md5_hash("now")}
+        self.database.datamodels.find_one.return_value = dict(_id=123, timestamp="now")
+        self.assertRaises(bottle.HTTPError, get_data_model, self.database)
```
wkoot: The docstring suggests you want to explicitly check for a 304

wkoot (approved):

components/shared_python/tests/shared/initialization/test_secrets.py
fniessink: Yes, the secrets are inserted with another method (insert_one), but since the return value of the method is not used it's not necessary to define `self.database.secrets.insert_one`.

components/external_server/tests/routes/test_datamodel.py
fniessink: Good point. Fixed by catching the exception and checking the status code.

PR #4276, opened 2022-07-22
3645 settings save endpoint

emdv88:


components/external_server/src/database/users.py
```diff
@@ -0,0 +1,20 @@
...
+    """Update the existing user or insert a new one."""
+    database.users.replace_one(
+        dict(username=user.username),
+        dict(user.to_dict()),
+        upsert=True,
+    )
+
+
+def get_user(database: Database, username: str) -> User | None:
+    """Update the existing user or insert a new one."""
```
fniessink: Docstring doesn't match function name and implementation.

components/external_server/src/routes/settings.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@bottle.get("/api/v3/settings", authentication_required=True)
+def get_settings(database: Database) -> dict:
+    """Retrieve settings for user."""
+    user = find_user(database)
+    return dict(settings=user.settings)
+
+
+@bottle.post("/api/v3/settings/update", authentication_required=True)
```
fniessink: Why not `@bottle.post("/api/v3/settings", ...)`, or maybe even `@bottle.put("/api/v3/settings", ...)`?

components/frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "quality-time-app",
   "version": "4.2.0-rc.5",
   "private": true,
-  "proxy": "http://localhost:5001",
+  "proxy": "http://127.0.0.1:5001",
```
fniessink: Why is this needed? Seems unrelated to the current story?

components/shared_server_code/src/shared/initialization/database.py
```diff
@@ -45,6 +45,7 @@ def create_indexes(database: Database) -> None:
     """Create any indexes."""
     database.datamodels.create_index("timestamp")
     database.reports.create_index("timestamp")
+    database.users.create_index("username", unique=True)
```
fniessink, line 48: Do we really need an index? Are we ever going to have so many users that the index is faster than a full collection scan?

components/shared_server_code/src/shared/utils/type.py
```diff
@@ -29,7 +30,12 @@ class User:
     email: str = ""
     common_name: str = ""
     verified: bool = False
+    settings: defaultdict = field(default_factory=lambda: defaultdict(dict))
```
fniessink, line 34: Maybe add a comment why a default dict is used?

fniessink (changes requested):

components/frontend/package.json
emdv88: This was needed for me to locally run the frontend for some reason. To be honest I thought I had it put back to localhost already, but will do so now

components/shared_server_code/src/shared/initialization/database.py
emdv88: good point

emdv88: On closer inspection, this is a way to ensure that the username is unique. Since we find users by username and especially now with multiple LDAP servers there is as far as I know no other location where uniqueness of usernames is guaranteed, we might run into problems otherwise.

emdv88: Of course to do it nicely, we should implement a proper workaround (or error message) for the case we get multiple users with identical usernames, but given the propability of this happening, I feel that just crashing is a good temporary solution.
In any case better than in a later stage potentially providing user data of a different.

PR #4286, opened 2022-07-24
Use Pydantic-version of the data model instead of the JSON-version in the notifier

### LINKED ISSUE #4200: Replace JSON-version of data model with Pydantic-version
Depends on #4199.

- [x] Use Pydantic version of data model in notifier
- [x] Use Pydantic version of data model in collector
- [x] Use Pydantic version of data model in internal server
- [x] Use Pydantic version of data model in external server (but only when used internally in the server, the frontend needs the version from the database to support time travel)
- [x] Use Pydantic version of data model in docs

fniessink:
Partially implements #4200.


components/notifier/src/quality_time_notifier.py
```diff
@@ -42,8 +39,9 @@ async def notify(log_level: int = None) -> NoReturn:
         await asyncio.sleep(sleep_duration)
 
 
-def record_health(filename: str = "/home/notifier/health_check.txt") -> None:
+def record_health() -> None:
     """Record the current date and time in a file to allow for health checks."""
+    filename = "/home/notifier/health_check.txt"
```
fniessink, line 44: Changed this to prevent a DeepSource warning about a potential security vulnerability.

components/shared_data_model/src/shared_data_model/meta/base.py
```diff
@@ -37,6 +37,14 @@ def __getitem__(self, key: str) -> ValueT:
         """Return the model with the specified key."""
         return self.__root__[key]
 
-    def items(self) -> MappingView:
+    def get(self, key: str) -> ValueT | None:
+        """Return the model with the specified key or None if the key does not exist."""
+        return self.__root__.get(key)
```
wkoot: Why not use `__getitem__` ?

docs/src/create_reference_md.py
```diff
@@ -46,58 +47,57 @@ def markdown_header(header: str, level: int = 1, index: bool = False) -> str:
     return index_preamble + "#" * level + f" {header}\n\n"
 
 
-def subject_sections(data_model, level) -> str:
+def subject_sections(level: int) -> str:
     """Return the subjects as Markdown sections."""
     markdown = ""
-    for subject in sorted(data_model["subjects"].values(), key=lambda item: str(item["name"])):
-        markdown += subject_section(data_model, subject, level)
+    for subject in sorted(DATA_MODEL.subjects.values(), key=get_model_name):
```
wkoot, line 53: Nice improvement

docs/src/create_reference_md.py
```diff
@@ -168,76 +168,78 @@ def metric_source_slug(metric_name: str, source_name: str) -> str:
...
-    metric_name = metric["name"]
-    metric_link = f"[{metric_name.lower()}]({slugify(metric_name)})"
-    source_link = f"[{source_name}]({slugify(source_name)})"
+    source = DATA_MODEL.sources[source_key]
+    metric = DATA_MODEL.metrics[metric_key]
+    metric_link = f"[{metric.name.lower()}]({slugify(metric.name)})"
+    source_link = f"[{source.name}]({slugify(source.name)})"
     markdown = markdown_paragraph(f"{source_link} can be used to measure {metric_link.lower()}.")
-    if documentation := source.get("documentation", {}).get(metric_key):
+    if documentation := source.documentation and source.documentation.get(metric_key):
```
wkoot: Does the walrus operator work on a single segment? I would probably put parenthesis around the first part to be sure

docs/src/create_reference_md.py
```diff
@@ -168,76 +168,78 @@ def metric_source_slug(metric_name: str, source_name: str) -> str:
...
     return markdown
 
 
-def parameter_description(parameter: dict) -> str:
+def parameter_description(parameter: Parameter) -> str:
     """Return the Markdown version of the parameter."""
-    short_name = parameter["short_name"]
-    if help_text := parameter.get("help", ""):
+    short_name = parameter.short_name
+    if help_text := parameter.help:
```
wkoot: This if/else block feels a bit dirty. Why not:
```
help_text = " " + parameter.help if parameter.help else ""
```

components/shared_data_model/src/shared_data_model/meta/base.py
fniessink: `__getitem__` is implemented as well (line 36). Not sure if that's what you're asking?

docs/src/create_reference_md.py
fniessink: I'll put in the parentheses, unless we get some linter complaining about unnecessary parentheses :-)

components/external_server/src/model/actions.py
```diff
@@ -63,12 +65,14 @@ def move_item(
     container: Report | Subject | Metric, item_to_move: Subject | Metric | Source, new_position: Position
 ) -> tuple[int, int]:
     """Change the item position."""
+    ItemsDictType = MutableMapping[ItemId, Metric | Source | Subject]
```
wkoot: This method feels like it's trying to be very generic, while it's really just a simple mixin. Is there not a different way to do it?

components/external_server/src/database/sessions.py
```diff
@@ -27,6 +27,6 @@ def delete(database: Database, session_id: SessionId) -> None:
     database.sessions.delete_one(dict(session_id=session_id))
 
 
-def get(database, session_id) -> SessionId:
+def get(database, session_id):
```
wkoot: ?

components/external_server/src/database/reports.py
```diff
@@ -56,7 +56,8 @@ def _get_change_key(change: Change) -> str:
 
 def metrics_of_subject(database: Database, subject_uuid: SubjectId, max_iso_timestamp: str = "") -> list[MetricId]:
     """Return all metric uuid's for one subject, without the entities, except for the most recent one."""
-    report_filter = {f"subjects.{subject_uuid}": DOES_EXIST}
+    report_filter: dict[str, bool | dict[str, bool] | dict[str, str]] = {}
```
wkoot, line 60: This is just to be taken at face value when there's no context, what's going on? :D

components/external_server/src/model/actions.py
fniessink: Yeah, I agree, but I didn't want to refactor, just get the typing correct. This is already becoming a huge change set and at least typing changes don't affect run time behavior.

components/external_server/src/database/sessions.py
fniessink: The function isn't returning a SessionId but a session from the database, so the `-> SessionId` is definitely wrong. @emdv88 created a Session model class in src/model/session.py, but this function is not returning a instance of that Session class so I'm not really sure how to fix this at the moment.

components/external_server/src/database/reports.py
fniessink: MongoDB filters are a beast :-)  What can I add to help the reader here? A link to MongoDB docs?

wkoot: some inline comments and link to docs will probably suffice

PR #4342, opened 2022-08-04
3645 default settings object

### LINKED ISSUE #3645: Store the user settings in the database
As a quality-time user I want the UI settings to be stored so they are kept between sessions.

But also keep them in the URL as query parameters because that allows people to:
- Share settings
- Create a favorite link with the correct settings
- Use settings without being logged in
- Pass the settings to the renderer

Tasks:
- Add a user to the Users collection in MongoDB on login of a user, if the user does not exist in the Users collection. The username is the unique key.
- Add a button to the settings panel: "Save current settings as default". Clicking this button results in the current settings (from the settings panel) to be saved in the Users collection.
- Change the "Reset all settings" button to use the saved defaults if present. If not, use the 'factory defaults'.
- Add a settings endpoint to the (external) server with GET and POST access. A GET call returns a JSON blob with the settings. The POST call takes the same JSON blob and stores it.
- These settings are stored in the Users collection:
  - All settings from the settings panel
- The frontend retrieves the settings on login of the user but does not ignore any URL query parameters.

Out of scope:
- Removing users and their settings (we assume the number of users using Quality-time is pretty limited)

emdv88:
Closes #3645.


emdv88:
This merge request should be all that is left to complete #3645.

fniessink (changes requested):
Autosave changed settings instead of having a save button.

PR #4447, opened 2022-08-24
When creating issues, include the rationale for the metric in the cre…

### LINKED ISSUE #4232: Include metric rationale in issues
# User story

As a quality manager I want Quality-time to include the metric rationale in issues that it creates so that the product owner can give the right priority to the issue.

## Acceptance criteria

- [x] New issues created from Quality-time contain the rationale for the metric in the issue description. 

## Assumptions

- An Issue tracker is configured.

## Open issues

- How to distinguish the rational from the rest of the description? A header? Make it a comment?

# Tasks

### External server

- [x] Expand the issue creation endpoint to include the rationale from the data model in the issue text

## Documentation

- [x] Update changelog

fniessink:
…ated issue so that people encountering the issue in the issue tracker have a better understanding of why the issue needs addressing. Closes #4232.

### DIFF
--- a/components/external_server/src/routes/metric.py
+++ b/components/external_server/src/routes/metric.py
@@ -5,6 +5,7 @@ from typing import Any, cast
 import bottle
 from pymongo.database import Database
 
+from shared_data_model import DATA_MODEL
 from shared.database.datamodels import latest_datamodel
 from shared.database.measurements import insert_new_measurement, latest_measurement
 from shared.database.reports import insert_new_report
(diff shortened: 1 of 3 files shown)


components/external_server/src/routes/metric.py
```diff
@@ -172,5 +173,6 @@ def create_issue_description(metric, subject, report) -> str:
     metric_url = dict(bottle.request.json)["metric_url"]
     return (
         f"Metric '[{metric.name}|{metric_url}]' of subject '{subject.name}' "
-        f"in Quality-time report '{report.name}' needs attention."
+        f"in Quality-time report '{report.name}' needs attention.\n\n"
+        f"Why address '{metric.name}'? {DATA_MODEL.metrics[metric.type()].rationale}"
```
wkoot, line 177: Are all current rationales suitable for including? Perhaps something to put in a hover/tooltip?

fniessink: This is text that's inserted into Jira issues. Not sure you can make a tooltip in a Jira issue?

fniessink: I checked, but I don't see an option to create a tooltip in the description of a Jira issue. All metrics have a rationale, so at least there will be some text. If the text is not suitable, the user can always delete it when editing the issue description in Jira. And if the rationale is never useful to have in the issue, I'm sure they'll let us know :-)

Issue #4444, opened 2022-08-24
When changing the metric type, don't delete sources that do not support the new type


fniessink:
**Is your feature request related to a problem? Please describe.**
Changing the type of a metric results in sources that do not support the new type to be deleted. If a user changes the type by accident and changes the type back, this may result in loss of configured sources.

**Describe the solution you'd like**
Don't remove sources that don't support the metric type, but keep them. Flag in the UI that a source is ignored (with a grey instead of red background label) because it doesn't support the metric. Ignore the source in the collector (that should happen automatically as there will be no collector the metric/source combination).

**Describe alternatives you've considered**
- Warning the user with a dialog. However, it's easy to ignore dialogs and Quality-time doesn't use dialogs in other place either.
- Making it impossible to change the type, but that also blocks the user when they know what they're doing.

**Additional context**
Related bug: #4443. It's probably easier to implement this if the bug is fixed first.

PR #4448, opened 2022-08-25
 When changing the type of a metric with configured sources, don't re…

### LINKED ISSUE #4443: Changing the type of a metric deletes the source, even if the source supports both the old and new type
**Describe the bug**
Changing the type of a metric deletes the source, even if the source supports both the old and new type.

**Expected behavior**
The sources that support both the old and the new metric type should not be deleted.

 - Quality-time version: v4.3.0

fniessink:
…move the sources that support both the old and the new metric type. Fixes #4443.

### DIFF
--- a/components/external_server/tests/routes/test_metric.py
+++ b/components/external_server/tests/routes/test_metric.py
@@ -25,6 +25,7 @@ from ..fixtures import (
     REPORT_ID,
     REPORT_ID2,
     SOURCE_ID,
+    SOURCE_ID2,
     SUBJECT_ID,
     SUBJECT_ID2,
     create_report,
@@ -42,7 +43,7 @@ class PostMetricAttributeTest(unittest.TestCase):
         self.data_model = dict(
             _id="id",
             metrics=dict(
-                old_type=dict(name="Old type", scales=["count"]),
+                old_type=dict(name="Old type", scales=["count"], sources=["source_type1", "source_type2"]),
                 new_type=dict(
                     scales=["count", "version_number"],
                     default_scale="count",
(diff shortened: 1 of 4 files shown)


docs/src/usage.md
```diff
@@ -205,7 +205,11 @@ After you've added a metric, the metric is visible in the subject's metric table
 :class: only-dark
 ```
 
-The first parameter is the "Metric type". The metric type determines what gets measured. When you change the metric type, the sources you can select in the "Sources" tab change accordingly. By default, the name of the metric is equal to its type, "Accessibility violations" in the example above, but you can change the metric name using the "Metric name" field.
+The first parameter is the "Metric type". The metric type determines what gets measured. By default, the name of the metric is equal to its type, "Accessibility violations" in the example above, but you can change the metric name using the "Metric name" field. When you change the metric type, the sources you can select in the "Sources" tab change accordingly.
+
+```{warning}
+If you change the type of a metric that has sources configured, sources that do not support the new metric type will be removed.
```
wkoot, line 211: The UI should display a warning if this is the case

components/external_server/tests/routes/test_metric.py
```diff
@@ -114,11 +115,15 @@ def test_post_metric_name(self, request):
         )
 
     def test_post_metric_type(self, request):
-        """Test that the metric type can be changed."""
+        """Test that the metric type can be changed and that sources that support the new type are not removed."""
```
wkoot, line 119: Should there not be a test that verifies changing `old_type` of `source_type2` will yield `new_type` without sources?

docs/src/usage.md
fniessink: We have a separate story to address that: #4444.

components/external_server/tests/routes/test_metric.py
fniessink: The metric has two sources, one of which supports the new metric type, the other doesn't. So when the metric type is changed from old type to new type, the source that supports the new type is retained and the other is removed. I think that covers all scenario's.

PR #4451, opened 2022-08-26
4444 make it impossible to change the metric type if that would mean sources have to be deleted

### LINKED ISSUE #4444: When changing the metric type, don't delete sources that do not support the new type

fniessink:
Closes #4444.

### DIFF
--- a/components/external_server/src/routes/metric.py
+++ b/components/external_server/src/routes/metric.py
@@ -132,12 +132,8 @@ def post_metric_attribute(metric_uuid: MetricId, metric_attribute: str, database
         return dict(ok=True)  # Nothing to do
     metric[metric_attribute] = new_value
     if metric_attribute == "type":
-        sources_to_keep = {
-            key: source
-            for key, source in metric["sources"].items()
-            if source["type"] in data_model["metrics"][new_value]["sources"]
-        }
-        metric.update(default_metric_attributes(database, new_value), sources=sources_to_keep)
+        # Update the metric attributes, but keep the sources
+        metric.update(default_metric_attributes(database, new_value), sources=metric["sources"])
     description = (
         f"{{user}} changed the {metric_attribute} of metric '{old_metric_name}' of subject "
         f"'{subject.name}' in report '{report.name}' from '{old_value}' to '{new_value}'."
--- a/components/frontend/src/measurement/MeasurementSources.test.js
+++ b/components/frontend/src/measurement/MeasurementSources.test.js
@@ -11,7 +11,7 @@ it('renders one measurement source', () => {
             <MeasurementSources metric={
                 {
                     type: "metric_type",
-                    sources: { source_uuid: { type: "sourcr_type", name: "Source name" } },
+                    sources: { source_uuid: { type: "source_type", name: "Source name" } },
                     latest_measurement: { sources: [{ source_uuid: "source_uuid" }] }
                 }
             }
--- a/components/frontend/src/measurement/MeasurementSources.test.js
+++ b/components/frontend/src/measurement/MeasurementSources.test.js
@@ -3,12 +3,15 @@ import { render, screen } from '@testing-library/react';
 import { DataModel } from '../context/DataModel';
 import { MeasurementSources } from './MeasurementSources';
 
+const dataModel = { metrics: { metric_type: { sources: ["source_type"] } } }
+
 it('renders one measurement source', () => {
     render(
-        <DataModel.Provider value={{}}>
+        <DataModel.Provider value={dataModel}>
             <MeasurementSources metric={
                 {
-                    sources: { source_uuid: { name: "Source name" } },
+                    type: "metric_type",
+                    sources: { source_uuid: { type: "sourcr_type", name: "Source name" } },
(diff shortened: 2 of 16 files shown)


components/external_server/src/routes/metric.py
```diff
@@ -132,12 +132,8 @@ def post_metric_attribute(metric_uuid: MetricId, metric_attribute: str, database
...
     metric[metric_attribute] = new_value
     if metric_attribute == "type":
-        sources_to_keep = {
-            key: source
-            for key, source in metric["sources"].items()
-            if source["type"] in data_model["metrics"][new_value]["sources"]
-        }
-        metric.update(default_metric_attributes(database, new_value), sources=sources_to_keep)
+        # Update the metric attributes, but keep the sources
+        metric.update(default_metric_attributes(database, new_value), sources=metric["sources"])
```
wkoot, line 136: This is simply returning to the original right?

components/frontend/src/measurement/MeasurementSources.test.js
```diff
@@ -3,12 +3,15 @@ import { render, screen } from '@testing-library/react';
...
+
 it('renders one measurement source', () => {
     render(
-        <DataModel.Provider value={{}}>
+        <DataModel.Provider value={dataModel}>
             <MeasurementSources metric={
                 {
-                    sources: { source_uuid: { name: "Source name" } },
+                    type: "metric_type",
+                    sources: { source_uuid: { type: "sourcr_type", name: "Source name" } },
```
wkoot: looks like a typo `sourcr_type`

components/frontend/src/source/Source.js
```diff
@@ -88,21 +89,47 @@ function Parameters({ metric, source, source_uuid, connection_error, parse_error
...
     const parseError = measurement_source?.parse_error || "";
-    const configurationTabLabel = connectionError || parseError ? <Label color='red'>{"Configuration"}</Label> : "Configuration";
+    const referenceManualURL = `https://quality-time.readthedocs.io/en/v${process.env.REACT_APP_VERSION}/reference.html`;
+    const configErrorMessage = <>
+        <p>
+            {sourceName} cannot be used to measure {metricName}. This configuration error occurs if the type of a
+            metric is changed to a metric type that is not supported by the configured source type(s).
+        </p>
+        <p>
+            There are several ways to remediate this error:
```
wkoot: Perhaps use "remedy" as it is also correct, but simpler

components/shared_server_code/src/shared/model/measurement.py
```diff
@@ -261,8 +262,12 @@ def update_measurement(self) -> None:
             self[scale].update_value_and_status()
 
     def sources_ok(self) -> bool:
-        """Return whether there are sources and none have parse or connection errors."""
+        """Return whether there are sources and none have configuration, parse, or connection errors."""
         sources = self.sources()
+        for source in sources:
```
wkoot, line 267: Does it make sense to short-circuit parse and connection errors before looping over the sources?

docs/src/changelog.md
```diff
@@ -8,6 +8,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
 
+## [Unreleased]
+
+### Deployment notes
+
+If your currently installed *Quality-time* version is not v4.4.0, please read the v4.4.0 deployment notes.
+
+### Fixed
+
+- When changing the type of a metric with configured sources, retain the sources whether they support the new metric type or not. Flag the sources that do not support the new metric type as a 'configuration error' and explain how the user can remediate the situation. Fixes [#4443](https://github.com/ICTU/quality-time/issues/4443) and closes [#4444](https://github.com/ICTU/quality-time/issues/4444).
```
wkoot: If earlier text was changed to remedy, then this one should be as well

components/external_server/src/routes/metric.py
fniessink: It sets default attribute values for the new type, yes. Except that the sources are not removed.

components/frontend/src/source/Source.js
fniessink: Or "fix", even easier :-)

components/shared_server_code/src/shared/model/measurement.py
fniessink: I don't think it is worth it, metrics typically have only one or a handful of sources.

PR #4456, opened 2022-08-30
When measuring violations with SonarQube as source, add the selected …

### LINKED ISSUE #4449: Make QT use the same severities for a link as selected in the source
I'm always frustrated when I click the link to see violations in Sonar and I end up at a page showing all violations.

I would like QT to use a link with the violation severities selected that are used in the source config.

image

image

fniessink:
…issue types (bug, vulnerability, and/or code smell) and severities (info, minor, major, critical, and/or blocker) to the landing URL so that the user sees only the relevant violations when navigating to SonarQube. Closes #4449.

### DIFF
--- a/components/collector/src/source_collectors/sonarqube/security_warnings.py
+++ b/components/collector/src/source_collectors/sonarqube/security_warnings.py
@@ -26,7 +26,7 @@ class SonarQubeSecurityWarnings(SonarQubeViolations):
             landing_path = "project/issues"
             extra_url_parameters = f"{self._query_parameter('severities')}&resolved=false&types=VULNERABILITY"
         else:
-            landing_path = "security_hotspots"
+            landing_path = "project/security_hotspots"
         return URL(f"{base_landing_url}/{landing_path}{common_url_parameters}{extra_url_parameters}")
 
     async def _get_source_responses(self, *urls: URL, **kwargs) -> SourceResponses:
--- a/components/collector/src/source_collectors/sonarqube/security_warnings.py
+++ b/components/collector/src/source_collectors/sonarqube/security_warnings.py
@@ -19,17 +19,15 @@ class SonarQubeSecurityWarnings(SonarQubeViolations):
         base_landing_url = await SourceCollector._landing_url(self, responses)  # pylint: disable=protected-access
         component = self._parameter("component")
         branch = self._parameter("branch")
+        common_url_parameters, extra_url_parameters = f"?id={component}&branch={branch}", ""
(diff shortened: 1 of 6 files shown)


components/collector/src/source_collectors/sonarqube/security_warnings.py
```diff
@@ -19,19 +19,17 @@ async def _landing_url(self, responses: SourceResponses) -> URL:
...
         component = self._parameter("component")
         branch = self._parameter("branch")
+        common_url_parameters, extra_url_parameters = f"?id={component}&branch={branch}", ""
         if "vulnerability" in security_types and "security_hotspot" in security_types:
-            landing_url = f"{base_landing_url}/dashboard?id={component}&branch={branch}"
+            landing_path = "dashboard"
         elif "vulnerability" in security_types:
-            landing_url = f"{base_landing_url}/project/issues?id={component}&resolved=false&branch={branch}"
+            landing_path = "project/issues"
+            extra_url_parameters = f"{self._query_parameter('severities')}&resolved=false&types=VULNERABILITY"
```
wkoot, line 28: should hardcoded `types=VULNERABILITY` not be combined with `_query_parameter(self.types_parameter)`?

wkoot: or is that automagically done by order of combining in `sonarqube/violations.py`.. hmm probably good to add a comment

components/collector/tests/source_collectors/sonarqube/test_violations.py
```diff
@@ -54,4 +54,24 @@ async def test_violations(self):
...
+    async def test_violations(self):
+        """Test that the number of violations is returned."""
+        response = await self.collect(get_request_json_return_value=self.json)
+        self.assert_measurement(
+            response, value="2", entities=self.expected_entities, landing_url=self.issues_landing_url
+        )
+
+    async def test_major_violations(self):
+        """Test that the number of major violations is returned."""
+        self.set_source_parameter("severities", ["major"])
```
wkoot, line 67: perhaps good to set two severities, to also verify that the comma sep urlencoded `%2C` works

wkoot (changes requested):

components/collector/src/source_collectors/sonarqube/security_warnings.py
fniessink: Using `_query_parameter(self.types_parameter)` would result in `&security_types=...` but we need `&types=...`. I'll add a comment.

PR #4505, opened 2022-09-05
Refactor shared server code to not use fake data model

fniessink:


components/external_server/tests/database/test_reports.py
```diff
@@ -58,7 +57,6 @@ def setUp(self) -> None:
                 {"report_uuid": REPORT_ID2, "subjects": {SUBJECT_ID2: {"metrics": {METRIC_ID3: {}, METRIC_ID4: {}}}}},
             ),
         ]
-        return super().setUp()
```
wkoot, line 61: What happened here?

components/external_server/tests/routes/test_source.py
```diff
@@ -273,23 +270,24 @@ def test_url_with_token(self, mock_get, request):
     @patch.object(requests, "get")
     def test_url_with_token_and_validation_path(self, mock_get, request):
         """Test that the source url can be changed and that the availability is checked."""
-        self.data_model["sources"]["owasp_zap"]["parameters"]["private_token"]["validation_path"] = "/rest/api"
```
wkoot, line 276: Is this already a default value in the data model?

components/shared_server_code/tests/shared/database/test_measurements.py
```diff
@@ -1,30 +1,25 @@
...
 
 from tests.fixtures import METRIC_ID, METRIC_ID2
 
+from ..base import DataModelTestCase
 
-class MeasurementsTest(unittest.TestCase):
-    """Unit test for getting and inserting measurements."""
 
-    @classmethod
-    def setUpClass(cls) -> None:
```
wkoot, line 19: Is it ok to change from setUpClass to setUp? We weren't using any of the implicit setUpClass aspects?

components/external_server/tests/database/test_reports.py
fniessink: Introduced by accident (by the IDE?) in a previous refactor I think, and now removed.

components/external_server/tests/routes/test_source.py
fniessink: Yes, Jira (that we now use as source in this unit test) has that parameter by default so no need to change the data model anymore.

components/shared_server_code/tests/shared/database/test_measurements.py
fniessink: Yes, this works. First `DataModelTestCase.setUpClass()` is invoked once to create the `DATA_MODEL` class attribute and then `MeasurementsTest.setUp()` is called before each unit test to set up the database fixture.

PR #4524, opened 2022-09-06
Allow for setting default labels on Jira issues created from Quality-…

### LINKED ISSUE #4468: Add default label(s) to Jira issues created from Quality-time
I want to quickly see in Jira which issues have been created from Quality-time
 
Issues created from Quality-time in Jira should have one or more default labels.

**Tasks**
- Add a field to the issue tracker configuration tab for "Labels".
- Add the default label(s) to the newly created issues in the 'create issue' endpoint.

fniessink:
…time. The labels can be configured in the issue tracker tab under the report header. Closes #4468.


components/external_server/src/model/issue_tracker.py
```diff
@@ -21,6 +21,32 @@ def as_dict(self) -> dict[str, str]:
...
+class IssueTrackerCredentials:
+    """Issue tracker credentials needed to create issues."""
+
+    username: str = ""
+    password: str = ""
+    private_token: str = ""
+
+    def basic_auth_credentials(self) -> tuple[str, str] | None:
+        """Return the basic authentication credentials, if any."""
+        return (self.username, self.password) if self.username or self.password else None
```
wkoot, line 34: Is it ok to return `None, pw` in case username is unset but password is not?

components/external_server/src/model/issue_tracker.py
```diff
@@ -21,6 +21,32 @@ def as_dict(self) -> dict[str, str]:
...
+
+    def auth_headers(self) -> dict[str, str]:
+        """Return the authorization headers, if any."""
+        return dict(Authorization=f"Bearer {self.private_token}") if self.private_token else {}
```
wkoot, line 38: If both basic auth credentials and token are set, shouldn't token take precedence? Are there cases in which we want to send both?

components/external_server/src/model/report.py
```diff
@@ -13,9 +13,10 @@ def issue_tracker(self) -> IssueTracker:
...
         parameters = issue_tracker_data.get("parameters", {})
         url = parameters.get("url", "")
-        project_key = parameters.get("project_key", "")
-        issue_type = parameters.get("issue_type", "")
-        username = parameters.get("username", "")
-        password = parameters.get("password", "")
-        private_token = parameters.get("private_token", "")
-        return IssueTracker(url, project_key, issue_type, username, password, private_token)
+        issue_parameters = IssueParameters(
+            parameters.get("project_key", ""), parameters.get("issue_type", ""), parameters.get("issue_labels", [])
```
wkoot, line 17: This feels a little messy. Perhaps use named args and put each entry on one line? hehe

components/external_server/src/model/issue_tracker.py
fniessink: No, Jira (the only issue tracker we currently support) needs either a private token or username *and* password.

components/external_server/src/model/report.py
fniessink: Ha, one of the reasons to introduce the IssueParameters and IssueTrackerCredential dataclasses was to work around the too-many-instance-variables Pylint warning! Using named args here would reintroduce that warning.

components/external_server/src/model/issue_tracker.py
fniessink: Currently we always send both and let Jira deal with it. So far that seems to work fine, either because users know they should provide either username/password or token, or because Jira knows what to do. Anyway, we really should improve the UI first to make it impossible for users to provide both types of credentials.

PR #4528, opened 2022-09-07
Move the contents of the "Notes on specific metrics" and "Notes on sp…

### LINKED ISSUE #4446: Include "notes on specific metrics" in the data model so they are exported to the reference manual
The "notes on specific metrics" can be removed from the user manual.

fniessink:
…ecific sources" from the user manual to the reference manual. Closes #4446.


docs/src/create_reference_md.py
```diff
@@ -142,6 +145,8 @@ def source_section(source: Source, source_key: str, level: int) -> str:
     """Return the source as Markdown section."""
     markdown = markdown_header(source.name, level, index=True)
     markdown += markdown_paragraph(source.description)
+    if documentation := (source.documentation and source.documentation.get("generic")):
```
wkoot: I think `source.documentation.get("generic") if source.documentation` would be more clear

docs/src/create_reference_md.py
```diff
@@ -142,6 +145,8 @@ def source_section(source: Source, source_key: str, level: int) -> str:
...
+        markdown += markdown_paragraph(documentation)
```
wkoot, line 151: How/where is non-generic documentation displayed in the docs?

components/shared_data_model/src/shared_data_model/sources/jenkins.py
```diff
@@ -24,6 +24,15 @@ def jenkins_access_parameters(*args, **kwargs):
     return access_parameters(*args, **kwargs)
 
 
+# Put the Jenkins token documentation in a temporary variable that doesn't trigger a security warning so we can
+# suppress the false positive Bandit warning below.
```
wkoot, line 28: hahaha, yikes

fniessink: Yes, I feel dirty.

docs/src/create_reference_md.py
fniessink: That's not valid Python? I've changed it like this so the walrus assignment doesn't depend on the fact that `a and b` returns `b` if `a` and `b` are both true-ish.

```suggestion
   if source.documentation and (documentation := source.documentation.get("generic")):
```

docs/src/create_reference_md.py
fniessink: Added a comment.

PR #4553, opened 2022-09-12
Add target=_blank to URLs in comments. Closes #4521.

### LINKED ISSUE #4521: add target="_blank" to auto generated anchors in comments
After entering/pasting a URL in the comment field, and leaving the field, QT automatically generates an anchor tag. This is useful however, when the user clicks it, the user moves away from QT which is undesired functionality.

When auto generating an anchor tag, add target="_blank" as parameter.

**Additional context**
image

fniessink:


components/external_server/tests/utils/test_functions.py
```diff
@@ -32,6 +35,29 @@ def test_md5_hash(self):
...
+
+    def test_sanitize_html_does_not_change_target(self):
+        """Test that existing target attributes are not changed."""
+        self.assertEqual(
+            f'<a href="{self.URL}" target="_top">{self.URL}</a>',
+            sanitize_html(f'<a href="{self.URL}" target="_top">{self.URL}</a>'),
+        )
+        self.assertEqual(
+            f'<a href="{self.URL}" target="_blank">{self.URL}</a>',
+            sanitize_html(f'<a href="{self.URL}" target="_blank">{self.URL}</a>'),
```
wkoot, line 54: Don't you want to test `target="_self"` here instead?

wkoot: Or `target="_top"` as documented

components/external_server/src/utils/functions.py
```diff
@@ -19,6 +19,7 @@
 # and Dlint complains 'insecure use of XML modules, prefer "defusedxml"'
 # but we give autolink_html clean html, so ignore the warning:
 from lxml.html.clean import autolink_html, clean_html  # noqa: DUO107, # nosec # pylint: disable=no-name-in-module
+import lxml  # noqa: DUO107, # nosec
```
wkoot: Is it possible to only import `fromstring` & `tostring` instead of the entire library ?

wkoot: Perhaps we should also look at using `BeautifulSoup` instead of `lxml`, although I guess we can't drop `lxml` entirely due to other requirements :(

components/external_server/tests/utils/test_functions.py
fniessink: It doesn't matter what the value of target is here, the point is that existing target attributes are not changed.

components/external_server/src/utils/functions.py
fniessink: Or bleach for cleaning HTML.

PR #4557, opened 2022-09-14
Add migration code to merge unmerged measurements

### LINKED ISSUE #4556: Clean up unmerged measurements
Clean up the unmerged measurements in the database caused by #4554.

fniessink:
Fixes #4556.

### DIFF
--- a/components/internal_server/src/quality_time_server.py
+++ b/components/internal_server/src/quality_time_server.py
@@ -13,13 +13,14 @@ import bottle  # skipcq: FLK-E402
 
 from shared.initialization.database import init_database  # skipcq: FLK-E402
 
-from initialization import init_bottle  # skipcq: FLK-E402
+from initialization import init_bottle, merge_unmerged_measurements  # skipcq: FLK-E402
 
 
 def serve() -> None:  # pragma: no cover-behave
     """Connect to the database and start the application server."""
     logging.getLogger().setLevel(logging.INFO)
     database = init_database()
+    merge_unmerged_measurements(database)
     init_bottle(database)
     server_port = os.environ.get("INTERNAL_SERVER_PORT", "5002")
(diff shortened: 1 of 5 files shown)


components/internal_server/src/quality_time_server.py
```diff
@@ -13,13 +13,14 @@
...
 
-from initialization import init_bottle  # skipcq: FLK-E402
+from initialization import init_bottle, merge_unmerged_measurements  # skipcq: FLK-E402
 
 
 def serve() -> None:  # pragma: no cover-behave
     """Connect to the database and start the application server."""
     logging.getLogger().setLevel(logging.INFO)
     database = init_database()
+    merge_unmerged_measurements(database)
```
wkoot, line 23: Do we have a followup story / reminder to remove this in the next release?

wkoot (approved):

components/internal_server/src/quality_time_server.py
fniessink: That would mean forcing people to install the upcoming release and not skip it. If the code is not too slow it doesn't harm to leave it in until the next major release.

wkoot (approved):

components/internal_server/src/quality_time_server.py
eric-nieuwland, line 23: Why not register in the database that this code has run?
This call could then be made conditional to absence of that registration.

fniessink: Good question. The advantage is that the code does not run multiple times. The disadvantage is that we need to set a flag which pollutes the database and/or would need to be removed at a later time. 

The migration only runs on startup and when there is nothing to do, should be pretty quick and not write anything to the database. If it takes more than a few seconds on our biggest database, we'll consider adding the flag.

PR #4587, opened 2022-09-20
Replace data picker widget with a supported version

### LINKED ISSUE #4467: Replace date picker with a date picker widget that supports React 18
Possible candidates:
- https://www.npmjs.com/package/react-semantic-ui-datepickers

Note: preferably a data picker that supports Semantic UI React.

Tasks:
- Update frontend code.
- Remove suppression of semantic-ui-calendar-react-17 in the deadpendency config.

fniessink:
Closes #4467.

### DIFF
--- a/components/frontend/src/App.js
+++ b/components/frontend/src/App.js
@@ -98,22 +98,15 @@ class App extends Component {
     }
 
     handleDateChange(_event, { name, value }) {
-        const today = new Date();
-        const today_string = String(today.getDate()).padStart(2, '0') + '-' + String(today.getMonth() + 1).padStart(2, '0') + '-' + today.getFullYear();
-        const new_report_date_string = value === today_string ? '' : value;
-        let reportDate = null;
         let parsed = registeredURLSearchParams(this.history);
-        if (new_report_date_string === "") {
-            parsed.delete("report_date")
+        if (!!value) {
+            parsed.set("report_date", value.toISOString().split("T")[0]);
         } else {
-            const isoDateString = new_report_date_string.split("-").reverse().join("-");
-            parsed.set("report_date", isoDateString);
-            reportDate = new Date(isoDateString)
-            reportDate.setHours(23, 59, 59);
+            parsed.delete("report_date")
         }
         const search = parsed.toString().replace(/%2C/g, ",")  // No need to encode commas
         this.history.replace({ search: search.length > 0 ? "?" + search : "" })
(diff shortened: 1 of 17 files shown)


components/frontend/src/App.js
```diff
@@ -33,15 +33,19 @@ class App extends Component {
         const pathname = this.history.location.pathname;
         const report_uuid = pathname.slice(1, pathname.length);
         const report_date_iso_string = registeredURLSearchParams(this.history).get("report_date") || "";
-        const report_date_string = isValidDate_YYYYMMDD(report_date_iso_string) ? report_date_iso_string.split("-").reverse().join("-") : "";
+        let reportDate = null;
+        if (isValidDate_YYYYMMDD(report_date_iso_string)) {
+            reportDate = new Date(report_date_iso_string);
+            reportDate.setHours(23, 59, 59);
```
wkoot, line 39: Why 1 second before midnight instead of just 0,0,0 ?

components/frontend/src/semantic_ui_react_wrappers/Form.test.js
```diff
@@ -0,0 +1,30 @@
...
+it('shows the form dropdown in darkmode', async () => {
+    await act(async () => {
+        const { container } = render(
+            <DarkMode.Provider value={true}>
+                <Form>
+                    <Form.Dropdown options={[{ key: "Hi", value: "Hi", text: "Hi" }]} value={"Hi"} />
+                </Form>
+            </DarkMode.Provider>
+        )
+        expect(container.querySelector(".inverted")).not.toBe(null)
```
wkoot, line 15: Why not just check true? Or is the logic on any value

components/frontend/src/App.js
wkoot: Oh, nvm, this was old code that's been moved. Is it tested elsewhere? Because in this PR I don't see any test on 23:59:59

wkoot (approved):

components/frontend/src/App.js
fniessink: It's one second before midnight because on the server the measurement end timestamp is compared with the report date and we want measurements that end on the report date to be included. Calling `setHours(0, 0, 0)` would effectively set the report date to the day before.

components/frontend/src/semantic_ui_react_wrappers/Form.test.js
fniessink: If I recall correctly it's not a boolean flag and setting it to false gives errors.

PR #4624, opened 2022-09-29
Show the number of times the desired reaction time deadline was missed

### LINKED ISSUE #4538: Show the amount of time the desired reaction time was not met
# User story

As a quality manager I want to show my manager and client whether metrics are reacted to in time so it's either clear the team is responding timely or that improvement actions are needed.

Quality-time shows per metric how many days, in the displayed period, metric desired reaction times were overrun. 

Suppose the desired reaction time for a red metric is three days. Then:
- If a metric is never red in the displayed period, the reaction time overrun is zero days
- If a metric is red for two days in the displayed period, the reaction time overrun is still zero days
- If a metric is red for four days in the displayed period, the reaction time overrun is one day
- If a metric is red for four days and then later for five days, the reaction time overrun is three days

## Acceptance criteria

- [ ] When the number of dates visible is > 1 then Quality-time shows the amount of time that metrics were not reacted to in time
- [ ] The time period is determined by the number of dates and the time between dates visible.

## Out of scope

- Averaging multiple metrics.
- Overrunning a desired reaction time for a red metric may be considered worse than overrunning a desired reaction time for a yellow metric. However, since the desired reaction time for a yellow metric can be made longer than the one for red metrics (and is longer by default), we assume overrunning the yellow reaction time is as bad as overrunning the red reaction time.
- Showing how often the deadline was missed. For the time being we assume that fixing issues in a timely manner is more important than how often a deadline is missed.
- Notifier notifying about (missed) deadlines.

## Open issues

- Can the overrun be calculated in the frontend or do we need to change the backend as well?

# Tasks

### Frontend

- [ ] Add column "Reaction time overrun" / "Amount of days the desired reaction times were overrun in the period <X>-<Y>"
- [ ] Show the column in the settings panel.

### External server

- [ ] Change GET report endpoint: include number of days overrun in the summary (if overrun cannot be calculated easily in the frontend)

## Feature tests

- [ ] Check overrun (if backend is changed to return number of days overrun)

## Documentation

- [ ] Update changelog
- [ ] Update screenshots
- [ ] Update software documentation

Reported by @Sebastiaan127001

fniessink:
Closes #4538.

### DIFF
--- a/components/frontend/src/measurement/Overrun.test.js
+++ b/components/frontend/src/measurement/Overrun.test.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Overrun } from './Overrun';
+
+it("renders nothing if there is no overrun", () => {
+    render(<Overrun measurements={[]} />)
+    expect(screen.queryAllByDisplayValue(/days/).length).toBe(0)
+})
+
+it("renders the days overrun if the metric has overrun its deadline", () => {
+    render(
+        <Overrun
+            metric_uuid="uuid"
+            measurements={[{metric_uuid: "uuid", start: "2020-01-01", end: "2020-12-31"}]}
+        />
+    )
+    expect(screen.queryAllByText(/days/).length).toBe(1)
+})
+
+it("merges the days overrun if the metric has consecutive measurements", () => {
+    render(
+        <Overrun
+            metric_uuid="uuid"
+            measurements={[{metric_uuid: "uuid", start: "2020-01-01", end: "2020-12-31"}]}
+        />
+    )
+    expect(screen.queryAllByText(/days/).length).toBe(1)
+})
--- a/components/frontend/src/measurement/Overrun.test.js
+++ b/components/frontend/src/measurement/Overrun.test.js
@@ -14,10 +14,10 @@ it("renders the days overrun if the metric has overrun its deadline", () => {
         <Overrun
             dates={dates}
             metric_uuid="uuid"
-            measurements={[{metric_uuid: "uuid", start: "2020-01-01", end: "2020-12-31"}]}
+            measurements={[{metric_uuid: "uuid", start: "2020-01-01", end: "2020-01-31"}]}
         />
     )
-    expect(screen.queryAllByText(/days/).length).toBe(1)
+    expect(screen.queryAllByText(/27 days/).length).toBe(1)
 })
 
 it("merges the days overrun if the metric has consecutive measurements", () => {
@@ -25,8 +25,13 @@ it("merges the days overrun if the metric has consecutive measurements", () => {
         <Overrun
             dates={dates}
             metric_uuid="uuid"
-            measurements={[{metric_uuid: "uuid", start: "2020-01-01", end: "2020-12-31"}]}
+            measurements={
+                [
+                    {metric_uuid: "uuid", start: "2020-01-01", end: "2020-01-10"},
+                    {metric_uuid: "uuid", start: "2020-01-10", end: "2020-01-20"}
+                ]
+            }
         />
     )
-    expect(screen.queryAllByText(/days/).length).toBe(1)
+    expect(screen.queryAllByText(/16 days/).length).toBe(1)
 })
--- a/components/frontend/src/measurement/Overrun.test.js
+++ b/components/frontend/src/measurement/Overrun.test.js
@@ -2,14 +2,17 @@ import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { Overrun } from './Overrun';
 
+const dates = [new Date("2020-01-01"), new Date("2020-12-31")]
+
 it("renders nothing if there is no overrun", () => {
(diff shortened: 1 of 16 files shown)


components/frontend/src/subject/SubjectTableHeader.js
```diff
@@ -140,6 +140,21 @@ const timeLeftHelp = <>
     </p>
 </>
 
+const overrunHelp = <>
+    <p>
+        The amount of days the desired reaction time to address the metric was exceeded in the displayed period.
```
wkoot:
```suggestion
        The amount of days the desired reaction time to address the metric, was exceeded in the displayed period.
```

components/frontend/src/subject/SubjectTableHeader.js
```diff
@@ -140,6 +140,21 @@ const timeLeftHelp = <>
...
+    </p>
+    <p>
+        Expand the report title (click <Icon fitted name="triangle right" />) to change the desired reaction times.
+    </p>
+    <p>
+        Hover the number of days to see an overview of when and for which statuses the metric had overruns in the displayed period.
```
wkoot:
```suggestion
        Hover over the number of days to see an overview of when and for which statuses the metric had overruns, in the displayed period.
```

components/frontend/src/subject/SubjectTableHeader.js
```diff
@@ -140,6 +140,21 @@ const timeLeftHelp = <>
...
+    </p>
+    <p>
+        Click the column header to sort the metrics by overrun.
```
wkoot:
```suggestion
        Click the column header to sort the metrics by overrun time.
```

wkoot: or perhaps "by time overrun"? hmm

components/frontend/src/utils.test.js
```diff
@@ -259,3 +259,61 @@ it("returns false when the user prefers light mode", () => {
...
+    expect(getMetricResponseOverrun("uuid", {}, {}, [{metric_uuid: "uuid", start: "2000-01-01", end: "2000-01-31"}])).toStrictEqual(
+        {
+            "overruns": [
+                {
+                    "status": "unknown",
+                    "start": "2000-01-01",
+                    "end": "2000-01-31",
+                    "actual_response_time": 30,
+                    "desired_response_time": 3,
+                    "overrun": 27
```
wkoot, line 283: hmm, the unit of overrun is the same as actual and desired times.. and the unit of all three is actually days! a little weird, but I guess adding `_days` to each would be even more messy :-)

components/frontend/src/measurement/Overrun.test.js
```diff
@@ -0,0 +1,32 @@
...
+        <Overrun
+            dates={dates}
+            metric_uuid="uuid"
+            measurements={[{metric_uuid: "uuid", start: "2020-01-01", end: "2020-12-31"}]}
+        />
+    )
+    expect(screen.queryAllByText(/days/).length).toBe(1)
+})
+
+it("merges the days overrun if the metric has consecutive measurements", () => {
```
wkoot, line 23: should there not be a test if both dates are overrun, i.e. 1+2 = 3?

components/frontend/src/utils.js
```diff
@@ -32,26 +32,74 @@ export function getMetricUnit(metric, dataModel) {
...
+
+export function getMetricResponseOverrun(metric_uuid, metric, report, measurements) {
+    const consolidatedMeasurements = getMetricResponseOverruns(metric_uuid, metric, report, measurements)
+    const scale = metric?.scale ?? "count"
+    let totalOverrun = 0;  // Amount of time the desired response time was not achieved for this metric
+    const overruns = []
+    consolidatedMeasurements.forEach((measurement) => {
+        const status = measurement?.[scale]?.status || "unknown"
+        if (status in metricReactionDeadline) {
+            const desiredResponseTime = getMetricDesiredResponseTime(report, status) * 24 * 60 * 60 * 1000
```
wkoot: Does it make sense to also extract `* 24 * 60 * 60 * 1000` to a separate function? :P

components/frontend/src/subject/SubjectTableHeader.js
fniessink:
```suggestion
        Click the column header to sort the metrics by the number of days overrun.
```

components/frontend/src/utils.test.js
fniessink: Yeah, I'm not a fan of Hungarian notation. Maybe we should do more with types in JS, but I haven't looked into that yet.

components/frontend/src/measurement/Overrun.test.js
fniessink: There's actually test data missing here. Good catch.

components/frontend/src/utils.js
fniessink: Constant you mean? Yeah, will do.

PR #4707, opened 2022-10-10
Make subject measurements endpoint more flexible

fniessink:
Make the subject measurements endpoint in the external server more flexible: instead of always returning measurements for the previous 28 weeks (max 7 dates with max 4 week intervals), make the endpoint return the measurements between two dates.

This not only saves time/bandwidth, but will also make it easier to implement #4538 because we receive exactly the right measurements to determine for how long metrics have missed their deadlines between the two dates.

### DIFF
--- a/components/external_server/src/routes/subject.py
+++ b/components/external_server/src/routes/subject.py
@@ -1,6 +1,5 @@
 """Subject routes."""
 
-from datetime import datetime, timedelta, timezone
 from typing import cast
 
 import bottle
@@ -119,18 +118,14 @@ def post_subject_attribute(subject_uuid: SubjectId, subject_attribute: str, data
 
 @bottle.get("/api/v3/subject/<subject_uuid>/measurements", authentication_required=False)
 def get_subject_measurements(subject_uuid: SubjectId, database: Database):
-    """Return all measurements for the subject within the last 28 weeks."""
+    """Return all measurements for the subject between the date of the report and the minimum date."""
     date_time = report_date_time()
+    min_date_time = report_date_time("min_report_date")
     metric_uuids: list[MetricId] = metrics_of_subject(database, subject_uuid, date_time)
-
-    report_timestamp = datetime.fromisoformat(date_time) if date_time else datetime.now(timezone.utc)
-    min_datetime = report_timestamp - timedelta(weeks=28)
-    min_iso_timestamp = min_datetime.isoformat()
-
     return dict(
         measurements=list(
             measurements_by_metric(
-                database, *metric_uuids, min_iso_timestamp=min_iso_timestamp, max_iso_timestamp=date_time
+                database, *metric_uuids, min_iso_timestamp=min_date_time, max_iso_timestamp=date_time
             )
         )
     )
--- a/components/external_server/src/utils/functions.py
+++ b/components/external_server/src/utils/functions.py
@@ -178,9 +178,9 @@ def md5_hash(string: str) -> str:
     return md5.hexdigest()
 
 
-def report_date_time() -> str:
+def report_date_time(attribute_name: str = "report_date") -> str:
     """Return the report date requested as query parameter if it's in the past, else return an empty string."""
-    if report_date_string := dict(bottle.request.query).get("report_date"):
+    if report_date_string := dict(bottle.request.query).get(attribute_name):
         iso_report_date_string = str(report_date_string).replace("Z", "+00:00")
         if iso_report_date_string < iso_timestamp():
             return iso_report_date_string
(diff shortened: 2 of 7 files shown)


tests/feature_tests/environment.py
```diff
@@ -19,7 +19,8 @@ def get(api, headers=None, internal=False):
         """Get the resource."""
         url = api_url(api, internal)
         if context.report_date:
-            url += f"?report_date={context.report_date}"
+            sep = "&" if "?" in url else "?"
```
wkoot, line 22: using urllib.parse would be less dirty, but ok it's testcode :-)

components/frontend/src/subject/Subject.js
```diff
@@ -107,7 +107,8 @@ export function Subject({
 
     useEffect(() => {
         if (dates.length > 1) {
-            get_subject_measurements(subject_uuid, report_date).then(json => {
+            const minReportDate = dates.slice().sort((d1, d2) => {return (d1.toISOString() > d2.toISOString())}).at(0);
```
wkoot, line 110: wow magical notation

components/frontend/src/api/subject.js
```diff
@@ -1,25 +1,28 @@
...
-  return fetch_server_api('post', `subject/${subject_uuid}/attribute/${attribute}`, { [attribute]: value }).then(reload)
+    return fetch_server_api('post', `subject/${subject_uuid}/attribute/${attribute}`, { [attribute]: value }).then(reload)
 }
 
-export function get_subject_measurements(subject_uuid, date) {
-  return fetch_server_api('get', api_with_report_date(`subject/${subject_uuid}/measurements`, date))
+export function get_subject_measurements(subject_uuid, date, minDate) {
+    const minReportDate = minDate.toISOString().split("T")[0] + "T00:00:00.000Z" // Ignore the time so we get all measurements for the min date
+    let api = api_with_report_date(`subject/${subject_uuid}/measurements`, date)
+    const sep = api.indexOf("?") < 0 ? "?" : "&"
```
wkoot, line 26: Is there no library that implements this logic?

components/external_server/src/routes/subject.py
```diff
@@ -119,18 +118,14 @@ def post_subject_attribute(subject_uuid: SubjectId, subject_attribute: str, data
 
 @bottle.get("/api/v3/subject/<subject_uuid>/measurements", authentication_required=False)
 def get_subject_measurements(subject_uuid: SubjectId, database: Database):
-    """Return all measurements for the subject within the last 28 weeks."""
+    """Return all measurements for the subject between the date of the report and the minimum date."""
     date_time = report_date_time()
+    min_date_time = report_date_time("min_report_date")
```
wkoot, line 123: What happens when either of these dts is "" instead of an isodt?

components/external_server/src/utils/functions.py
```diff
@@ -178,9 +178,9 @@ def md5_hash(string: str) -> str:
     return md5.hexdigest()
 
 
-def report_date_time() -> str:
+def report_date_time(attribute_name: str = "report_date") -> str:
     """Return the report date requested as query parameter if it's in the past, else return an empty string."""
-    if report_date_string := dict(bottle.request.query).get("report_date"):
+    if report_date_string := dict(bottle.request.query).get(attribute_name):
```
wkoot, line 183: shouldn't we parse frontend data to datetime and then write it back with isoformat(), to validate input?

components/frontend/src/api/subject.js
fniessink: You mean for the date handling? moment.js maybe? Might be worth investigating at some point.

components/external_server/src/routes/subject.py
fniessink: If one or both are missing you get more measurements. For example, if the min report date is missing there is no lower boundary on the measurements query so basically you get all measurements before the report date. If both are missing you'll get all measurements. But in practice the frontend always passes a min report date.

components/external_server/src/utils/functions.py
fniessink: I guess that depends on whether we consider the API internal (i.e. just meant for the frontend) or if we also want to offer it to external consumers. I think at the moment we have only documented the API endpoints for export and import of reports, so all the other endpoints are _de facto_ for the frontend only.

components/frontend/src/subject/Subject.js
fniessink: Yeah, crappy JS. slice() is needed to make a copy because sort() alters the array. It seems to be the "canonical" way to do this.

PR #4711, opened 2022-10-11
Fix AzureDevopsUserStoryPoints

### LINKED ISSUE #4710: User story points for Azure DevOps source broken
**Describe the bug**
The default field is no longer "StoryPoints", but "Effort" - this needs to be changed in the API call.
The count scale measurement does not allow for floats, data should be cast to int before returning.

wkoot:
Closes https://github.com/ICTU/quality-time/issues/4710.

### DIFF
--- a/components/collector/src/source_collectors/performancetest_runner/performancetest_scalability.py
+++ b/components/collector/src/source_collectors/performancetest_runner/performancetest_scalability.py
@@ -15,7 +15,7 @@ class PerformanceTestRunnerScalability(PerformanceTestRunnerBaseClass):
         return str(min(breaking_points))
 
     async def _parse_total(self, responses: SourceResponses) -> Value:
-        """Override to compute the total number of users from the responses."""
+        """Override to compute the total number of virtual users from the responses."""
         breaking_points = [
             (await self.__breaking_point_vusers(response), await self.__max_vusers(response)) for response in responses
         ]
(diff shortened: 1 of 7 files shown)


components/collector/src/source_collectors/performancetest_runner/performancetest_scalability.py
```diff
@@ -15,7 +15,7 @@ async def _parse_value(self, responses: SourceResponses) -> Value:
         return str(min(breaking_points))
 
     async def _parse_total(self, responses: SourceResponses) -> Value:
-        """Override to compute the total number of vusers from the responses."""
+        """Override to compute the total number of users from the responses."""
```
fniessink: "vusers" is actually correct here, it's short for "virtual users"

```suggestion
        """Override to compute the total number of virtual users from the responses."""
```

components/shared_data_model/src/shared_data_model/sources/azure_devops.py
```diff
@@ -146,7 +146,7 @@
             metrics=["failed_jobs", "job_runs_within_time_period", "source_up_to_dateness", "unused_jobs"],
         ),
         lookback_days=Days(
-            name="Number of days to look back in selecting job builds to consider",
+            name="Number of days to look back in selecting items to consider",
```
fniessink: "items" is very abstract. Is there really not a term that's better than "job builds" and less abstract than "items"?

fniessink (approved):

PR #4767, opened 2022-10-26
When creating issues in the issue tracker from Quality-time, mention …

### LINKED ISSUE #4743: Jira audit trail
# User story

As security officer I want Quality-time to note who creates an issue in the issue tracker so that creating issues is non-repudiationable.

## Acceptance criteria

- [x] Issues created in the issue tracker contain the LDAP username and email address of the user who created the issue from Quality-time in the description.

## Assumptions

- Users may have the right to change the description of issues in the issue tracker but the issue tracker keeps the old version of the description so that there is always a way to find out who created the issue.
- We don't assume that users have the same username in LDAP and in the issue tracker, so we don't @-mention the user in the description.

# Tasks

### External server

- [x] Change endpoint `api/v3/metric/<metric_uuid>/issue/new` to add "This issue was created by {username } {email address}." in `create_issue_description(metric, subject, report) -> str` in `components/external_server/src/routes/metric.py`
- [x] Change the auth plugin in the external server to pass the username and email address to the wrapped end point if the endpoint indicates it needs the user data via een boolean flag. For example:

```python
@bottle.post("/api/v3/metric/<metric_uuid>/issue/new", permissions_required=[EDIT_REPORT_PERMISSION], pass_user=True)
def add_metric_issue(metric_uuid: MetricId, database: Database, user: User):
    ...
    issue_description = create_issue_description(metric, subject, report, user)
    ...
```

## Documentation

- [x] Update changelog

fniessink:
…the user who created the issue in the issue description for traceability. Closes #4743.

### DIFF
--- a/components/external_server/src/routes/metric.py
+++ b/components/external_server/src/routes/metric.py
@@ -10,7 +10,7 @@ from shared.database.datamodels import latest_datamodel
 from shared.database.measurements import insert_new_measurement, latest_measurement
 from shared.database.reports import insert_new_report
 from shared.model.metric import Metric
-from shared.utils.type import MetricId, SubjectId
+from shared.utils.type import MetricId, SubjectId, User
 
 from database.datamodels import default_metric_attributes
 from database.reports import latest_report_for_uuids, latest_reports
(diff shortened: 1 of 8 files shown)


components/external_server/src/routes/metric.py
```diff
@@ -169,11 +169,12 @@ def add_metric_issue(metric_uuid: MetricId, database: Database):
         return dict(ok=True, issue_url=issue_tracker.browse_url(issue_key))
 
 
-def create_issue_description(metric, subject, report) -> str:
+def create_issue_description(metric, subject, report, user: User) -> str:
```
wkoot, line 172: Where did types for the other args go?

components/external_server/tests/routes/plugins/test_route_auth_plugin.py
```diff
@@ -6,9 +6,11 @@
 import bottle
 
 from shared.routes.plugins import InjectionPlugin
+from shared.utils.type import User
 
 from routes.plugins.auth_plugin import AuthPlugin, EDIT_REPORT_PERMISSION
 
+
```
wkoot: eh?

wkoot (approved):

components/external_server/src/routes/metric.py
fniessink: Those are complicated dicts based on JSON from the database so a bit hard to create types for. Also a bit too much to try and tackle that issue as part of this story.

components/external_server/tests/routes/plugins/test_route_auth_plugin.py
fniessink: No idea what happened there, I'll remove it.

PR #4772, opened 2022-10-31
4614 add a default epic link to jira issues created from quality time

### LINKED ISSUE #4614: Add a default epic link to Jira issues created from Quality-time
# User story

As a  quality manager I want to have issues created from Quality-time to have a default epic link so I can find them in one place in the issue tracker.

## Acceptance criteria

- [ ] New issues have the configured epic link in the issue tracker.

## Assumptions

- Projects have a limited number of epics that we can present in a single choice combobox.

## Out of scope

- Pagination on the epics returned by the issue tracker.
- Showing the issue epic link in the issue cards in the Issues column.
- Dealing with changes in the issue tracker after it has been configured (e.g. issue types disappearing). If this happens, the user may see a client error. We'll deal with this if it happens frequently enough.
 
## Related issues

- #4468

# Tasks

### Frontend

- [x] Add epic link single choice field that contains the epics belonging to the selected project. Field is not mandatory and allows None as choice.
- [x] If the current issue type does not support epic links, show a yellow warning below the field with the title "Epic links not supported" and the text "The issue type '<issue type>' in project '<project key>' does not support linking issues to an epic when creating issues, so new issues will not be linked to an epic."

### External server

- [x] Retrieving the options: Extend issue tracker code to get epic links and return them as options in `def get_options(self) -> Options`. Extend the Options model to include epic links. Query the issue meta data to check whether epic links are allowed for the currently selected project and issue type.
- [x] Creating the issue: add an epic link if an epic link is configured and  epic links are allowed for the currently selected project and issue type:
```python
if epic_link and self.__epi_links_supported():  # pragma: no cover-behave
    json["fields"]["epic_link"] = self.__get_epic_link_somehow()
```

## Feature tests

- [x] Configure epic link in report.
- [x] Attempt to create an issue

## Documentation

- [x] Update change-log
- [x] Update screenshots (probably not needed)

fniessink:
Closes #4614.

### DIFF
--- a/components/external_server/src/model/issue_tracker.py
+++ b/components/external_server/src/model/issue_tracker.py
@@ -49,6 +49,7 @@ class IssueParameters:
     project_key: str
     issue_type: str
     issue_labels: Optional[list[str]] = None
+    epic_link: str = ""
 
 
 JiraIssueSuggestionJSON = dict[str, list[dict[str, str | dict[str, str]]]]
@@ -69,6 +70,7 @@ class Options(AsDictMixin):
     projects: list[Option]
     issue_types: list[Option]
     fields: list[Option]
+    epic_links: list[Option]
 
 
 @dataclass
@@ -83,6 +85,10 @@ class IssueTracker:
     issue_types_api = issue_creation_api + "/createmeta/%s/issuetypes"
     issue_browse_url = "%s/browse/%s"
     suggestions_api: str = "%s/rest/api/2/search?jql=summary~'%s~10' order by updated desc&fields=summary&maxResults=20"
+    epics_api: str = (
+        '%s/rest/api/2/search?jql=type=epic and ("Epic Status" != Done or "Epic Status" is empty) and '
+        "project=%s&fields=summary&maxResults=100"
+    )
 
     def __post_init__(self) -> None:
         """Strip any trailing slash from the URL so we can add paths without worrying about double slashes."""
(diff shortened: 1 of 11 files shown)


components/external_server/src/model/issue_tracker.py
```diff
@@ -83,6 +85,10 @@ class IssueTracker:
     issue_types_api = issue_creation_api + "/createmeta/%s/issuetypes"
     issue_browse_url = "%s/browse/%s"
     suggestions_api: str = "%s/rest/api/2/search?jql=summary~'%s~10' order by updated desc&fields=summary&maxResults=20"
+    epics_api: str = (
+        '%s/rest/api/2/search?jql=type=epic and ("Epic Status" != Done or "Epic Status" is empty) and '
+        "project=%s&fields=summary&maxResults=100"
```
wkoot, line 90: Didn't we have the max results for Jira defined as a constant somewhere?

components/external_server/src/model/issue_tracker.py
```diff
@@ -183,6 +198,19 @@ def __get_field_options(self, issue_types: list[Option]) -> list[Option]:  # pra
...
+        field_names = [field.name.lower() for field in fields]
+        if "epic link" not in field_names:
+            return []
+        api_url = self.epics_api % (self.url, self.issue_parameters.project_key)
+        epics = []
+        try:
+            epics = self.__get_json(api_url)["issues"]
+        except Exception as reason:  # pylint: disable=broad-except
+            logging.warning("Getting epics at %s failed: %s", api_url, reason)
+        return [Option(str(epic["key"]), str(f'{epic["fields"]["summary"]} ({epic["key"]})')) for epic in epics]
```
wkoot, line 212: Is there always a field summary? Is this the epic title/header?

wkoot (approved):

components/external_server/src/model/issue_tracker.py
fniessink: Yes, because we asked for the summary field via the `epics_api`.

fniessink: And it is the long version of the epic title. There is also a short version that Jira uses when it displays the epic as label, but it seemed more informative to use the long version in the epic link dropdown menu. The dropdown menu also shows the epic key by the way. Also, the epic summary is a mandatory field in Jira: when you create a new epic you have to enter both a short title and a summary.

components/external_server/src/model/issue_tracker.py
fniessink: Yes, but that constant lives in the collector. I could move it to the data model, I'll check if that's feasible.

fniessink: We should really get the maxResults from Jira instead of hardcoding it. But it's not immediately clear how to do that. I'll add an issue to the backlog.

PR #4898, opened 2022-11-18
Support GitLab CI pipelines as source for the source-up-to-dateness metric

### LINKED ISSUE #3927: Allow for using a GitLab CI-pipeline as source for the 'source up-to-dateness' metric
GitLab already supports the source-up-to-dateness metric for files in GitLab. Use the same solution as for Azure DevOps: present both file path and pipeline parameters to the user and indicate in the tooltip that files get preference.

Note: Jenkins and Azure DevOps don't need to be changed. In case of Jenkins the source up-to-dateness of jobs can be measured (job artifacts have the same date, so no need to configure artifacts separately). In case of Azure DevOps Quality-time already supports both files and pipelines.

Reported by @Sebastiaan127001

fniessink:
Closes #3927.

### DIFF
--- a/components/collector/src/source_collectors/gitlab/source_up_to_dateness.py
+++ b/components/collector/src/source_collectors/gitlab/source_up_to_dateness.py
@@ -84,14 +84,14 @@ class GitLabPipelineUpToDateness(TimePassedCollector, GitLabJobsBase):
         return await super()._landing_url(responses)
 
     async def _parse_source_response_date_time(self, response: Response) -> datetime:
-        """Override to get the date and time of the commit or the pipeline."""
+        """Override to get the date and time of the pipeline."""
         jobs = await self._jobs(SourceResponses(responses=[response]))
         build_dates = [self._build_date(job) for job in jobs]
         return datetime.combine(max(build_dates, default=datetime.min.date()), datetime.min.time())
 
 
 class GitLabSourceUpToDateness(SourceCollector, ABC):
-    """Factory class to create a collector to get the up-to-dateness of either jobs or files."""
+    """Factory class to create a collector to get the up-to-dateness of either pipelines or files."""
 
     def __new__(cls, session: aiohttp.ClientSession, source):
         """Create an instance of either the file up-to-dateness collector or the jobs up-to-dateness collector."""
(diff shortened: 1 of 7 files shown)


components/collector/src/source_collectors/gitlab/source_up_to_dateness.py
```diff
@@ -68,3 +72,35 @@ async def _parse_value(self, responses: SourceResponses) -> Value:
...
+class GitLabPipelineUpToDateness(TimePassedCollector, GitLabJobsBase):
+    """Collector class to measure the up-to-dateness of a job/pipeline."""
+
+    async def _landing_url(self, responses: SourceResponses) -> URL:
+        """Override to return a landing URL for the pipeline."""
+        if responses and (json := await responses[0].json()):
+            return URL(json[0]["pipeline"]["web_url"])
+        return await super()._landing_url(responses)
+
+    async def _parse_source_response_date_time(self, response: Response) -> datetime:
```
wkoot, line 86: Is it not easier to return a `datetime.date` object, as we only report on amount of days anyway?

components/collector/tests/source_collectors/gitlab/test_source_up_to_dateness.py
```diff
@@ -46,7 +50,22 @@ async def test_source_up_to_dateness_folder(self):
...
+        """Test that the age of a pipeline can be measured."""
+        self.set_source_parameter("file_path", "")
+        build_date = datetime.fromisoformat(self.gitlab_jobs_json[0]["created_at"].strip("Z")).date()
+        expected_age = (date.today() - build_date).days
+        with self.patched_client_session_head():
+            response = await self.collect(get_request_json_return_value=self.gitlab_jobs_json)
+        self.assert_measurement(response, value=str(expected_age), landing_url="https://gitlab/project/-/pipelines/1")
+
+    async def test_file_landing_url_on_failure(self):
+        """Test that the landing url is the API url when GitLab cannot be reached."""
```
wkoot, line 63: Where is the default value for `file_path` set?

wkoot (approved):

components/collector/src/source_collectors/gitlab/source_up_to_dateness.py
fniessink: It is mostly use to report duration in days, that's correct, but returning dates here would amount to premature rounding, so not changing.

components/collector/tests/source_collectors/gitlab/test_source_up_to_dateness.py
fniessink: In the superclass of the test class.

PR #4944, opened 2022-11-28
Make log level of Python backend containers configurable via environment variables

### LINKED ISSUE #4943: Make loglevel configurable via env variable
**Describe the bug**
The log level of the collector is hardcoded.

**Expected behaviour**
Log level should be changeable via an environment variable.

**Additional context**
Also check other containers.

fniessink:
Closes #4943.

### DIFF
--- a/components/collector/src/quality_time_collector.py
+++ b/components/collector/src/quality_time_collector.py
@@ -11,10 +11,9 @@ import source_collectors  # pylint: disable=unused-import # lgtm [py/unused-impo
 from base_collectors import Collector
 
 
-async def collect(log_level: int | str | None = None) -> NoReturn:
+async def collect() -> NoReturn:
     """Collect the measurements indefinitely."""
-    if log_level is None:
-        log_level = str(os.getenv("COLLECTOR_LOG_LEVEL", "WARNING"))
+    log_level = str(os.getenv("COLLECTOR_LOG_LEVEL", "WARNING"))  # cast to str to keep mypy happy
     logging.getLogger().setLevel(log_level)
     await Collector().start()
 
--- a/components/collector/src/quality_time_collector.py
+++ b/components/collector/src/quality_time_collector.py
@@ -2,6 +2,8 @@
 
 import asyncio
 import logging
+import os
+from typing import NoReturn
 
 # Make sure subclasses are registered
 import metric_collectors  # pylint: disable=unused-import # lgtm [py/unused-import]
@@ -9,11 +11,13 @@ import source_collectors  # pylint: disable=unused-import # lgtm [py/unused-impo
 from base_collectors import Collector
 
 
-async def collect(log_level: int | None = None) -> None:
+async def collect(log_level: int | str | None = None) -> NoReturn:
     """Collect the measurements indefinitely."""
-    logging.getLogger().setLevel(logging.ERROR if log_level is None else log_level)
+    if log_level is None:
+        log_level = str(os.getenv("COLLECTOR_LOG_LEVEL", "WARNING"))
+    logging.getLogger().setLevel(log_level)
     await Collector().start()
 
 
 if __name__ == "__main__":
-    asyncio.run(collect(logging.WARNING))  # pragma: no cover
+    asyncio.run(collect())  # pragma: no cover
--- a/docs/src/deployment.md
+++ b/docs/src/deployment.md
@@ -136,13 +136,16 @@ The options for configuring logging are limited at the moment. The MongoDB daemo
     command: --quiet
 ```
 
-Other logging settings are currently fixed. These include:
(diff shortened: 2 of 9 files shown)


components/collector/src/quality_time_collector.py
```diff
@@ -2,18 +2,22 @@
...
 import source_collectors  # pylint: disable=unused-import # lgtm [py/unused-import]
 from base_collectors import Collector
 
 
-async def collect(log_level: int | None = None) -> None:
+async def collect(log_level: int | str | None = None) -> NoReturn:
     """Collect the measurements indefinitely."""
-    logging.getLogger().setLevel(logging.ERROR if log_level is None else log_level)
+    if log_level is None:
+        log_level = str(os.getenv("COLLECTOR_LOG_LEVEL", "WARNING"))
```
wkoot: Why cast to string, isn't it already always a string?

components/collector/src/quality_time_collector.py
```diff
@@ -2,18 +2,22 @@
...
+    logging.getLogger().setLevel(log_level)
```
wkoot, line 17: Would `log_level or os.getenv("COLLECTOR_LOG_LEVEL", "WARNING")` not suffice?

components/collector/src/quality_time_collector.py
```diff
@@ -2,18 +2,22 @@
...
+from typing import NoReturn
 
 # Make sure subclasses are registered
 import metric_collectors  # pylint: disable=unused-import # lgtm [py/unused-import]
 import source_collectors  # pylint: disable=unused-import # lgtm [py/unused-import]
 from base_collectors import Collector
 
 
-async def collect(log_level: int | None = None) -> None:
+async def collect(log_level: int | str | None = None) -> NoReturn:
```
wkoot: Is `collect` ever called with the `log_level` param as int? Can the parameter be removed in favour of the env var?

docs/src/deployment.md
```diff
@@ -136,13 +136,16 @@ The options for configuring logging are limited at the moment. The MongoDB daemo
     command: --quiet
 ```
 
-Other logging settings are currently fixed. These include:
+The collector, external server, and internal server all have log level `WARNING` as default. This can be overridden by setting an environment variable to `DEBUG`, `INFO`, `WARNING`, `CRITICAL` or `ERROR`.
```
wkoot: Both the string and numeric values listed in https://docs.python.org/3/library/logging.html#levels should be possible

wkoot: The notifier is also set to log level WARNING

docs/src/software.md
```diff
@@ -67,18 +67,19 @@ The [Dockerfile](https://github.com/ICTU/quality-time/blob/master/components/ext
...
-| `LDAP_ROOT_DN` | `dc=example,dc=org` | LDAP root distinguished name.                                                                                                                                                                                                                                   |
-| `LDAP_LOOKUP_USER_DN` | `cn=admin,dc=example,dc=org` | LDAP lookup user distinguished name.                                                                                                                                                                                                            |
-| `LDAP_LOOKUP_USER_PASSWORD` | `admin` | LDAP lookup user password.                                                                                                                                                                                                                                     |
+| Name | Default value | Description |
+| :--- | :------------ |:------------|
+| `EXTERNAL_SERVER_PORT` | `5001` | Port of the external server. |
+| `EXTERNAL_SERVER_LOG_LEVEL` | `WARNING` | Log level. Allowed values are `DEBUG`, `INFO`, `WARNING`, `CRITICAL`, and `ERROR`. |
```
wkoot: The order of critical and error should be switched

components/collector/src/quality_time_collector.py
fniessink: Mypy doesn't understand that if the default for getenv is a string, getenv will never return None.

fniessink: I'll add a comment.

components/collector/src/quality_time_collector.py
fniessink: That would make it impossible to pass 0 as log level, which is an allowed value, see https://docs.python.org/3/library/logging.html#logging-levels.

docs/src/deployment.md
fniessink: Disagree. This is our "interface" so we can simply decide which values we allow.

docs/src/software.md
fniessink: Good catch!

components/collector/src/quality_time_collector.py
fniessink: I assumed it was, from the unit tests. But assumptions are the mother of all fuckups...

PR #5221, opened 2023-01-09
Add SourceCollector._include_entity

### LINKED ISSUE #4714: Refactor collector classes to all use the same logic to decide which entities to include
Refactor all collector classes to implement at method:

```python
class SourceCollector:
   ...
   def _include_entity(self, entity) -> bool:
       """Return whether to include the entity in the measurement."""
       return True
```

For the Bandit collector we get something like this:
```python
class BanditSecurityWarnings(JSONFileSourceCollector):
    """Bandit collector for security warnings."""

    def _parse_json(self, json: JSON, filename: str) -> Entities:
        """Override to parse the security warnings."""
        return Entities(
            [
                Entity(
                    key=f'{warning["test_id"]}:{warning["filename"]}:{warning["line_number"]}',
                    location=f'{warning["filename"]}:{warning["line_number"]}',
                    issue_text=warning["issue_text"],
                    issue_severity=warning["issue_severity"].capitalize(),
                    issue_confidence=warning["issue_confidence"].capitalize(),
                    more_info=warning["more_info"],
                )
                for warning in cast(JSONDict, json).get("results", [])
                if self._include_entity(warning)
            ]
        )

    def _include_entity(self, source_entity_data) -> bool:
        if not super()._include_entity(source_entity_data):
            return False
        severities = self._parameter("severities")
        confidence_levels = self._parameter("confidence_levels")
        return source_entity_data["issue_severity"].lower() in severities and source_entity_data["issue_confidence"].lower() in confidence_levels
```

wkoot:
Closes #4714

### DIFF
--- a/components/collector/src/base_collectors/source_collector.py
+++ b/components/collector/src/base_collectors/source_collector.py
@@ -153,8 +153,9 @@ class SourceCollector(ABC):
         Either this method or self._parse_entities() need to be overridden in the subclass to implement the actual
         parsing of the source responses.
         """
+        included_entities = [entity for entity in await self._parse_entities(responses) if self._include_entity(entity)]
         return SourceMeasurement(
-            entities=await self._parse_entities(responses),
+            entities=Entities(included_entities),
             total=await self._parse_total(responses),
             value=await self._parse_value(responses),
         )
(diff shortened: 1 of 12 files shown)


wkoot:
Open questions:
- AxeAccessibilityCollector checks violations on impact and tags
- AzureDevopsJobs + AzureDevopsPipelines do include checks on Job, before making Entities
- AzureDevopsIssues do include checks on work item, before making Entities
- what about TransactionEntity.is_to_be_included and SlowTransactionsCollector._is_to_be_included_and_is_slow ?
- GitLabJobsBase does include checks on Job, before making Entities
- GitLabPipelineUpToDateness checks pipeline status and trigger, but only returns aggregate
- JenkinsJobRunsWithinTimePeriod has both _include_job (from parent) and its own _include_build
- JenkinsJobs yields jobs based on include checks on Job, before making Entities
- JiraIssues _include_issue omits items which cannot be parsed by current logic, which needs to be implemented
- QualityTimeMetrics sneakily also does __update_entity when an entity is selected for inclusion
- TestNGTests checks both beforeClass/afterClass and has complicated test_results_to_count logic
- TrelloBase implementations have a different structure but both use a __ignore_card

fniessink (approved):

PR #5271, opened 2023-01-24
Implement _include_entity for all source collectors

### LINKED ISSUE #5240: Implement _include_entity for all source collectors
# User story

Handle the open points from #4714, as listed in #5221

# Tasks

## Collector

- [x] AxeAccessibilityCollector checks violations on impact and tags
- [x] AzureDevopsJobs + AzureDevopsPipelines do include checks on Job, before making Entities
- [x] AzureDevopsIssues do include checks on work item, before making Entities
- [x] what about TransactionEntity.is_to_be_included and SlowTransactionsCollector._is_to_be_included_and_is_slow ?
- [x] GitLabJobsBase does include checks on Job, before making Entities
- [x] GitLabPipelineUpToDateness checks pipeline status and trigger, but only returns aggregate
- [x] JenkinsJobRunsWithinTimePeriod has both _include_job (from parent) and its own _include_build
- [x] JenkinsJobs yields jobs based on include checks on Job, before making Entities
- [x] JiraIssues _include_issue omits items which cannot be parsed by current logic, which needs to be implemented
- [x] QualityTimeMetrics sneakily also does __update_entity when an entity is selected for inclusion
- [x] ~~TestNGTests checks both beforeClass/afterClass and has complicated test_results_to_count logic~~
- [x] TrelloBase implementations have a different structure but both use a __ignore_card

wkoot:
Closes #5240


components/collector/src/source_collectors/trello/issues.py
```diff
@@ -19,30 +19,33 @@ async def _parse_entities(self, responses: SourceResponses) -> Entities:
...
         if "overdue" in cards_to_count and card_is_overdue():
-            return False
+            return True
         if "inactive" in cards_to_count and card_is_inactive():
-            return False
-        return bool(cards_to_count)
+            return True
+        # TODO - warrants comment or logic change: do we only include if we don't count overdue nor inactive?
+        #  Doesn't this mean that inactive and overdue cards are always counted when we are not supposed to count them?
+        return not bool(cards_to_count)
```
fniessink: If you change a ignore() method to include() and the ignore() method has two checks, you should not only invert the checks, but also reverse to order of the steps to get the exact same semantics I think? I should write a test program, but I'm too lazy ;-)

wkoot: I tried the following approach but it results in several failing tests (even if I keep the botton return as `not bool`):
```
if "overdue" not in cards_to_count and card_is_overdue():
  return False
if "inactive" not in cards_to_count and card_is_inactive():
  return False
return True
```

wkoot: Ah, I see the issue.

wkoot: Oh. Setting the parameter means _only_ count that type, not inclusive with any type

fniessink (approved):

PR #5287, opened 2023-01-26
Add options to the technical debt dropdown menu to set and clear acce…

### LINKED ISSUE #5125: Add an option "No, and reset technical debt end date, clear value, and remove linked issues"
Add an option "No, and reset technical debt end date, clear value" to the "Accept technical debt" dropdown to also:
- Reset the technical debt value.
- Clear the end date.

Open issues:
- What to with issues that have not been resolved and/or closed? Alternatives considered:
  - ~~Maybe add a comment to the issue that the technical debt was removed in Quality-time?~~
  - ~~Close or resolve the issue from Quality-time?~~
  - ~~Refuse to perform this action, since not all linked issues have been resolved and/or closed.~~
  - ~~Make this "No, and reset technical debt end date, clear value, and remove *closed/resolved* linked issues"~~
  - ~~Have a separate button to remove closed/resolved or all issues and not include it in this option.~~

Related issue: #5127.

fniessink:
…pted technical debt including the technical debt target and technical debt end date.

### DIFF
--- a/components/external_server/src/model/report.py
+++ b/components/external_server/src/model/report.py
@@ -27,5 +27,11 @@ class Report(SharedReport):
     def desired_response_time(self, status: str) -> int:
         """Return the desired response time for the metric status."""
         # Note that the frontend also has these constant, in src/defaults.js.
-        default = dict(debt_target_met=60, target_not_met=7, unknown=3, near_target_met=21).get(status, 3)
+        defaults = dict(
+            debt_target_met=60,
+            near_target_met=21,
+            target_not_met=7,
+            unknown=3,
+        )
+        default = defaults.get(status, defaults["unknown"])
         return int(self.get("desired_response_times", {}).get(status, default))
--- a/components/external_server/src/model/report.py
+++ b/components/external_server/src/model/report.py
@@ -23,3 +23,9 @@ class Report(SharedReport):
             parameters.get("username", ""), parameters.get("password", ""), parameters.get("private_token", "")
         )
         return IssueTracker(url, issue_parameters, credentials)
+
+    def desired_response_time(self, status: str) -> int:
+        """Return the desired response time for the metric status."""
+        # Note that the frontend also has these constant, in src/defaults.js.
+        default = dict(debt_target_met=60, target_not_met=7, unknown=3, near_target_met=21).get(status, 3)
+        return int(self.get("desired_response_times", {}).get(status, default))
--- a/components/external_server/src/routes/metric.py
+++ b/components/external_server/src/routes/metric.py
@@ -154,7 +154,9 @@ def post_metric_debt(metric_uuid: MetricId, database: Database):
     report = latest_report_for_uuids(latest_reports(database, data_model), metric_uuid)[0]
     metric, subject = report.instance_and_parents_for_uuid(metric_uuid=metric_uuid)
     if new_accept_debt:
+        # Get the latest measurement to get the current metric value:
         latest = latest_measurement(database, Metric(data_model, metric, metric_uuid))
+        # Only if the metric has at least one measurement can a technical debt target be set:
         new_debt_target = latest.value() if latest else None
         new_end_date = (date.today() + timedelta(days=report.desired_response_time("debt_target_met"))).isoformat()
     else:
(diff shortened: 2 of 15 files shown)


components/frontend/src/metric/MetricDebtParameters.js
```diff
@@ -21,11 +21,21 @@ function AcceptTechnicalDebt({ metric, metric_uuid, reload }) {
...
-                { key: false, text: "No", value: false }]}
-            set_value={(value) => set_metric_attribute(metric_uuid, "accept_debt", value, reload)}
+                { key: "yes", text: "Yes", value: "yes" },
+                { key: "yes+", text: "Yes, and also set technical debt target and end date", value: "yes+" },
+                { key: "no", text: "No", value: "no" },
+                { key: "no+", text: "No, and also clear technical debt target and end date", value: "no+" }
+            ]}
+            set_value={(value) => {
+                const acceptDebt = value.startsWith("yes")
+                if (value.endsWith("+")) {
```
wkoot: Do all (supported) browsers accept this in a keyfield?

components/frontend/src/defaults.js
```diff
@@ -1 +1,2 @@
-export const metricReactionDeadline = {target_not_met: 7, unknown: 3, near_target_met: 21}
+// Note that the external server also has these constants, in src/model/report.py.
+export const metricReactionDeadline = {debt_target_met: 60, target_not_met: 7, unknown: 3, near_target_met: 21}
```
wkoot: Perhaps handy to put each var on a newline so future diffs are cleaner

components/external_server/src/model/report.py
```diff
@@ -23,3 +23,9 @@ def issue_tracker(self) -> IssueTracker:
             parameters.get("username", ""), parameters.get("password", ""), parameters.get("private_token", "")
         )
         return IssueTracker(url, issue_parameters, credentials)
+
+    def desired_response_time(self, status: str) -> int:
+        """Return the desired response time for the metric status."""
+        # Note that the frontend also has these constant, in src/defaults.js.
+        default = dict(debt_target_met=60, target_not_met=7, unknown=3, near_target_met=21).get(status, 3)
```
wkoot: Is there no central location to define such constants?

wkoot: If the `.get` fallback is 3 to be the same as `unknown`, then perhaps it makes more sense to do `dict.get(status) or dict["unknown"]`

components/external_server/src/routes/metric.py
```diff
@@ -145,6 +146,37 @@ def post_metric_attribute(metric_uuid: MetricId, metric_attribute: str, database
...
+@bottle.post("/api/v3/metric/<metric_uuid>/debt", permissions_required=[EDIT_REPORT_PERMISSION])
+def post_metric_debt(metric_uuid: MetricId, database: Database):
+    """Turn the technical debt on or off, including technical debt target and end date."""
+    new_accept_debt = dict(bottle.request.json)["accept_debt"]
+    data_model = latest_datamodel(database)
+    report = latest_report_for_uuids(latest_reports(database, data_model), metric_uuid)[0]
+    metric, subject = report.instance_and_parents_for_uuid(metric_uuid=metric_uuid)
+    if new_accept_debt:
+        latest = latest_measurement(database, Metric(data_model, metric, metric_uuid))
+        new_debt_target = latest.value() if latest else None
```
wkoot, line 160: Huh, for which scenario is the `else None`?

components/external_server/src/routes/metric.py
```diff
@@ -145,6 +146,37 @@ def post_metric_attribute(metric_uuid: MetricId, metric_attribute: str, database
...
+    old_accept_debt = metric.get("accept_debt") or False
+    old_debt_target = metric.get("debt_target")
+    old_end_date = metric.get("debt_end_date")
+    if old_accept_debt == new_accept_debt and old_debt_target == new_debt_target and old_end_date == new_end_date:
+        return dict(ok=True)  # Nothing to do
+    metric.update(accept_debt=new_accept_debt, debt_target=new_debt_target, debt_end_date=new_end_date)
+    description = (
+        f"{{user}} changed the technical debt for metric '{metric.name}' of subject "
+        f"""'{subject.name}' in report '{report.name}' from '{"accepted" if old_accept_debt else "not accepted"}' """
+        f"""to '{"accepted" if new_accept_debt else "not accepted"}'."""
```
wkoot: This may state that debt is changed from "accepted to accepted", where the check is also performed on target and date

components/external_server/tests/routes/test_metric.py
```diff
@@ -356,6 +361,109 @@ def test_post_position_last_next(self, request):
...
+        self.report["subjects"][SUBJECT_ID]["metrics"][METRIC_ID]["debt_target"] = "100"
+        request.json = dict(accept_debt=False)
+        self.assertDictEqual(dict(ok=True), post_metric_debt(METRIC_ID, self.database))
+        updated_report = self.database.reports.insert_one.call_args[0][0]
+        self.assert_delta(
+            "technical debt for metric 'name' of subject 'Subject' in report 'Report' from "
+            "'not accepted' to 'not accepted'",
+            report=updated_report,
+        )
+        self.assertFalse(updated_report["subjects"][SUBJECT_ID]["metrics"][METRIC_ID]["accept_debt"])
```
wkoot, line 444: Shouldn't the value of `debt_target` (also) be asserted?

components/frontend/src/metric/MetricDebtParameters.js
fniessink: I haven't checked, but I'll use a simple string just to be sure (for example `yes_and_change_debt_attributes`)

components/external_server/src/model/report.py
fniessink: > Is there no central location to define such constants?

No, we don't share code between the frontend and the backend containers (yet). 

> If the .get fallback is 3 to be the same as unknown, then perhaps it makes more sense to do dict.get(status) or dict["unknown"]

Yes.

components/external_server/src/routes/metric.py
fniessink: For when the metric has no measurements yet, and thus no value that we can use as technical debt target. I'll add a comment.

components/external_server/src/routes/metric.py
fniessink: That's correct. I'll make it more complete.

PR #5505, opened 2023-02-23
Configuring a custom baseimage for www container

### LINKED ISSUE #5312: Make proxy base image configurable
`ARG  IMAGE_NAME=registry.access.redhat.com/ubi8/nginx-120`

Related issues: #5310 and #5311.

wkoot:
Closes #5312


wkoot:
I cannot actually test with `IMAGE_NAME=registry.access.redhat.com/ubi8/nginx-120`, but `IMAGE_NAME: nginx:stable-perl` as documented works. The dockerfile does assume nginx, but I don't think this needs to be documented explicitly as it is susceptible to documentation rot.

fniessink (approved):

PR #5524, opened 2023-02-28
Update default value in calendar date source upon restart of external…

### LINKED ISSUE #5448: have the calendar date source start at today
When adding the calendar date source, the default value is in 2021. When setting the date, the user must first change the year, which can be tedious, especially if you need to do this by clicking.

It would be useful if the calendar date started at today.

**Additional context**
image

EDIT
What still needs to happen to complete this issue:
- change data model so that it understands dynamic values in addition to static values
- change default value for calendar date source in today's date

Sebastiaan127001:
… server. Partially fixes [#5448]


Sebastiaan127001:
This is only a partial fix!

PR #5587, opened 2023-03-07
Refactor notifier

fniessink:
Refactor the notifier to not use "summarised" report data, but to get the measurements directly from a (new) internal server measurements end point. This is one step towards getting rid of the servers adding measurements to reports.

### DIFF
--- a/components/notifier/src/notifier/notifier.py
+++ b/components/notifier/src/notifier/notifier.py
@@ -2,7 +2,6 @@
 
 import asyncio
 import logging
-import os
 from datetime import datetime, timezone
 from typing import NoReturn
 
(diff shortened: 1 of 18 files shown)


components/notifier/src/notifier/notifier.py
```diff
@@ -0,0 +1,70 @@
...
+
+import aiohttp
+
+from destinations.ms_teams import notification_text, send_notification
+from strategies.notification_strategy import NotificationFinder
+
+
+async def notify() -> NoReturn:
+    """Notify our users periodically of the number of red metrics."""
+    log_level = str(os.getenv("NOTIFIER_LOG_LEVEL", "WARNING"))
```
wkoot: Does it not make more sense to read/set these args in `quality_time_notifier.py` and pass as param to notify?

PR #5590, opened 2023-03-08
Fix: Victory Bar charts animation were defered and not showing, this fixes it

ptrr:
By explicitly adding the prop for animation on VictoryBar component the chart starts the animation immediately.

This seems to be a bug in Victory itself. Related issue: https://github.com/FormidableLabs/victory/issues/2104


fniessink (approved):
I refactored your solution a little bit so that the bar chart and the pie chart use the same animate settings.

ptrr:
Looks good! :) Thanks for the addition.

PR #5784, opened 2023-03-28
Query Jira API for maxResults param

### LINKED ISSUE #4789: Get the maxResults Jira parameter from Jira itself instead of hardcoding it
The maxResults parameter is used in the collector en the external server (IssueTracker class).

wkoot:
Closes #4789

### DIFF
--- a/components/collector/src/source_collectors/jira/issues.py
+++ b/components/collector/src/source_collectors/jira/issues.py
@@ -13,7 +13,8 @@ class JiraIssues(JiraBase):
     """Jira collector for issues."""
 
     SPRINT_NAME_RE = re.compile(r",name=(.*),startDate=")
-    MAX_RESULTS = 500  # Maximum number of issues to retrieve per page. Jira allows at most 500.
+    DEFAULT_MAX_RESULTS: int = 500  # Fallback for maximum number of issues to retrieve per page from Jira
+    max_results: int | None = None
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


components/collector/src/source_collectors/jira/issues.py
```diff
@@ -115,3 +117,12 @@ def __get_sprint_names(cls, sprint_texts: list[str]) -> str:
         matches = [cls.SPRINT_NAME_RE.search(sprint_text) for sprint_text in sprint_texts]
         sprint_names = [match.group(1) for match in matches if match]
         return ", ".join(sorted(sprint_names))
+
+    async def _determine_max_results(self) -> int:
+        """Maximum number of issues to retrieve per page."""
+        if isinstance(self.MAX_RESULTS, int):
+            return self.MAX_RESULTS
```
fniessink: Can't we make this a cached property so that `self.MAX_RESULTS` isn't needed anymore?

fniessink (changes requested):

components/collector/src/source_collectors/jira/issues.py
wkoot: Well, it's a bit hairy due to the asyncio coroutine :( I originally wanted to fetch it in `__init__` but then I need to use a different url library, such as requests. I don't think we have an async `cached_property` version somewhere in `collector_utilities`, is this desirable?

components/collector/src/source_collectors/jira/issues.py
```diff
@@ -115,3 +118,12 @@ def __get_sprint_names(cls, sprint_texts: list[str]) -> str:
         matches = [cls.SPRINT_NAME_RE.search(sprint_text) for sprint_text in sprint_texts]
         sprint_names = [match.group(1) for match in matches if match]
         return ", ".join(sorted(sprint_names))
+
+    async def _determine_max_results(self) -> int:
+        """Maximum number of issues to retrieve per page."""
+        if isinstance(self.max_results, int):  # NB: using cached_property is not doable on asyncio coroutine
+            return self.max_results  # type: ignore[unreachable]
```
fniessink: Maybe the mypy warning can be gotten rid of by typing max_results as None or int?

```python
    max_results: int | None = None
```

fniessink (approved):

components/collector/src/source_collectors/jira/issues.py
wkoot: Excellent, seems to work.

PR #5785, opened 2023-03-28
Add the possibility in missing metrics metrics to exclude certain sub…

### LINKED ISSUE #5119: Have the possibility in missing metrics metrics to exclude certain subject from a report.
Add a parameter "Subjects to ignore (Subject names or identifiers)".

The Quality-time missing metrics collector will ignore metrics that are missing in the list of subjects to ignore.

Submitted by @Sebastiaan127001

wkoot:
…ject from a report

Closes #5119

### DIFF
--- a/components/collector/src/source_collectors/quality_time/missing_metrics.py
+++ b/components/collector/src/source_collectors/quality_time/missing_metrics.py
@@ -2,6 +2,7 @@
 
 from typing import cast
 
+from collector_utilities.functions import match_string_or_regular_expression
 from collector_utilities.type import URL
 from model import SourceMeasurement, SourceResponses
 from model.entity import Entities, Entity
@@ -27,7 +28,15 @@ class QualityTimeMissingMetrics(QualityTimeCollector):
         landing_url = await self._landing_url(responses)
         total = self.__nr_of_possible_metric_types(data_model, reports)
         entities = self.__missing_metric_type_entities(data_model, reports, landing_url)
-        return SourceMeasurement(total=str(total), entities=entities)
+        included_entities = Entities(entity for entity in entities if self._include_entity(entity))
+        return SourceMeasurement(total=str(total), entities=included_entities)
+
+    def _include_entity(self, entity: Entity) -> bool:
+        """Return whether to include the entity in the measurement."""
+        subjects_to_ignore = self._parameter("subjects_to_ignore")
+        if len(subjects_to_ignore) == 0:
+            return True
+        return not match_string_or_regular_expression(entity["subject"], subjects_to_ignore)
 
     def __nr_of_possible_metric_types(self, data_model: dict, reports: list[dict]) -> int:
         """Return the number of possible metric types in the reports."""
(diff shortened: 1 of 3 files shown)


components/collector/src/source_collectors/quality_time/missing_metrics.py
```diff
@@ -27,7 +28,15 @@ async def _parse_source_responses(self, responses: SourceResponses) -> SourceMea
...
         total = self.__nr_of_possible_metric_types(data_model, reports)
         entities = self.__missing_metric_type_entities(data_model, reports, landing_url)
-        return SourceMeasurement(total=str(total), entities=entities)
+        included_entities = Entities(entity for entity in entities if self._include_entity(entity))
+        return SourceMeasurement(total=str(total), entities=included_entities)
+
+    def _include_entity(self, entity: Entity) -> bool:
+        """Return whether to include the entity in the measurement."""
+        subjects_to_ignore = self._parameter("subjects_to_ignore")
+        if len(subjects_to_ignore) == 0:
```
fniessink:
```suggestion
        if not subjects_to_ignore:
```

components/collector/tests/source_collectors/quality_time/test_missing_metrics.py
```diff
@@ -90,3 +90,11 @@ async def test_nr_of_missing_metrics_without_correct_report(self):
         self.reports["reports"] = []
         response = await self.collect(get_request_json_side_effect=[self.data_model, self.reports])
         self.assert_measurement(response, parse_error="No reports found with title or id")
+
+    async def test_subjects_to_ignore(self):
+        """Test that the number of non-ignored missing metrics is returned."""
+        self.set_source_parameter("subjects_to_ignore", ["S2"])
+        response = await self.collect(get_request_json_side_effect=[self.data_model, self.reports])
+        self.assert_measurement(
+            response, value=str(int(len(self.entities) / 2)), total=self.expected_software_metrics  # TODO - total/2?
```
fniessink: Why the TODO? The total doesn't change if the user decides to ignore subjects, so the TODO is not needed, I'd say.

components/shared_data_model/src/shared_data_model/sources/quality_time.py
```diff
@@ -236,6 +236,13 @@
             },
             metrics=["metrics"],
         ),
+        subjects_to_ignore=MultipleChoiceWithAdditionParameter(
+            name="Subjects to ignore (Subject names or identifiers)",
```
fniessink:
```suggestion
            name="Subjects to ignore (subject names or identifiers)",
```

fniessink (changes requested):
The help text says the user can enter subject names or identifiers to filter subjects, but the _include_entity() method only filters by name as far as I can tell?

components/collector/tests/source_collectors/quality_time/test_missing_metrics.py
wkoot: That was also what I thought, but I wanted to double-check :-)

wkoot:
> The help text says the user can enter subject names or identifiers to filter subjects, but the _include_entity() method only filters by name as far as I can tell?

Yes, I'm not entirely sure in which way one would query the identifiers - within the Entity we currently only have access to urls (and uuids), subject_type and metric_type. It was indeed something for the review :-)

fniessink (approved):

PR #5845, opened 2023-04-08
Bars for subjects or tags with little metrics in the dashboard would …

### LINKED ISSUE #5792: Bar charts are unreadable with a minimum (<3) number of metrics in that subject or with that label
**Describe the bug**
When selecting (or comparing) 2 or more dates, the bar charts are unreadable when there are a minimum (<3) number of metrics in that subject or with that label.

**To Reproduce**
1. Select 2 or more number of dates (Settings - Number of dates)
2. Optional select "Time between dates"
3. See bar charts 

**Expected behaviour**
Shows all bar charts with a fixed width or a minimum (readable) width. 
Suggestion: use the width as currently used for 20 metrics.

**Screenshots**
image

fniessink:
…get too narrow to be legible. Fixes #5792.

### DIFF
--- a/components/frontend/src/dashboard/StatusBarChart.js
+++ b/components/frontend/src/dashboard/StatusBarChart.js
@@ -1,21 +1,28 @@
 import React from 'react';
-import { VictoryBar, VictoryContainer, VictoryStack } from 'victory';
+import { VictoryBar, VictoryStack } from 'victory';
 import { pluralize, STATUSES, STATUS_COLORS, STATUS_NAME, sum } from '../utils';
 
-function nr_metrics_label(nr_metrics) {
-    return nr_metrics === 0 ? "no metrics" : nr_metrics + pluralize(" metric", nr_metrics)
+function nrMetricsLabel(nrMetrics) {
+    return nrMetrics === 0 ? "No metrics" : nrMetrics + pluralize(" metric", nrMetrics)
 }
 
-export function StatusBarChart({ animate, colors, label, tooltip, summary, maxY, style }) {
+export function StatusBarChart({ animate, colors, label, tooltip, summary, maxY, style, width, height }) {
     const nrMetrics = sum(summary[Object.keys(summary)[0]])
-    const barRatio = Math.max(5, nrMetrics) / maxY
+    const nrDates = Object.keys(summary).length
+    // Calculate how many metrics this chart displays compared to the chart with the most metrics.
+    // The ratio is used below to set the bar width so that bars in charts that represent fewer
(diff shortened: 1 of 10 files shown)


components/frontend/src/dashboard/StatusBarChart.js
```diff
@@ -19,7 +21,7 @@ export function StatusBarChart({ animate, colors, label, tooltip, summary, maxY,
         })
         return (
             <VictoryBar
-                barRatio={barRatio}
+                barWidth={(props) => ratio * (props.width - nrDates * 2 * props.style.padding - 2 * horizontalPadding) / nrDates}
```
wkoot: Does this work the same on retina and regular displays?

components/frontend/src/dashboard/StatusBarChart.js
```diff
@@ -33,18 +35,15 @@ export function StatusBarChart({ animate, colors, label, tooltip, summary, maxY,
     bars.reverse();
     colors.reverse();
     return (
-        <svg viewBox="0 0 400 400">
```
wkoot, line 36: Wasn't this needed for tags under the cards, when the screen space is very narrow?

components/frontend/src/dashboard/StatusBarChart.js
```diff
@@ -8,7 +8,9 @@ function nr_metrics_label(nr_metrics) {
 
 export function StatusBarChart({ animate, colors, label, tooltip, summary, maxY, style }) {
     const nrMetrics = sum(summary[Object.keys(summary)[0]])
-    const barRatio = Math.max(5, nrMetrics) / maxY
+    const nrDates = Object.keys(summary).length
+    const horizontalPadding = 50 + (7 - nrDates) * 10
```
wkoot: Perhaps a comment to describe the calculation would help, seems like magic numbers to me

components/frontend/src/dashboard/StatusBarChart.js
fniessink: Unfortanately not. It looked great on my Retina display, but not a wide screen. Will fix this.

components/frontend/src/dashboard/StatusBarChart.js
fniessink: Made the graphs responsive, so no longer needed.

components/frontend/src/dashboard/StatusBarChart.js
fniessink: Code changed, but explanation added for the new calculation.

PR #5894, opened 2023-04-13
Allow for filtering out specific Axe measurement entities through regular expressions

### LINKED ISSUE #3328: Allow for filtering out specific measurement entities through regular expressions
This feature already exists for OWASP sources, I believe, but it would also be useful for other sources like AXE core.

For example, AXE core has limitations and therefore generates false positives. 
Some tools cannot be configured properly to remove these violations. In those cases, it would be useful to manage this in QT. That is the added value of QT.

Add two fields:
- Entities to include (regex)
- Entities to exclude (regex)

Add a possibility to the data model to exclude entity attributes from matching.

Implement a generic matcher in the base collector.

Note: Order of including/excluding doesn't matter, explain this in the UI somewhere.

We first need to refactor the collector to allow for filtering entities by regular expression in one location (in the base SourceCollector): #4714 and #5240.

This is related to https://github.com/ICTU/quality-time/issues/4615

wkoot:
Closes #3328

### DIFF
--- a/components/shared_data_model/src/shared_data_model/sources/axe.py
+++ b/components/shared_data_model/src/shared_data_model/sources/axe.py
@@ -43,6 +43,7 @@ ELEMENT_EXCLUDE_FILTER = MultipleChoiceWithAdditionParameter(
     name="Elements to exclude (regular expressions)",
     short_name="element exclude filter",
     help="Elements to exclude can be specified by regular expression.",
+    placeholder="none",
     metrics=["accessibility"],
 )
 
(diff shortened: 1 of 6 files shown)


components/shared_data_model/src/shared_data_model/sources/axe.py
```diff
@@ -31,6 +31,21 @@
...
+    help="Elements to include can be specified by regular expression.",
+    placeholder="all",
+    metrics=["accessibility"],
+)
+
+ELEMENT_EXCLUDE_FILTER = MultipleChoiceWithAdditionParameter(
+    name="Elements to exclude (regular expressions)",
+    short_name="element exclude filter",
+    help="Elements to exclude can be specified by regular expression.",
+    metrics=["accessibility"],
```
fniessink:
```suggestion
    help="Elements to exclude can be specified by regular expression.",
    placeholder="none",
    metrics=["accessibility"],
```

fniessink (approved):

PR #6114, opened 2023-05-10
When using SonarQube security hotspots as source for the security war…

### LINKED ISSUE #5956: Show hotspots that are acknowledged in Quality-time
If a developer acknowledges a hotspot in SonarQube it disappears from Quality-time as security warning, but the hotspot may not have been addressed yet.

Show hotspots in Quality-time with status "to review" as well as "acknowledged".

CC: @denniebouman

fniessink:
…nings metric, allow for filtering hotspots by hotspot status ('to review', 'acknowledged', 'fixed', 'safe'). Closes #5956.

### DIFF
--- a/components/collector/src/source_collectors/sonarqube/security_warnings.py
+++ b/components/collector/src/source_collectors/sonarqube/security_warnings.py
@@ -45,17 +45,12 @@ class SonarQubeSecurityWarnings(SonarQubeViolations):
                 )
             )
         if "security_hotspot" in security_types:
-            # Note: SonarQube is a bit inconsistent. For issue search, the SonarQube status parameter is called
-            # "statuses", but for hotspots it's called "status".
-            api_urls.append(
-                URL(f"{base_url}/api/hotspots/search?projectKey={component}&branch={branch}&status=TO_REVIEW&ps=500")
-            )
+            api_urls.append(URL(f"{base_url}/api/hotspots/search?projectKey={component}&branch={branch}&ps=500"))
         return await super()._get_source_responses(*api_urls, **kwargs)
 
     async def _parse_source_responses(self, responses: SourceResponses) -> SourceMeasurement:
         """Override to parse the selected security types."""
         security_types = self._parameter(self.types_parameter)
-        review_priorities = [priority.upper() for priority in self._parameter("review_priorities")]
         vulnerabilities = (
             await super()._parse_source_responses(SourceResponses(responses=[responses[0]]))
             if "vulnerability" in security_types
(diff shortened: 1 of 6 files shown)


components/shared_data_model/src/shared_data_model/sources/sonarqube.py
```diff
@@ -280,10 +292,10 @@ def violation_entity_attributes(include_review_priority=False, include_resolutio
             security_types=MultipleChoiceParameter(
                 name="Security issue types (measuring security hotspots requires SonarQube 8.2 or newer)",
                 short_name="types",
-                placeholder="all security issue types",
+                placeholder="vulnerability",
```
wkoot, line 295: Hehe, extra bugfix

components/collector/src/source_collectors/sonarqube/security_warnings.py
```diff
@@ -45,17 +45,12 @@ async def _get_source_responses(self, *urls: URL, **kwargs) -> SourceResponses:
                 )
             )
         if "security_hotspot" in security_types:
-            # Note: SonarQube is a bit inconsistent. For issue search, the SonarQube status parameter is called
-            # "statuses", but for hotspots it's called "status".
```
wkoot, line 49: Shouldn't we leave this comment in somewhere, so we don't mistakenly fix it again? Or is this new API a drop-in replacement?

components/collector/src/source_collectors/sonarqube/security_warnings.py
```diff
@@ -75,8 +70,16 @@ async def _parse_source_responses(self, responses: SourceResponses) -> SourceMea
             entities=vulnerabilities.entities + hotspots,
         )
 
-    async def __entity(self, hotspot) -> Entity:
-        """Create the security warning entity."""
+    def __include_hotspot(self, hotspot) -> bool:
+        """Return whether to include the hotspot."""
+        review_priorities = self._parameter("review_priorities")
+        review_priority = hotspot["vulnerabilityProbability"].lower()
```
wkoot, line 76: I had to read twice, but to be sure: we used to check on param.upper and now we check on value.lower so it's semantically the same. But that assumes that users have always typed the parameters in lowercase, no?

components/shared_data_model/src/shared_data_model/sources/sonarqube.py
fniessink: Yeah, it didn't seem important enough to fix separately.

components/collector/src/source_collectors/sonarqube/security_warnings.py
fniessink: It's tempting, but since we don't use the status parameter of the hotspot search endpoint anymore it feels a bit weird to keep a comment about it in the code.

components/collector/src/source_collectors/sonarqube/security_warnings.py
fniessink: No, the parameter is multiple choice, so there's no typing involved.

PR #6172, opened 2023-05-22
[notifier] Add ruff to quality tools.

fniessink:


components/notifier/src/strategies/notification_strategy.py
```diff
@@ -2,16 +2,25 @@
...
-    def get_notifications(self, reports, measurements, most_recent_measurement_seen: datetime) -> list[Notification]:
+    def get_notifications(
+        self,
+        reports: list[Report],
+        measurements: list[Measurement],
+        most_recent_measurement_seen: datetime,
+    ) -> list[Notification]:
         """Return the reports that have a webhook and metrics that require notifying."""
-        measurements.sort(key=lambda measurement: measurement["end"])
+        measurements.sort(key=lambda measurement: str(measurement["end"]))
```
wkoot: Hmm, a little dirty but if measurement end is always in `datetime.isoformat()` and we always use UTC (from now on), then I guess this is sorted as expected. Perhaps add an inline comment to point that out

components/notifier/tests/notifier/test_notifier.py
```diff
@@ -58,84 +64,89 @@ def setUp(self):
...
     @patch("logging.error")
     @patch("asyncio.sleep")
     @patch("notifier.notifier.get_reports_and_measurements")
-    async def test_exception(self, mocked_get, mocked_sleep, mocked_log_error):
+    async def test_exception(self, mocked_get: Mock, mocked_sleep: Mock, mocked_log_error: Mock):
         """Test that an exception while retrieving the reports is handled."""
         mocked_get.side_effect = OSError
         mocked_sleep.side_effect = RuntimeError
-        try:
+        with contextlib.suppress(RuntimeError):
```
wkoot, line 90: Nice one to keep in mind!

components/notifier/tests/notifier_utilities/test_data.py
```diff
@@ -26,37 +26,52 @@ class NotifyUtilitiesTests(unittest.TestCase):
...
-            dict(metric_uuid=METRIC_ID2, end=self.now, start=self.now, count=dict(status="target_not_met", value="0")),
-            dict(
-                metric_uuid=METRIC_ID2,
-                end=self.just_now,
-                start=self.just_now,
-                count=dict(status="target_met", value="10"),
-            ),
+            {
+                "metric_uuid": METRIC_ID,
+                "end": self.now,
```
wkoot, line 36: Is there a special reason for end,start instead of start,end?

components/notifier/tests/strategies/test_notification_strategy.py
```diff
@@ -1,11 +1,14 @@
 """"Unit tests for the notification strategies."""
 
 import unittest
-from datetime import datetime, timezone
+from datetime import datetime, UTC
 
 from strategies.notification_strategy import NotificationFinder
 
 
+Measurements = dict[str, str | dict[str, str]]
```
wkoot: Why like this instead of from models?

components/notifier/tests/notifier_utilities/test_data.py
fniessink: No, no special reason. I checked whether Black can help with this, but apparently not (TLDR; it would cause side-effects in some situations). See https://github.com/psf/black/issues/523.

components/notifier/tests/strategies/test_notification_strategy.py
fniessink: Good catch. Fixed.

components/notifier/src/strategies/notification_strategy.py
fniessink: Done.

PR #6234, opened 2023-05-26
Don't attempt to collect all projects and repositories from Harbor wh…

### LINKED ISSUE #6220: Harbor source forces collector to restart
**Describe the bug**
Harbor source forces collector to restart. 
Filtering Projects or Repositories with Harbor as source occurs when all the results (artifacts) from Harbor are collected. When there are many results from/in Harbor, this takes longer than the set timeout of the health check, which forces a restart of the collector container.
To fix this bug, filter before the results are collected.

**To Reproduce**
1. Configure Security Warnings Metric
2. Add Harbor as source and configure (there should be many artifacts in Harbor)
3. Collector will be restarted after healthcheck timeout (10 minutes?)
4. Measurement is absent (no recent measurement)

**Expected behaviour**
Vulnerabilities will be shown in QT (if present) and collector will not be restarted

 - Quality-time version: v4.10.0

fniessink:
…en the user has configured filters on projects and/or repositories. Throw a timeout error when collecting measurement data from a source takes longer than the configured time between measurement attempts (this can be changed via the  environment variable which has a default value of 20 seconds). Fixes #6220.


components/collector/src/base_collectors/config.py
```diff
@@ -0,0 +1,8 @@
+"""Configuration."""
+
+import os
+
+
+MAX_SLEEP_DURATION = int(os.environ.get("COLLECTOR_SLEEP_DURATION", 20))
+MEASUREMENT_LIMIT = int(os.environ.get("COLLECTOR_MEASUREMENT_LIMIT", 30))
+MEASUREMENT_FREQUENCY = int(os.environ.get("COLLECTOR_MEASUREMENT_FREQUENCY", 15 * 60))
```
wkoot: But wait, there's more! https://quality-time.readthedocs.io/en/latest/software.html#id7

components/collector/src/source_collectors/harbor/security_warnings.py
```diff
@@ -98,24 +122,6 @@ async def _create_entity(self, artifact: dict) -> Entity:
             fixable=scan_summary.get("fixable", 0),
         )
 
-    def _include_entity(self, entity: Entity) -> bool:
```
wkoot, line 111: We built this in #4714 and #5240. I think it makes sense to revise the `collect>parse>include` flow to something like `collect>filter>parse>include`, so we have a generic way to implement such pre-work filtering. Some passthrough method in `base_collectors/source_collector.py`?

components/collector/src/source_collectors/harbor/security_warnings.py
```diff
@@ -48,23 +48,47 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
...
+        return False
+
     async def _get_source_responses_for_project(self, project, repositories_api: URL) -> SourceResponses:
         """Return the source responses for a specific project."""
         responses = SourceResponses()
         repositories_responses = await super()._get_source_responses(repositories_api)
         for repositories_response in repositories_responses:
             for repository in await repositories_response.json():
                 repository_name = repository["name"].removeprefix(f"{project['name']}/")
+                if self._skip_repository(repository_name):
```
wkoot, line 84: The next line does some string munging, which is currently used to do the filtering.

components/collector/src/source_collectors/harbor/security_warnings.py
fniessink: Yes, but the things to filter on are different for each source because we're dealing with the raw data at that point in the flow. So I'm not sure how to solve that. And won't in this PR.

components/collector/src/source_collectors/harbor/security_warnings.py
fniessink: The next line quotes the repository name so it works when used in the URL for the artifacts endpoint. The filtering is done on the unquoted repository name.

PR #6252, opened 2023-05-31
Changes _www_1 tot -www-1.

ptrr:


fniessink:
This would break the application tests.

PR #6259, opened 2023-05-31
Use Python or JS script for health check instead of curl.

### LINKED ISSUE #6253: Replace curl for health checks with a small python script
Curl under Alpine has updates very frequently, breaking the build at inconvenient times.

fniessink:
Components:
- [x] external server
- [x] internal server
- [x] renderer
- [x] frontend

Closes #6253.

### DIFF
--- a/components/external_server/src/healthcheck.py
+++ b/components/external_server/src/healthcheck.py
@@ -1,0 +1,15 @@
+"""Health check."""
+
+from http import HTTPStatus
+from os import getenv
+from urllib.request import urlopen
+import sys
+
+external_server_port = getenv("EXTERNAL_SERVER_PORT", "5001")
+# pylint: disable=invalid-name
+exit_code = 1
+try:
+    with urlopen(f"http://localhost:{external_server_port}/api/health") as response:  # nosec B310
+        exit_code = 0 if response.status == HTTPStatus.OK else 1
+finally:
+    sys.exit(exit_code)
--- a/components/frontend/healthcheck.js
+++ b/components/frontend/healthcheck.js
@@ -1,0 +1,18 @@
+const http = require('http');
+
+const options = {
+    host: 'localhost',
+    method: 'GET',
+    port: process.env.FRONTEND_PORT || 5000,
+    path: '/favicon.ico',
+};
+
+const healthCheck = http.request(options, (response) => {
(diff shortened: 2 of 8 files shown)


components/frontend/healthcheck.js
```diff
@@ -0,0 +1,18 @@
+const http = require('http');
+
+const options = {
+    host: 'localhost',
+    method: 'GET',
+    port: process.env.FRONTEND_PORT || 5000,
+    path: '/favicon.ico',
```
wkoot, line 7: Is this the most trustworthy url?

components/external_server/src/healthcheck.py
```diff
@@ -0,0 +1,15 @@
...
+import sys
+
+external_server_port = getenv("EXTERNAL_SERVER_PORT", "5001")
+# pylint: disable=invalid-name
+exit_code = 1
+try:
+    with urlopen(f"http://localhost:{external_server_port}/api/health") as response:  # nosec B310
+        exit_code = 0 if response.status == HTTPStatus.OK else 1
+finally:
+    sys.exit(exit_code)
```
wkoot, line 15: I was thinking of a nice way to use `raise SystemExit` instead, but it's a bit tedious with the status NOK check

components/internal_server/src/healthcheck.py
```diff
@@ -0,0 +1,15 @@
+"""Health check."""
```
wkoot, line 1: I guess this code duplication is fine for now, as it will disappear.

wkoot (commented):
Missing coverage is fine, as we implicitly test it in functional tests. A broken healthcheck will crash the containers :) Too bad we got code duplication but only one of the js files will remain

components/frontend/healthcheck.js
fniessink: Yeah, why not? It's small and always there.

components/external_server/src/healthcheck.py
fniessink: I like this solution because it prevents a bare exception check that linters would complain about.

components/internal_server/src/healthcheck.py
fniessink: Agreed.

Issue #6193, opened 2023-05-23
Update SonarQube rules


fniessink:
components/shared_data_model/src/shared_data_model/sources/sonarqube.py
metrics="suppressed_violations"
Add: python:NoSonar

components/shared_data_model/src/shared_data_model/sources/sonarqube.py
metrics="long_units"
Add: python:S138

components/shared_data_model/src/shared_data_model/sources/sonarqube.py
metrics="many_parameters"
Add: kotlin:S107

PR #6272, opened 2023-06-02
6193 update sonarqube rules

denniebouman:
Added SonarQube rules for suppressed violations, long units and many parameters. Fixes: (#6193)


fniessink (changes requested):
Can you add an entry to the changelog in docs/src/changelog.md? Thanks

fniessink (approved):

PR #6324, opened 2023-06-11
6323 move data model to shared code component

### LINKED ISSUE #6323: Move data model to shared code component

fniessink:
Note: this branch is on top of 6321-renamed-shared_server_code-to-shared_code.


docs/src/development.md
```diff
@@ -125,14 +125,14 @@ ci/pip-install.sh
 python src/quality_time_notifier.py
 ```
 
-#### Preparing the shared components
+#### Preparing the shared component
 
-*Quality-time* has two components that only contain shared code. The shared data model is used by all Python components. The code in the shared server code component is used by the external server.
+*Quality-time* has one component that only contain shared code. The shared code is used by all Python components.
```
wkoot, line 130: I would write, "which only contains"

wkoot: Isn't the data model also used by the frontend?

docs/src/development.md
```diff
@@ -395,7 +395,7 @@ This section assumes you have created a Python virtual environment, activated it
 To run the unit tests and measure unit test coverage of the backend components (this assumes you have created a Python virtual environment, activated it, and installed the requirements as described [above](#developing)):
 
 ```console
-cd components/external_server  # or components/shared_data_model, components/shared_server_code, components/collector, components/notifier
+cd components/external_server  # or components/collector, components/notifier, components/shared_code
```
wkoot, line 398: Merge conflict?

components/shared_code/src/shared/routes/plugins/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Bottle route plugins."""
+
+from .injection_plugin import InjectionPlugin
```
wkoot, line 3: Is this one of the files which was removed elsewhere? Seems weird to only have one route, do we happen to only have one plugin?

.circleci/config.yml
```diff
@@ -10,17 +10,16 @@ jobs:
...
           password: $DOCKERHUB_PASSWORD
-    parallelism: 6
+    parallelism: 5
     steps:
       - checkout
       - run: |
           case $CIRCLE_NODE_INDEX in
             0) component=components/collector;;
             1) component=components/notifier;;
             2) component=components/external_server;;
```
wkoot, line 20: I don't recall seeing this line changed in the other PR, was it another branch?

wkoot (approved):

.circleci/config.yml
ptrr: There is one less component. I don't know why this change is here and not in the remove internal server branch.

docs/src/development.md
fniessink: Which is used when the clause can be left out, which (pun intended) is not the case above.

Yes, the frontend uses the data model, but by downloading it from the API-server, not by using the Python code directly.

docs/src/development.md
fniessink: Fixed.

components/shared_code/src/shared/routes/plugins/__init__.py
fniessink: The injection plugin is in shared because it was also used by the internal server. It can be moved back to the API-server. But not now, in a next PR.

.circleci/config.yml
fniessink: I think I forgot to change it one time and fixed it later on.

fniessink:
Closed in favour of 6323-move-data-model-to-shared-code-component-2

PR #6344, opened 2023-06-13
Move shared code that is only used by one component to that component

### LINKED ISSUE #6343: Move shared code only used by the api server to the api server

fniessink:
Closes #6343.


components/shared_code/src/shared/database/reports.py
```diff
@@ -1,78 +1,12 @@
...
-        overview["_id"] = str(overview["_id"])
-    return overview or {}
+from shared.model.report import Report
+from shared_data_model import DATA_MODEL
 
 
-def report_exists(database: Database, report_uuid: ReportId) -> bool:
-    """Return whether a report with the specified report uuid exists."""
-    return report_uuid in database.reports.distinct("report_uuid")
+def get_reports(database: Database) -> list[Report]:
```
wkoot: This looks like a simplified implementation of the existing `api_server.src.database.latest_reports`, which is also changed in this PR. I'm on the fence regarding partial implementation, placing everything in shared will bloat the module but it also avoids bits of similar logic sprinkled around components

PR #6404, opened 2023-06-21
Improve deployment instructions.

fniessink:


docs/src/deployment.md
```diff
@@ -10,6 +10,15 @@ This document describes how to deploy, and if needed move, the *Quality-time* ap
 
 This document assumes docker-compose is used to deploy the containers. The [docker folder](https://github.com/ICTU/quality-time/tree/master/docker) of the *Quality-time* repository contains different compose files for running *Quality-time* in development and continuous integration mode. You can use these compose files as basis for your own deployment configuration.
 
+To deploy *Quality-time* locally, follow these steps:
+
+1. Make a directory, e.g. `quality_time` and change directory to it.
+2. Download the docker compose files from the docker folder](https://github.com/ICTU/quality-time/tree/master/docker) and place them in the folder created in the previous step.
+3. Set the version number: `export QUALITY_TIME_VERSION=v4.10.0`.
```
wkoot: New version 5x

docs/src/deployment.md
```diff
@@ -10,6 +10,15 @@ This document describes how to deploy, and if needed move, the *Quality-time* ap
...
+4. Run `docker compose up`.
+5. Go to http://localhost to access the application.
```
wkoot, line 19: Technically, this works as the composefile specifies port 80.. But in most cases, a local deploy will not be on a privileged port. Should we address this, and explain how to use the override.yml?

fniessink: Added. The proxy port is an environment variable so `export PROXY_PORT=1080` is sufficient.

PR #6486, opened 2023-07-03
Show warning when user session expires. Show warning if session has e…

### LINKED ISSUE #5327: Show the user as logged out in the frontend as soon as the session expires
When the user session expires, the frontend does not automically show the user as logged out.

As soon as the session expires, log out the user from the frontend.

In App.js, onComponentMount, set an event handler that fires when the session expires and invoked `set_user(null)`

CC: @Sebastiaan127001

fniessink:
…xpired while user was away. Closes #5327.

### DIFF
--- a/components/frontend/src/App.js
+++ b/components/frontend/src/App.js
@@ -39,16 +39,9 @@ class App extends Component {
             reportDate.setHours(23, 59, 59);
         }
         this.login_forwardauth();
+        this.initUserSession()
         this.connect_to_nr_measurements_event_source();
-        if (new Date(localStorage.getItem("session_expiration_datetime")) < new Date()) {
-            this.set_user(null)  // The session expired while the user was away
-        }
-        this.setState(
-            {
-                report_uuid: report_uuid, report_date: reportDate, loading: true,
-                user: localStorage.getItem("user"), email: localStorage.getItem("email")
-            },
-            () => this.reload());
+        this.setState({ report_uuid: report_uuid, report_date: reportDate, loading: true }, () => this.reload());
     }
 
     componentWillUnmount() {
@@ -90,7 +83,7 @@ class App extends Component {
 
     check_session(json) {
         if (json.ok === false && json.status === 401) {
-            this.set_user(null);
+            this.setUserSession();
             if (this.login_forwardauth() === false) {
                 show_message("warning", "Your session expired", "Please log in to renew your session");
             }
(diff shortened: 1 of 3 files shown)


components/frontend/src/App.js
```diff
@@ -39,16 +39,9 @@ class App extends Component {
...
-        if (new Date(localStorage.getItem("session_expiration_datetime")) < new Date()) {
-            this.set_user(null)  // The session expired while the user was away
-        }
-        this.setState(
-            {
-                report_uuid: report_uuid, report_date: reportDate, loading: true,
-                user: localStorage.getItem("user"), email: localStorage.getItem("email")
-            },
-            () => this.reload());
+        this.setState({ report_uuid: report_uuid, report_date: reportDate, loading: true }, () => this.reload());
```
wkoot, line 44: Nice simplification from before. I guess converting `report_uuid` to camelCase is annoying :o)

components/frontend/src/App.js
```diff
@@ -152,24 +145,47 @@ class App extends Component {
...
+        }
+    }
+
+    setUserSession(username, email, sessionExpirationDateTime) {
+        if (username) {
+            const emailAddress = email && email.indexOf("@") > -1 ? email : null;
+            this.setState({ user: username, email: emailAddress });
+            localStorage.setItem("user", username);
+            localStorage.setItem("email", emailAddress);
+            localStorage.setItem("session_expiration_datetime", sessionExpirationDateTime.toISOString());
```
wkoot, line 177: Why is `session_expiration_datetime` still snake case here, is that something expected by localStorage?

components/frontend/src/App.js
```diff
@@ -152,24 +145,47 @@ class App extends Component {
...
+            this.sessionExpirationTimeoutId = setTimeout(() => {
+                this.setUserSession(null);
+                show_message("warning", "Your session expired", "Please log in to renew your session");
+            }, sessionExpirationDateTime - new Date())
+        } else {
+            this.setState({ user: null, email: null });
```
wkoot, line 182: Why is `setState` sometimes followed by reload but sometimes not?

components/frontend/src/App.js
```diff
@@ -152,24 +145,47 @@ class App extends Component {
...
+            } else {
+                // Session is still active, restore it from local storage.
+                this.setUserSession(localStorage.getItem("user"), localStorage.getItem("email"), sessionExpirationDateTime)
+            }
+        }
+    }
+
+    setUserSession(username, email, sessionExpirationDateTime) {
+        if (username) {
+            const emailAddress = email && email.indexOf("@") > -1 ? email : null;
```
wkoot, line 173: We don't have a util method or package to check email validity?

components/frontend/src/App.js
```diff
@@ -152,24 +145,47 @@ class App extends Component {
...
+            this.setState({ user: username, email: emailAddress });
+            localStorage.setItem("user", username);
+            localStorage.setItem("email", emailAddress);
+            localStorage.setItem("session_expiration_datetime", sessionExpirationDateTime.toISOString());
+            this.sessionExpirationTimeoutId = setTimeout(() => {
+                this.setUserSession(null);
+                show_message("warning", "Your session expired", "Please log in to renew your session");
```
wkoot: The same happens in `check_session` and `initUserSession`, does it make sense to extract?

components/frontend/src/App.js
fniessink: Yeah, that's a too big a change to include here, it's used everywhere.

components/frontend/src/App.js
fniessink: I didn't check, but it would cause users to be logged out after an upgrade. Not a big issues, but not worth including in this PR also, I think.

components/frontend/src/App.js
fniessink: The difference is that after some state changes new data needs to be loaded and after other state changes this is not needed. For example, if the date is changed to last month, old data needs to be loaded. But if the user logs out, no data needs to be loaded.

components/frontend/src/App.js
fniessink: This doesn't check validity so much, but whether an email address is present. I can't really remember why this needed. Maybe LDAP doesn't always return an email address? Anyway, outside the scope of this PR ;-)

components/frontend/src/App.js
fniessink: Yup.

PR #6538, opened 2023-07-07
In the login dialog, don't tell the user credentials are invalid when…

### LINKED ISSUE #4704: When a user tries to log in without a (VPN) connection, QT reports invalid credentials.
**Describe the bug**
When a user tries to log in without a (VPN) connection, QT reports invalid credentials.

**To Reproduce**
1. Make sure there is a connection.
2. Navigate to Quality-time 
3. Break the (VPN) connection.
4. Try to log in with correct credentials.

**Expected behaviour**

**Screenshots**
image

**Additional context**
NB: fixing https://github.com/ICTU/quality-time/issues/3580 may interfere by this issue by showing a toast that the server can't be reached. Make sure the user either gets an error message in the modal dialog or via the toast, but not both.

fniessink:
… in fact the server couldn't be reached. Fixes #4704.

### DIFF
--- a/components/frontend/src/header_footer/Menubar.js
+++ b/components/frontend/src/header_footer/Menubar.js
@@ -10,7 +10,7 @@ import './Menubar.css';
 function Login({ set_user }) {
     const [username, setUsername] = useState('');
     const [password, setPassword] = useState('');
-    const [login_error, setLoginError] = useState(false);
+    const [error, setError] = useState('');
 
     function submit() {
         login(username, password)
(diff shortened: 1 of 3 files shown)


components/frontend/src/header_footer/Menubar.js
```diff
@@ -10,31 +10,40 @@ import './Menubar.css';
...
+        <Modal trigger={<Button secondary><Icon name='user' />Login</Button>} size='tiny' onClose={() => setError("")} >
             <Modal.Header content='Login' />
             <Modal.Content>
-                <Form error={login_error} warning={!login_error} onSubmit={() => submit()} >
+                <Form error={!!error} warning={!error} onSubmit={() => submit()} >
                     <Form.Input autoFocus id='username' name='username' label='Username' onChange={(_event, { value }) => setUsername(value)} />
                     <Form.Input id='password' name='password' type='password' label='Password' onChange={(_event, { value }) => setPassword(value)} />
-                    <Message error header='Invalid credentials' content='Username and/or password are invalid. Please try again.' />
-                    <Message warning header='Heads up' content='Changes you make after you log in, such as adding metrics, changing metric targets, and marking issues as false positive, are logged.' />
+                    <Message error={!!error} warning={!error} header={messageHeader} content={messageContent} />
```
wkoot, line 46: What is the meaning of "warning", is it just the colour of the message box?

components/frontend/src/header_footer/Menubar.js
```diff
@@ -10,31 +10,40 @@ import './Menubar.css';
...
+        messageHeader = "Invalid credentials"
+        messageContent = "Username and/or password are invalid. Please try again."
+    }
     return (
-        <Modal trigger={<Button secondary><Icon name='user' />Login</Button>} size='tiny' onClose={() => setLoginError(false)} >
+        <Modal trigger={<Button secondary><Icon name='user' />Login</Button>} size='tiny' onClose={() => setError("")} >
             <Modal.Header content='Login' />
             <Modal.Content>
-                <Form error={login_error} warning={!login_error} onSubmit={() => submit()} >
+                <Form error={!!error} warning={!error} onSubmit={() => submit()} >
```
wkoot, line 43: Why `!!error` and `!error`? Shouldn't they both set the same amount of bangs?

components/frontend/src/header_footer/Menubar.js
fniessink: Yes, it's the style of the message box. See https://react.semantic-ui.com/collections/message/ -> props.

components/frontend/src/header_footer/Menubar.js
fniessink: The error flag must be a boolean. `!!x` makes sure the result of the expression is a boolean no matter what the type of `x` is. See https://brianflove.com/2014-09-02/whats-the-double-exclamation-mark-for-in-javascript/

wkoot: Yeah I did know that, but I was surprised that the behaviour between error and warning was different :) Good to know then, I guess

PR #6541, opened 2023-07-09
Notify the user why editing is not possible

### LINKED ISSUE #3380: Notify the user why editing is not possible
It would be useful to notify the user when the user tries to edit a field when it's not editable.
This can occur in the following situations:

- the user is not logged in
- the user is using the 'time machine' functionality where he/she looks into a view of the past, fields are not editable on purpose

fniessink:
Fixes #3380.

### DIFF
--- a/components/api_server/src/routes/report.py
+++ b/components/api_server/src/routes/report.py
@@ -114,7 +114,7 @@ def export_report_as_pdf(report_uuid: ReportId):
     renderer_host = os.environ.get("RENDERER_HOST", "renderer")
     renderer_port = os.environ.get("RENDERER_PORT", "9000")
     render_url = f"http://{renderer_host}:{renderer_port}/api/render"
-    query_string = f"?{bottle.request.query_string}" if bottle.request.query_string else ""
+    query_string = "?hide_toasts=true" + f"&{bottle.request.query_string}" if bottle.request.query_string else ""
     report_path = parse.quote(f"{report_uuid}{query_string}")
     response = requests.get(f"{render_url}?path={report_path}", timeout=120)
     response.raise_for_status()
--- a/components/api_server/src/routes/report.py
+++ b/components/api_server/src/routes/report.py
@@ -114,6 +114,7 @@ def export_report_as_pdf(report_uuid: ReportId):
     renderer_host = os.environ.get("RENDERER_HOST", "renderer")
     renderer_port = os.environ.get("RENDERER_PORT", "9000")
     render_url = f"http://{renderer_host}:{renderer_port}/api/render"
+    # Tell the frontend to not display toast messages to prevent them from being included in the PDF:
     query_string = "?hide_toasts=true" + (f"&{bottle.request.query_string}" if bottle.request.query_string else "")
     report_path = parse.quote(f"{report_uuid}{query_string}")
     response = requests.get(f"{render_url}?path={report_path}", timeout=120)
--- a/components/api_server/src/routes/report.py
+++ b/components/api_server/src/routes/report.py
@@ -114,7 +114,7 @@ def export_report_as_pdf(report_uuid: ReportId):
     renderer_host = os.environ.get("RENDERER_HOST", "renderer")
     renderer_port = os.environ.get("RENDERER_PORT", "9000")
     render_url = f"http://{renderer_host}:{renderer_port}/api/render"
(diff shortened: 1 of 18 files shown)


components/api_server/src/routes/report.py
```diff
@@ -114,7 +114,7 @@ def export_report_as_pdf(report_uuid: ReportId):
     renderer_host = os.environ.get("RENDERER_HOST", "renderer")
     renderer_port = os.environ.get("RENDERER_PORT", "9000")
     render_url = f"http://{renderer_host}:{renderer_port}/api/render"
-    query_string = f"?{bottle.request.query_string}" if bottle.request.query_string else ""
+    query_string = "?hide_toasts=true" + (f"&{bottle.request.query_string}" if bottle.request.query_string else "")
```
wkoot, line 118: From code perspective, I don't really understand why there is `hide_toasts`. Perhaps something for a comment or otherwise in the developer documentation?

tests/application_tests/src/test_report.py
```diff
@@ -47,7 +47,7 @@ def setUp(self):
         self.driver = webdriver.Remote("http://selenium:4444", options=firefox_options)
         self.driver.implicitly_wait(10)
         self.wait = WebDriverWait(self.driver, 10)
-        self.driver.get("http://www:80")
+        self.driver.get("http://www:80?hide_toasts=true")
```
wkoot: Do we also test case false?

components/frontend/src/App.js
```diff
@@ -165,6 +175,12 @@ class App extends Component {
                 // Session is still active, restore it from local storage.
                 this.setUserSession(localStorage.getItem("user"), localStorage.getItem("email"), sessionExpirationDateTime)
             }
+        } else {
+            showMessage(
+                "info",
+                "Not logged in",
+                "You are not logged in. Editing is not possible until you do."
```
wkoot: I would say `until you do so`, or `until you are`.

components/frontend/src/utils.test.js
```diff
@@ -191,45 +199,43 @@ it('clears the array value', () => {
...
-    const { result } = renderHook(() => useURLSearchQuery(history, "key", "string", "default"))
+    history.push("?key=value")
+    const { result } = renderHook(() => useURLSearchQuery("key", "string", "default"))
     act(() => { result.current[1]("default") })
     expect(history.location.search).toEqual("")
 })
 
 it('returns registered URL search parameters only', () => {
-    const history = createMemoryHistory({ initialEntries: ['?unregistered_key=value&report_date=2022-02-11'] })
+    //const history = createMemoryHistory({ initialEntries: ['?unregistered_key=value&report_date=2022-02-11'] })
```
wkoot: Can be removed?

components/frontend/src/widgets/toast.js
```diff
@@ -1,8 +1,12 @@
 import { toast } from 'react-toastify';
+import { registeredURLSearchParams } from '../utils'
 
 export function showMessage(type, title, description) {
-    const toast_message = title && description ? <><h4>{title}</h4><p>{description}</p></> : title;
-    toast(toast_message, { type: type, autoClose: 20000 });
+    const hideToasts = registeredURLSearchParams().get("hide_toasts");
```
wkoot, line 5: Is it not a bit weird to infer the frontend logic from the API call it made itself, or am I reading this incorrectly?

components/api_server/src/routes/report.py
fniessink: Added a comment:

```suggestion
    # Tell the frontend to not display toast messages to prevent them from being included in the PDF:
    query_string = "?hide_toasts=true" + (f"&{bottle.request.query_string}" if bottle.request.query_string else "")
```

tests/application_tests/src/test_report.py
fniessink: Added a comment to explain:

```suggestion
        self.driver.get("http://www:80?hide_toasts=true")  # Hide toasts so they don't block other visual elements
```

components/frontend/src/widgets/toast.js
fniessink: That's because there are two frontends running: one in the user's browser and one in the headless browser of the renderer. When the user clicks "export report to PDF" the renderer opens the report, that the user has open in their browser, in a headless browser. The report when opened in the headless browser should not display toasts because we do not want them to end up in the PDF.

PR #6558, opened 2023-07-10
Add a note to the reference manual about how to configure Jenkins pri…

### LINKED ISSUE #6557: Add note about Jenkins private token to docs
When a user uses Jenkins to get a file source (e.g. JUnit XML) from it and the user generates a private token, Jenkins needs Quality-time to pass the private token as password (and it needs the user name as well), which is a bit surprising.

See https://www.jenkins.io/doc/book/system-administration/authenticating-scripted-clients/

Add a note to the docs about this, comparable to the note about using GitLab as source for files.

fniessink:
…vate tokens in Quality-time. Closes #6557.

### DIFF
--- a/components/shared_code/src/shared_data_model/sources/jenkins.py
+++ b/components/shared_code/src/shared_data_model/sources/jenkins.py
@@ -12,6 +12,8 @@ from shared_data_model.parameters import (
     access_parameters,
 )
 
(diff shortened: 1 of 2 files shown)


Sebastiaan127001 (approved):
ziet er goed uit :)

PR #6571, opened 2023-07-11
When the frontend cannot reach the server, notify the user of the pro…

### LINKED ISSUE #3580: When VPN connection drops no error occurs
When accessing QT over a VPN and the connection is suddenly broken, QT still appears to be functioning, but it cannot update any field anymore without telling the user why.

fniessink:
…blem. When the server is reachable, notify the user as well. Fixes #3580.

### DIFF
--- a/components/frontend/src/App.js
+++ b/components/frontend/src/App.js
@@ -15,8 +15,17 @@ class App extends Component {
     constructor(props) {
         super(props);
         this.state = {
-            datamodel: {}, reports: [], report_uuid: '', report_date: null, reports_overview: {},
-            nr_measurements: 0, loading: true, user: null, email: null, last_update: new Date()
+            datamodel: {},
+            reports: [],
+            report_uuid: '',
+            report_date: null,
+            reports_overview: {},
+            nrMeasurements: 0,
+            nrMeasurementsStreamConnected: true,  // Assume initial connection will be successful
+            loading: true,
+            user: null,
+            email: null,
+            last_update: new Date()
         };
         history.listen(({ location, action }) => this.on_history({ location, action }));
     }
@@ -40,7 +49,7 @@ class App extends Component {
         }
         this.login_forwardauth();
         this.initUserSession()
-        this.connect_to_nr_measurements_event_source();
+        this.connectToNrMeasurementsEventSource();
         this.setState({ report_uuid: report_uuid, report_date: reportDate, loading: true }, () => this.reload());
     }
 
@@ -54,7 +63,7 @@ class App extends Component {
             this.changed_fields = json.availability ? json.availability.filter((url_key) => url_key.status_code !== 200) : null;
             this.check_session(json)
         }
-        const show_error = () => showMessage("error", "Server unreachable", "Couldn't load data from the server. Please try again later.");
+        const show_error = () => showMessage("error", "Server unreachable", "Couldn't load data from the server.");
         this.loadAndSetState(show_error)
     }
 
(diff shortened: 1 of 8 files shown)


components/frontend/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "quality-time-app",
   "version": "5.0.1",
   "private": true,
-  "proxy": "http://127.0.0.1:5001",
+  "proxy": "http://0.0.0.0:5001",
```
wkoot: This can only work if it's a listen address, but I assume it isn't?

components/frontend/src/App.js
```diff
@@ -134,20 +143,28 @@ class App extends Component {
...
+        });
+        this.source.addEventListener("delta", (message) => {
+            const newNrMeasurements = Number(message.data);
+            if (newNrMeasurements !== self.state.nrMeasurements) {
+                self.setState({ nrMeasurements: newNrMeasurements }, () => self.reload());
             }
-        }, false);
+        });
+        this.source.addEventListener("error", () => {
+            showMessage("error", "Server unreachable", "Trying to reconnect to server...")
```
wkoot, line 165: Do we not test this?

components/frontend/src/App.js
```diff
@@ -218,7 +235,7 @@ class App extends Component {
                 handleDateChange={(e, { name, value }) => this.handleDateChange(e, { name, value })}
                 last_update={this.state.last_update}
                 loading={this.state.loading}
-                nr_measurements={this.state.nr_measurements}
+                nr_measurements={this.state.nrMeasurements}
```
wkoot: Why keep `nr_measurements`, is it a large refactor?

components/frontend/src/App.js
```diff
@@ -134,20 +143,28 @@ class App extends Component {
...
-        this.source.addEventListener('delta', function (e) {
-            self.setState({ nr_measurements: Number(e.data) }, () => self.reload());
-        }, false);
-        this.source.addEventListener('error', function (e) {
-            if (e.readyState === EventSource.CLOSED || e.readyState === EventSource.OPEN) {
-                self.setState({ nr_measurements: 0 });
+        this.source.addEventListener("init", (message) => {
+            const newNrMeasurements = Number(message.data);
+            if (!self.state.nrMeasurementsStreamConnected) {
+                showMessage("success", "Connected to server", "Successfully reconnected to server.")
```
wkoot, line 152: Is there a risk of toast-spam when a connection is (very) unstable?

components/frontend/package.json
fniessink: I was getting error messages from node, saying it couldn't proxy requests, but I can't reproduce it now, so reverting this change.

components/frontend/src/App.js
fniessink: We weren't. I will see if we can mock the SSE.

components/frontend/src/App.js
fniessink: The refactoring is pretty small, but if the code touched doesn't have tests SonarQube is going to punish me for it :-)

Anyway, I've made the change.

components/frontend/src/App.js
fniessink: The timeout for the reconnection attempt is 20 seconds, which is about as long as the toasters are visible, so that matches.

components/frontend/src/App.js
fniessink: 100% coverage, yippee!

PR #6604, opened 2023-07-16
Use another third-party package for the data picker. Fixes #5446.

### LINKED ISSUE #5446: Entering dates by hand does not accept hyphens
when entering an end date for technical debt, the input field suggests (in a gray prefilled text) the mandatory format. But when trying to enter this, I cannot enter the hyphen character. Try this by typing:
"2"
"0"
"2"
"3
"-"
and see what happens.

image

In addition to this.
The calendar date source does not show the mandatory format which makes it not consistent with similar input fields.

image

fniessink:


components/frontend/package.json
```diff
@@ -8,14 +8,15 @@
     "fomantic-ui-css": "^2.9.2",
     "history": "^5.3.0",
     "react": "^18.2.0",
+    "react-datepicker": "^4.16.0",
```
wkoot, line 11: Does this also solve a lot of other gripes we've had with the datepicker?

components/frontend/src/fields/DateInput.css
```diff
@@ -1,3 +1,15 @@
-form > div.field {
-    width: 100%;  /* Unfortunately, the date picker does not support fluid, so use this as a work-around */
+.react-datepicker__input-container > input, .react-datepicker-wrapper {
+    width: 100% !important;  /* Unfortunately, the date picker does not support fluid, so use this as a work-around */
```
wkoot, line 2: Is the comment still valid, also for this new package?

components/frontend/src/widgets/DatePicker.css
```diff
@@ -0,0 +1,3 @@
+.react-datepicker__close-icon::after {
+    background-color: grey !important;
```
wkoot, line 2: Does this also work for dark mode?

components/frontend/src/fields/DateInput.css
fniessink: Yup!

components/frontend/src/widgets/DatePicker.css
fniessink: Yes, I checked :-)

components/frontend/package.json
fniessink: This one has proper keyboard navigation and seems to be actively maintained. Only disadvantage is that the calendar popup has no dark mode, but since it's a bit smaller than the previous one it is not too bad. If necessary we can always tinker with the CSS later on.

PR #6625, opened 2023-07-18
Show the date and time when a measurement entity was first seen by Q…

### LINKED ISSUE #6351: Show first seen timestamp of measurement entities
I'm always frustrated when the dev team asks me what had changed when looking at a metric that reports more issues that yesterday. For instance, the number of security warnings metric having OWASP dependency-check as a source. Regularly, the tool reports new CVE's and adds them to the list. Currently, Quality-time does not show, in the measurement entities tab, which warnings are new since today.

It would be useful if the entities table also had a column with a 'first reported date'.

**Tasks**

- [x] Keep track of the first seen timestamp in the metric data. 
- [x] Copy the timestamp when adding a new measurement for the same metric (like we copy the user data).
- [x] Add a column to the measurement entity table "First seen" (make sure it sorts correctly).
- [x] Implement this generically for all metrics with entities.
- [x] When there is no first seen date, make the column cell empty.

Out of scope:
- Make sure the measurement entities column is sorted by "First seen" (descending) by default. See #6644.

   
image

fniessink:
…uality-time. Closes #6351.

### DIFF
--- a/components/collector/src/model/entity.py
+++ b/components/collector/src/model/entity.py
@@ -3,6 +3,8 @@
 import re
 from collections.abc import Iterable
 
+from shared.utils.functions import iso_timestamp
+
 QUOTED_SLASH = re.compile("%2f", re.IGNORECASE)
 
 
@@ -10,7 +12,7 @@ class Entity(dict):
     """Class to hold information about an individual measurement entity, e.g. one violation, or one security warning."""
 
     def __init__(self, key: str, **attributes) -> None:
-        kwargs = {"key": self.safe_entity_key(key)}
+        kwargs = {"key": self.safe_entity_key(key), "first_seen": iso_timestamp()}
         kwargs.update(**attributes)
         super().__init__(**kwargs)
 
@@ -24,8 +26,7 @@ class Entities(list[Entity]):
     """Class to hold a list of unique entities."""
 
     def __init__(self, entities: Iterable[Entity] = ()) -> None:
-        """Extend to filter duplicate entities."""
-        super().__init__()
+        """Override to filter duplicate entities."""
         self.__keys: set[str] = set()  # Keep track of the keys in a set to make adding entities faster
         self.extend(entities)
 
--- a/components/collector/tests/source_collectors/source_collector_test_case.py
+++ b/components/collector/tests/source_collectors/source_collector_test_case.py
@@ -127,8 +127,9 @@ class SourceCollectorTestCase(unittest.IsolatedAsyncioTestCase):
         """Assert that the measurement source attribute has the expected value."""
         attribute_value = getattr(measurement.sources[source_index], attribute_key)
         if isinstance(expected_attribute_value, list):
-            for pair in zip(expected_attribute_value, attribute_value, strict=True):
-                self.assertDictEqual(pair[0], pair[1])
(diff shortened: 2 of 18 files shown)


components/shared_code/tests/shared/model/test_measurement.py
```diff
@@ -250,6 +250,37 @@ def test_entity_user_data(self):
...
+                },
+            ],
+        )
+        measurement_2 = Measurement(
+            self.metric(),
+            sources=[
+                {
+                    "source_uuid": SOURCE_ID,
+                    "type": "azure_devops",
+                    "entities": [{"key": "key1", "first_seen": "2023-07-16"}],
```
wkoot, line 271: Perhaps good to set this date to "2023-07-17", so it isn't the exact same as source 2

wkoot, line 271:
```suggestion
                    "entities": [{"key": "key1", "first_seen": "2023-07-17"}],
```

components/frontend/src/source/SourceEntities.test.js
```diff
@@ -105,6 +105,15 @@ it('sorts the entities by status rationale', async () => {
     assertOrder(["A", "B", "C"])
 })
 
+it('sorts the entities by first seen date', async () => {
+    render(<DataModel.Provider value={data_model}><SourceEntities metric={metric} source={source} /></DataModel.Provider>)
+    assertOrder(["C", "B", "A"])
+    await userEvent.click(screen.getByText(/first seen/))
```
wkoot, line 114: But from line 40 it seems that none of these entities have "first seen" attributes. If first seen would be the same as the datetime, I'd expect the order to be C-A-B.

components/collector/src/model/entity.py
```diff
@@ -3,14 +3,16 @@
...
+
 QUOTED_SLASH = re.compile("%2f", re.IGNORECASE)
 
 
 class Entity(dict):
     """Class to hold information about an individual measurement entity, e.g. one violation, or one security warning."""
 
     def __init__(self, key: str, **attributes) -> None:
-        kwargs = {"key": self.safe_entity_key(key)}
+        kwargs = {"key": self.safe_entity_key(key), "first_seen": iso_timestamp()}
```
wkoot, line 15: I'd expect a `datetime` object, but I guess this is fine since it's never interacted with in the backend (only copied if value is set)?

components/collector/tests/source_collectors/source_collector_test_case.py
```diff
@@ -127,8 +127,9 @@ def __assert_measurement_source_attribute(
         """Assert that the measurement source attribute has the expected value."""
         attribute_value = getattr(measurement.sources[source_index], attribute_key)
         if isinstance(expected_attribute_value, list):
-            for pair in zip(expected_attribute_value, attribute_value, strict=True):
-                self.assertDictEqual(pair[0], pair[1])
+            for expected_entity, actual_entity in zip(expected_attribute_value, attribute_value, strict=True):
+                expected_entity["first_seen"] = actual_entity["first_seen"] = "ignore first seen timestamp"
```
wkoot, line 131: hahahah, nice abuse of the fact it's just a dumb string :D

components/collector/src/model/entity.py
fniessink: Correct.

PR #6722, opened 2023-07-26
Add DevOps metric Change Failure Rate

### LINKED ISSUE #3499: Add DevOps metric Change Failure Rate
# User story

As a quality manager I want to measure change failure rate so that I can monitor the DevOps effectiveness of my team.

Change failure rate is defined as the percentage of deployments causing a failure in production.

This metric requires a "look back period" that the user can configure. 

Because this metric is a "rate", it needs a numerator (the number of failed deployments in the look back period) and a denominator (the number of deployments in the look back period). 

## Acceptance criteria

- [x] Users can add a change-failure-rate metric
- [x] Users can configure Azure DevOps as source for the metric (both for number of deployments and number of failures). AzureDevOps needs to be added as source only once.
- [x] #7374

## Assumptions

- The number of failures can be retrieved from Jira and Azure DevOps in the same way as issues are retrieved, i.e. using a JQL or WIQL query. This allows for ignoring e.g. minor bugs or including only bugs with a certain label.
- Deployments are automated and can be retrieved from a CI-tool. In other words, we don't support deployments that are manually tracked in issue trackers.
- Bugs have a timestamp. Deployments have a timestamp. Bugs are caused by the deployment with the most recent timestamp <= timestamp of the bug.
- The sources need a "look back period". We assume users won't shoot themselves in the foot and configure different lookback periods for different sources.

## Out of scope

- GitLab as source for issues, i.e. failed deployments in this case.
- Manual Number is integer only, so can only support this metric but adding it twice, once for the numerator and once for the denominator. That's too confusing so out of scope.
- [x] #7677

# Tasks

### Shared code

- [x] Add metric `change_failure_rate` to the data model, with Jira, GitLab CI, Jenkins, and Azure DevOps as supporting sources. Add a `documentation` attribute (comparable to the documentation of the test cases metric, explaining to users that two sources need to be added to this metric.

### Collector

- [x] Add ChangeFailureRate metric collector in `src\metric_collectors\change_failure_rate.py`
- [x] Add AzureDevOpsChangeFailureRate collector
- [x] Add JiraChangeFailureRate collector
- [x] Add JenkinsChangeFailureRate collector
- [x] Add GitLabChangeFailureRate collector

## Documentation

- [ ] Update changelog

wkoot:
Closes #3499

### DIFF
--- a/components/collector/src/source_collectors/azure_devops/change_failure_rate.py
+++ b/components/collector/src/source_collectors/azure_devops/change_failure_rate.py
@@ -1,0 +1,94 @@
+"""Azure DevOps Server change failure rate collector."""
+
+from datetime import datetime
+from itertools import pairwise
+from typing import cast
+
+from collector_utilities.date_time import MAX_DATETIME, days_ago, parse_datetime
+from collector_utilities.type import URL, Value
+from model import Entities, Entity, SourceResponses
+
+from .base import AzureDevopsPipelines
+from .issues import AzureDevopsIssues
+
+
+class AzureDevopsChangeFailureRate(AzureDevopsIssues, AzureDevopsPipelines):
+    """Azure DevOps Server change failure rate collector."""
+
+    def _item_select_fields(self) -> list[str]:
+        """Extend to also request created date field to calculate issue age."""
+        return [*super()._item_select_fields(), "System.CreatedDate"]
+
+    async def _get_source_responses(self, *urls: URL) -> SourceResponses:
+        """Override to first get the AzureDevopsPipelines responses, then extend with AzureDevopsIssues."""
+        pipeline_ids = await self._active_pipelines()
+        api_pipelines_urls = [await self._api_pipelines_url(pipeline_id) for pipeline_id in pipeline_ids]
+        responses = await AzureDevopsPipelines._get_source_responses(self, *api_pipelines_urls)  # noqa: SLF001
+        responses.extend(await AzureDevopsIssues._get_source_responses(self, *urls))  # noqa: SLF001
+        return responses
+
+    async def _parse_entities(self, responses: SourceResponses) -> Entities:
+        issues = Entities()
+        deployments = Entities()
+        for response in responses:
+            if response.url.name == "wiql":  # lookup for AzureDevopsIssues ids, no entities
+                continue
+            if response.url.name == "workitemsbatch":  # issues from AzureDevopsIssues
+                issue_responses = SourceResponses(responses=[response, response], api_url=response.url)
+                # first response is ignored by AzureDevopsIssues._parse_entities, as it is for the workItems (wiql) api
+                parsed_issues = await AzureDevopsIssues._parse_entities(self, issue_responses)  # noqa: SLF001
+                issues.extend([issue for issue in parsed_issues if self._include_issue(issue)])
+            else:  # deployments from AzureDevopsPipelines
+                deployments.extend(await self._parse_pipeline_entities(response))
+        if deployments:
+            deployments.sort(key=lambda d: parse_datetime(d["build_date"]))  # oldest to newest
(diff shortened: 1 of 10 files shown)


components/collector/src/source_collectors/azure_devops/change_failure_rate.py
```diff
@@ -0,0 +1,93 @@
...
+        return responses
+
+    async def _parse_entities(self, responses: SourceResponses) -> Entities:
+        issues = Entities()
+        deployments = Entities()
+        for response in responses:
+            if response.url.name == "wiql":  # lookup for AzureDevopsIssues ids, no entities
+                continue
+            if response.url.name == "workitemsbatch":  # issues from AzureDevopsIssues
+                issue_responses = SourceResponses(responses=[None, response], api_url=response.url)
```
wkoot: Running into a mypy complaint here that `None` is not a `Response`. How to proceed?
1. Keeping track of the wiql response two lines above and passing that, but introduces unneeded complexity
2. Passing the same `response` twice because `AzureDevopsIssues._parse_entities` discards the first, but dirty
3. Modifying the `Responses` type to `list[Response | None]`, since `SourceResponses` also accepts `None`

fniessink: 4. Create a dummy `Response` and pass that as the first response? 

If 4. is not possible, I'd opt for 3.

components/collector/tests/source_collectors/azure_devops/test_change_failure_rate.py
```diff
@@ -0,0 +1,155 @@
...
+        get_response.json = AsyncMock(side_effect=get_request_json_side_effect)
+
+        post_response = AsyncMock()
+        post_json_return_value = post_request_json_side_effect[-1]
+        post_request_json_side_effect.append(STOP_SENTINEL)
+        post_response.json = AsyncMock(return_value=post_json_return_value, side_effect=post_request_json_side_effect)
+
+        if response_url_mock:
+            # this is the object within collector_utilities.type.Response.url, which cannot be patched directly
+            # because it uses reify to turn the yarl.URL (which contains the property .name) a cached property
```
fniessink:
```suggestion
            # because it uses reify to turn the yarl.URL (which contains the property .name) into a cached property
```

components/shared_code/src/shared_data_model/metrics.py
```diff
@@ -37,12 +37,15 @@
         documentation="""Because this metric is a "rate", it needs a numerator (the number of failed deployments
 in the look back period) and a denominator (the number of deployments in the look back period). The metric can be
 configured with a combination of Jira, and either Jenkins or GitLab CI as source. Jira is used as source for the number
-of failures (numerator). Either Jenkins or GitLab CI is used as source for the number of deployments (denominator).""",
+of failures (numerator). Either Jenkins or GitLab CI is used as source for the number of deployments (denominator).
+
+Configuring Azure DevOps as source for the metric, means that it is used both for number of failures (numerator) and
```
fniessink:
```suggestion
When configuring Azure DevOps as source for the metric it is used both for number of failures (numerator) and for
```

fniessink (approved):

components/collector/src/source_collectors/azure_devops/change_failure_rate.py
wkoot: The init requires many args (it's `aiohttp.ClientResponse`) which we don't really have in this context, although I could steal them from the current response.. but that would basically be the same as copying it. I'll go for 3

wkoot: This turned out not to be viable, as it would require typing `Responses` as a `Sequence` instead. This raises many issues in the codebase itself

PR #6745, opened 2023-08-01
Don't get the data model from the database if time travel is not needed.

fniessink:


components/api_server/src/model/queries.py
```diff
@@ -1,11 +1,15 @@
...
     # If the parameter key can't be found (this can happen when the parameter is removed from the data model),
     # err on the safe side and assume it was a password type
-    parameter_type = (
-        data_model["sources"].get(source_type, {}).get("parameters", {}).get(parameter, {"type": "password"})["type"]
-    )
-    return str(parameter_type) == "password"
+    if data_model is None:
+        data_model = DATA_MODEL.dict()
+    try:
+        return bool(data_model["sources"][source_type]["parameters"][parameter]["type"] == "password")
```
wkoot, line 13: Is the bool cast only needed to satisfy typing inspection? :(

components/api_server/src/routes/report.py
```diff
@@ -27,6 +31,12 @@
 from .plugins.auth_plugin import EDIT_REPORT_PERMISSION
 
 
+def report_not_found_response(report_uuid: ReportId) -> dict[str, str | bool]:
+    """Response for a report that can not be found."""
+    bottle.response.status = HTTPStatus.NOT_FOUND
```
wkoot: How is this passed? Looks a bit dirty to have it implicit

components/api_server/src/routes/report.py
```diff
@@ -126,32 +135,29 @@ def export_report_as_pdf(report_uuid: ReportId):
...
+    report = latest_report(database, report_uuid)
+    if not report:
+        return report_not_found_response(report_uuid)
+    if "public_key" in bottle.request.query:
+        public_key = bottle.request.query["public_key"]
+    else:  # default to own public key
+        document = database.secrets.find_one({"name": EXPORT_FIELDS_KEYS_NAME}, {"public_key": True, "_id": False})
+        if not document:  # pragma: no feature-test-cover
+            bottle.response.status = 500
+            bottle.response.content_type = "application/json"
```
wkoot: Why do we not do this on line 82, and the various other returns?

components/api_server/src/routes/report.py
```diff
@@ -126,32 +135,29 @@ def export_report_as_pdf(report_uuid: ReportId):
...
-    return None
+    report = latest_report(database, report_uuid)
+    if not report:
+        return report_not_found_response(report_uuid)
+    if "public_key" in bottle.request.query:
+        public_key = bottle.request.query["public_key"]
+    else:  # default to own public key
+        document = database.secrets.find_one({"name": EXPORT_FIELDS_KEYS_NAME}, {"public_key": True, "_id": False})
+        if not document:  # pragma: no feature-test-cover
+            bottle.response.status = 500
```
wkoot:
```suggestion
            bottle.response.status = HTTPStatus.BAD_REQUEST
```

components/api_server/src/routes/report.py
```diff
@@ -126,32 +135,29 @@ def export_report_as_pdf(report_uuid: ReportId):
...
+            bottle.response.content_type = "application/json"
+            return {"error": "Cannot find the public key of this Quality-time instance."}
```
wkoot:
```suggestion
            return {"ok": False, "error": "Cannot find the public key of this Quality-time instance."}
```

components/api_server/src/routes/report.py
```diff
@@ -178,8 +185,9 @@ def post_report_attribute(report_uuid: ReportId, report_attribute: str, database
 )
 def post_report_issue_tracker_attribute(report_uuid: ReportId, tracker_attribute: str, database: Database):
     """Set the issue tracker attribute."""
-    data_model = latest_datamodel(database)
-    report = latest_report(database, data_model, report_uuid)
+    report = latest_report(database, report_uuid)
+    if report is None:
+        return report_not_found_response(report_uuid)
```
wkoot: Hmm this pattern is so common, is there a nice way to make a kind of `latest_report_or_not_found` wrapper?

components/api_server/src/routes/report.py
```diff
@@ -202,28 +210,30 @@ def post_report_issue_tracker_attribute(report_uuid: ReportId, tracker_attribute
...
 def get_report_issue_tracker_options(report_uuid: ReportId, database: Database):
     """Get options for the issue tracker attributes such as project key and issue type."""
-    data_model = latest_datamodel(database)
-    report = latest_report(database, data_model, report_uuid)
+    report = latest_report(database, report_uuid)
+    if report is None:
+        return report_not_found_response(report_uuid)
     issue_tracker = report.issue_tracker()
-    return issue_tracker.get_options().as_dict()
+    return issue_tracker.get_options().as_dict() | {"ok": True}
```
wkoot, line 248: Why is an empty options dict okay? Is it not better to have the `get_options()` return the "ok" key, I assume it should already return this key anyway?

components/api_server/src/routes/source.py
```diff
@@ -200,16 +200,16 @@ def _source_description(data_model, items, edit_scope, parameter_key, old_value)
...
 
-def _availability_checks(data_model, source, parameter_key: str) -> list[dict[str, str | int]]:
+def _availability_checks(source, parameter_key: str) -> list[dict[str, str | int]]:
     """Check the availability of the URLs."""
-    source_model = data_model["sources"][source["type"]]
-    parameters = source_model["parameters"]
-    token_validation_path = parameters.get("private_token", {}).get("validation_path", "")
+    parameters = DATA_MODEL.sources[source["type"]].parameters
+    private_token = cast(PrivateToken, parameters.get("private_token"))
+    token_validation_path = private_token.validation_path if private_token else ""
```
wkoot, line 207: Too bad, it seems that casting None circumvents the `__init__`. Otherwise, an empty object would've been nice :(

components/api_server/src/model/queries.py
fniessink: Yes, without the `bool()` mypy says:

```console
src/model/queries.py:13: error: Returning Any from function declared to return "bool"  [no-any-return]
```

I guess mypy can't determine the type returned by `==/__eq__` because `data_model` is typed as a `dict` without specification for the keys and values.

components/api_server/src/routes/report.py
fniessink: Setting the response content type to JSON is not needed when returning a dict according to the Bottle docs: "As mentioned above, Python dictionaries (or subclasses thereof) are automatically transformed into JSON strings and returned to the browser with the Content-Type header set to application/json." 

I've removed the `bottle.response.content_type = "application/json"`.

components/api_server/src/routes/report.py
fniessink: Setting the response status is indeed a bit dirty in Bottle, but it seems to be the 'right' way: "The [Response](https://bottlepy.org/docs/dev/api.html#bottle.Response) class stores the HTTP status code as well as headers and cookies that are to be sent to the client. Similar to [bottle.request](https://bottlepy.org/docs/dev/api.html#bottle.request) there is a thread-local [bottle.response](https://bottlepy.org/docs/dev/api.html#bottle.response) instance that can be used to adjust the current response."

components/api_server/src/routes/report.py
fniessink: It's not a bad request, the problem is that the server has no public key in the database. This shouldn't happen.

components/api_server/src/routes/report.py
fniessink: I've introduced a `with_report` decorator that does this.

components/api_server/src/routes/report.py
fniessink: I'd prefer not to, because `IssueTracker.get_options()` isn't a route so it does not know that it's return value is passed as a response.

components/api_server/src/routes/source.py
fniessink: I guess the ideal solution would if `parameters` was a typed dict so mypy would know that the values of the key `private_token` are of type `PrivateToken`. But I'm not going to touch the data model before we've upgraded to Pydantic 2.

components/api_server/src/routes/report.py
wkoot: Yeah I see, plus it's also how we're doing it throughout

components/api_server/src/routes/report.py
wkoot: Whoops, it was a lazy copypaste I think. Meant internal server error of course :-)

components/api_server/src/routes/report.py
wkoot: But should the dict then not always be merged with `{"ok": True}`?

fniessink: That's what `| {"ok": True}` does!

wkoot: Whoops new syntax. I read it as a NOR

PR #6758, opened 2023-08-02
Include the selected tags in the URL query parameter so the selected …

### LINKED ISSUE #4551: Include tag selection in URL parameters so that it can be shared
When copying the URL from a report from the location bar in the browser, the selection for a tag is not included. This way, the one receiving the link and clicking it does not get the same view as the sender.

When a specific tag is selected in a report, this should be included in the URL parameters, like 'URL?report_date=2022-09-07&tag=maintainability'.

fniessink:
…tags are retained when navigating between reports and when the URL is shared with another user. Closes #4551.

### DIFF
--- a/components/frontend/src/report/Report.js
+++ b/components/frontend/src/report/Report.js
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext } from 'react';
 import { Message } from 'semantic-ui-react';
 import { DataModel } from '../context/DataModel';
 import { accessGranted, EDIT_REPORT_PERMISSION, Permissions } from '../context/Permissions';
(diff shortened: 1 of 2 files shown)


components/frontend/src/report/Report.js
```diff
@@ -118,12 +118,7 @@ export function Report({
         window.scrollBy(0, 163);  // Correct for menubar and subject title margin
     }
 
-    const [selectedTags, setSelectedTags] = useState([]);
-    useEffect(() => {
-        // Make sure we only filter by tags that are actually used in this report
-        setSelectedTags(prev_tags => prev_tags.filter(tag => getReportTags(report).includes(tag)))
-    }, [report]);
-
+    const [selectedTags, toggleSelectedTag] = useURLSearchQuery("selected_tags", "array");
```
wkoot, line 121: Nice, already existing logic!

wkoot (approved):

PR #7023, opened 2023-09-15
Store environment variables for components in .env files.

### LINKED ISSUE #491: Read server configs from Docker env files
I would like to read all server settings (e.g. DATABASE_URL, DEBUG, LDAP_URL) from the environment instead of having these hardcoded through several python files. Docker env files might be a good place to start as it would allow easy differentiation between otap stages.

I suppose application "secrets" could be injected by adding an extra env file (gitignored); as it seems you can enter as many as you want;

https://stackoverflow.com/questions/32130123/passing-multiple-env-files-in-docker-run-command#answer-32130684 (this example isn't using docker-compose though)

fniessink:
Fixes #491.

### DIFF
--- a/components/api_server/src/healthcheck.py
+++ b/components/api_server/src/healthcheck.py
@@ -1,11 +1,11 @@
 """Health check."""
 
-from http import HTTPStatus
-from os import getenv
-from urllib.request import urlopen
+import os
 import sys
+from http import HTTPStatus
+from urllib.request import urlopen
 
-api_server_port = getenv("API_SERVER_PORT", "5001")
+api_server_port = os.environ["API_SERVER_PORT"]
 exit_code = 1
 try:
     with urlopen(f"http://localhost:{api_server_port}/api/health") as response:  # nosec B310 # noqa: S310
--- a/components/api_server/src/initialization/database.py
+++ b/components/api_server/src/initialization/database.py
@@ -1,12 +1,12 @@
 """Database initialization."""
 
 import logging
-import os
 
 import pymongo
 from pymongo.database import Database
 
 from shared.initialization.database import client
+from shared.utils.env import getenv
 
 from .datamodel import import_datamodel
 from .report import import_example_reports, initialize_reports_overview
@@ -26,7 +26,7 @@ def init_database() -> Database:  # pragma: no feature-test-cover
     import_datamodel(database)
     initialize_secrets(database)
     initialize_reports_overview(database)
-    if os.environ.get("LOAD_EXAMPLE_REPORTS", "True").lower() == "true":
+    if getenv("LOAD_EXAMPLE_REPORTS").lower() == "true":
         import_example_reports(database)
     return database
 
--- a/components/api_server/src/quality_time_server.py
+++ b/components/api_server/src/quality_time_server.py
@@ -5,9 +5,10 @@ from gevent import monkey
 monkey.patch_all()
 
 import logging
-import os
 
 import bottle
+
+from shared.utils.env import getenv, loadenv
 
 from initialization.database import init_database
 from initialization.bottle import init_bottle
@@ -15,12 +16,13 @@ from initialization.bottle import init_bottle
 
 def serve() -> None:  # pragma: no feature-test-cover
     """Connect to the database and start the application server."""
-    log_level = str(os.getenv("API_SERVER_LOG_LEVEL", "WARNING"))
+    loadenv(".env", "default.env")
+    log_level = getenv("API_SERVER_LOG_LEVEL")
     logger = logging.getLogger()
     logger.setLevel(log_level)
     database = init_database()
     init_bottle(database)
-    server_port = os.getenv("API_SERVER_PORT", "5001")
+    server_port = getenv("API_SERVER_PORT")
     bottle.run(server="gevent", host="0.0.0.0", port=server_port, reloader=True, log=logger)  # nosec, # noqa: S104
 
 
--- a/components/api_server/src/routes/auth.py
+++ b/components/api_server/src/routes/auth.py
@@ -3,7 +3,6 @@
 import base64
 import hashlib
 import logging
-import os
 import re
 import string
 from datetime import datetime, timedelta, UTC
@@ -14,6 +13,8 @@ import bottle
 from ldap3 import ALL, Connection, Server, ServerPool, AUTO_BIND_NO_TLS
 from ldap3.core import exceptions
 from pymongo.database import Database
+
+from shared.utils.env import getenv
 
 from database.users import upsert_user, get_user
 from database import sessions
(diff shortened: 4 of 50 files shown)


docker/docker-compose.override.yml
```diff
@@ -9,20 +9,24 @@ services:
...
+      - DATABASE_HOST=localhost
       - NOTIFIER_LOG_LEVEL=${NOTIFIER_LOG_LEVEL:-INFO}
     build:
       context: ../components
       dockerfile: notifier/Dockerfile
   api_server:
     environment:
       - API_SERVER_LOG_LEVEL=${API_SERVER_LOG_LEVEL:-INFO}
+      - DATABASE_HOST=localhost
+      - LDAP_URL=ldap://localhost:389
```
wkoot: Should this not be the ldap container instead of localhost?

tests/feature_tests/ci/test.sh
```diff
@@ -14,7 +14,7 @@ python3 -m venv venv
 . venv/bin/activate
 ci/pip-install.sh
 coverage erase
-RENDERER_HOST=localhost python tests/quality_time_api_server_under_coverage.py &> ../../build/quality_time_api_server.log &
+env RENDERER_HOST=localhost DATABASE_HOST=localhost LDAP_URL=ldap://localhost:389 python tests/quality_time_api_server_under_coverage.py &> ../../build/quality_time_api_server.log &
```
wkoot: Why `env` ?

components/api_server/src/healthcheck.py
```diff
@@ -5,7 +5,7 @@
 from urllib.request import urlopen
 import sys
 
-api_server_port = getenv("API_SERVER_PORT", "5001")
+api_server_port = getenv("API_SERVER_PORT")
```
wkoot: Might be better to then `os.environ["API_SERVER_PORT"]`, as `getenv` is essentially a `os.environ.get` which returns `None` instead of throwing an exception when the var isn't defined

components/api_server/src/initialization/database.py
```diff
@@ -26,7 +26,7 @@ def init_database() -> Database:  # pragma: no feature-test-cover
     import_datamodel(database)
     initialize_secrets(database)
     initialize_reports_overview(database)
-    if os.environ.get("LOAD_EXAMPLE_REPORTS", "True").lower() == "true":
+    if getenv("LOAD_EXAMPLE_REPORTS").lower() == "true":
```
wkoot, line 29: Why not explicit bool cast?

components/api_server/src/quality_time_server.py
```diff
@@ -5,22 +5,24 @@
...
+from shared.utils.env import getenv, loadenv
+
 from initialization.database import init_database
 from initialization.bottle import init_bottle
 
 
 def serve() -> None:  # pragma: no feature-test-cover
     """Connect to the database and start the application server."""
-    log_level = str(os.getenv("API_SERVER_LOG_LEVEL", "WARNING"))
+    loadenv(".env", "default.env")
```
wkoot, line 19: I don't think we should load the `default.env` which contains example values

components/api_server/src/routes/auth.py
```diff
@@ -78,12 +79,12 @@ def get_credentials() -> tuple[str, str]:
...
     """Get LDAP config from environment."""
-    ldap_search_filter_template = os.environ.get("LDAP_SEARCH_FILTER", "(|(uid=$username)(cn=$username))")
+    ldap_search_filter_template = getenv("LDAP_SEARCH_FILTER")
     return {
-        "ldap_root_dn": os.environ.get("LDAP_ROOT_DN", "dc=example,dc=org"),
-        "ldap_urls": os.environ.get("LDAP_URL", "ldap://localhost:389").split(","),
-        "ldap_lookup_user_dn": os.environ.get("LDAP_LOOKUP_USER_DN", "cn=admin,dc=example,dc=org"),
-        "ldap_lookup_user_pw": os.environ.get("LDAP_LOOKUP_USER_PASSWORD", "admin"),
+        "ldap_root_dn": getenv("LDAP_ROOT_DN"),
+        "ldap_urls": getenv("LDAP_URL").split(","),
```
wkoot, line 85: Too bad that the env file is not json.. Maybe it's worthwhile to implement json instead?

components/api_server/src/routes/auth.py
```diff
@@ -130,8 +131,8 @@ def verify_user(database: Database, username: str, password: str) -> User:
 @bottle.post("/api/v3/login", authentication_required=False)
 def login(database: Database) -> dict[str, bool | str]:
     """Log the user in. Add credentials as JSON payload, e.g. {username: 'user', password: 'pass'}."""
-    if os.environ.get("FORWARD_AUTH_ENABLED", "").lower() == "true":  # pragma: no feature-test-cover
-        forward_auth_header = str(os.environ.get("FORWARD_AUTH_HEADER", "X-Forwarded-User"))
+    if getenv("FORWARD_AUTH_ENABLED").lower() == "true":  # pragma: no feature-test-cover
```
wkoot, line 134: bool cast?

components/shared_code/src/shared/utils/env.py
```diff
@@ -0,0 +1,19 @@
...
+
+def getenv(key: str) -> str:
+    """Return the environment variable's value as string."""
+    return os.getenv(key, "")
+
+
+def loadenv(*filenames: str) -> None:
+    """Load the environment variables from the files.
+
+    By default, load_dotenv doesn't override existing environment variables so pass file with the defaults last.
```
wkoot, line 16: So a user would need to specify all settings, since the defaults contain example values?

components/shared_code/src/shared/initialization/database.py
```diff
@@ -1,21 +1,24 @@
...
 
-DEFAULT_DATABASE_URL = "mongodb://root:root@localhost:27017"
+from shared.utils.env import getenv
 
 
-def client(url: str | None = DEFAULT_DATABASE_URL) -> pymongo.MongoClient:  # pragma: no feature-test-cover
+def client() -> pymongo.MongoClient:  # pragma: no feature-test-cover
     """Return a pymongo client."""
-    database_url = os.environ.get("DATABASE_URL", url)
+    # DATABASE_URL is deprecated, use the new environment variables if DATABASE_URL is not set
```
wkoot, line 12: Does it make sense to use `DeprecationWarning` or at least warning logging?

docker/docker-compose.override.yml
fniessink: Yeah, that was a mistake. Removed.

tests/feature_tests/ci/test.sh
fniessink: I wasn't sure the environment variables were influencing the later docker compose call. Now running the tests without `env`. I'll remove it, if possible.

components/api_server/src/initialization/database.py
fniessink: `getenv` (note this is our own `getenv`, not `os.getenv`) always returns a string.

components/api_server/src/quality_time_server.py
fniessink: They are not example values, they are the default values for the environment variables.

components/api_server/src/routes/auth.py
fniessink: I don't see what that would bring us? I think this is the only place where an environment variable is a little bit more complicated than a simple string.

components/api_server/src/routes/auth.py
fniessink: Nope, `getenv` always returns a string.

components/shared_code/src/shared/utils/env.py
fniessink: No, the defaults.env files contain the default values for the environment variables, as suggested by the name ;-)

components/shared_code/src/shared/initialization/database.py
fniessink: Good point.

components/api_server/src/initialization/database.py
wkoot: I meant, `if bool(getenv(..)):`, but it's incorrect as it'll always return true as long as there's no empty string :(

components/api_server/src/quality_time_server.py
wkoot: The LDAP settings contain: `LDAP_ROOT_DN=dc=example,dc=org`

PR #7375, opened 2023-10-29
Implement ChangeFailureRate for Jira and Jenkins or GitLab CI

### LINKED ISSUE #7374: Users can configure Jira as source for the metric (but only the number of failures) and either Jenkins or GitLab CI as source for the metric (for the number of deployments).
Extends #3499.

wkoot:
Closes #7374

### DIFF
--- a/components/collector/src/metric_collectors/change_failure_rate.py
+++ b/components/collector/src/metric_collectors/change_failure_rate.py
@@ -1,0 +1,83 @@
+"""Change failure rate collector."""
+
+from collections.abc import Sequence
+from datetime import datetime
+from itertools import pairwise
+from typing import ClassVar
+
+from base_collectors import MetricCollector
+from collector_utilities.date_time import MAX_DATETIME, parse_datetime
+from model import Entities, Entity, MetricMeasurement, SourceMeasurement
+
+
+class ChangeFailureRate(MetricCollector):
+    """Change failure rate collector."""
+
+    # Partial sources for issues and deployments, which need to be combined within this metric
+    ISSUE_SOURCE_TYPES: ClassVar[list[str]] = ["jira"]
+    DEPLOYMENT_SOURCE_TYPES: ClassVar[list[str]] = ["gitlab", "jenkins"]
+
+    async def collect(self) -> MetricMeasurement | None:
+        """Override to calculate the composite value."""
+        if (measurement := await super().collect()) is None:
+            return None
+        issues = self.issue_entities(measurement.sources)
+        deployments = self.deployment_entities(measurement.sources)
+        if deployments:
+            deployments.sort(key=lambda d: self.deployment_timestamp(d))  # oldest to newest
+            for this_dep, next_dep in pairwise(deployments):
+                this_dep["failed"] = self.issues_in_interval(
+                    issues,
+                    self.deployment_timestamp(this_dep),
+                    self.deployment_timestamp(next_dep),
+                )
+            latest_dep = deployments[-1]  # latest deployment was not iterated by pairwise; do it separately
+            latest_dep["failed"] = self.issues_in_interval(issues, self.deployment_timestamp(latest_dep), MAX_DATETIME)
+        for source in measurement.sources:
+            if self.source_type(source) in self.ISSUE_SOURCE_TYPES:
+                # Set the values to zero as this metric only counts failed deployments
+                source.total = source.value = "0"
+            if self.source_type(source) in self.DEPLOYMENT_SOURCE_TYPES:
+                source.total = str(len(source.entities))
+                # Only select failed deployments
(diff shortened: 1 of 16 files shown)


components/collector/src/metric_collectors/change_failure_rate.py
```diff
@@ -0,0 +1,78 @@
...
+        issues = self.issue_entities(measurement.sources)
+        for deployment in deployments:
+            issue_length = len(issues)
+            if issue_length == 0:
+                break  # no more issues; there are no more failed deployments
+            issues = Entities(
+                issue for issue in issues if self.deployment_timestamp(deployment) > self.issue_timestamp(issue)
+            )
+            if len(issues) != issue_length:
+                deployment["failed"] = True  # there were issues after this deployment, so it counts as failed
```
fniessink: I don't understand how this works: you iterate through the deployments from new to old with the goal to find out which deployments failed. I assume a deployment is considered failed if there are issues between the deployment and the next deployment (if any), but I don't see that happening in the code. Moreover, the issue filtering in line 31 seems to find the issues *before* the deployment. I'd expect this loop to look something like this:

```suggestion
        next_timestamp = datetime.max
        for deployment in deployments:
            this_timestamp = self.deployment_timestamp(deployment)
            deployment["failed"] = bool(issues_between(issues, this_timestamp, next_timestamp))
            next_timestamp = this_timestamp
```

And:

```python
    @staticmethod
    def issues_between(issues, timestamp1, timestamp2) -> Entities:
        """Return the issues between the two timestamps."""
        return Entities(issue for issue in issues if timestamp1 < self.issue_timestamp(issue) <= timestamp2)
```

components/shared_code/src/shared_data_model/metrics.py
```diff
@@ -44,6 +44,23 @@
...
+    "change_failure_rate": Metric(
+        name="Change failure rate",
+        description="The percentage of deployments causing a failure in production.",
+        rationale="The change failure rate is an indicator of the DevOps effectiveness of a team.",
+        documentation="""Because this metric is a "rate", it needs a numerator (the number of failed deployments
+in the look back period) and a denominator (the number of deployments in the look back period). The metric can be
+configured with a combination of Jira, and either Jenkins or GitLab CI as source.
+
+Configuring Jira as source for the metric, means that it is used as the number of failures (numerator). It needs to be
+configured with either Jenkins or GitLab CI as source for the number of deployments (denominator).""",
```
fniessink:
```suggestion
configured with a combination of Jira, and either Jenkins or GitLab CI as source. Jira is used as source for the number of failures (numerator). Either Jenkins or GitLab CI is used as source for the number of deployments (denominator).""",
```

components/shared_code/src/shared_data_model/sources/jenkins.py
```diff
@@ -37,17 +37,19 @@ def jenkins_access_parameters(*args, **kwargs) -> dict[str, Parameter]:
...
 ALL_JENKINS_METRICS = [
+    "change_failure_rate",
     "failed_jobs",
     "job_runs_within_time_period",
     "source_up_to_dateness",
     "source_version",
     "unused_jobs",
 ]
 
+_JOB_ENTITY_NAME_NAME = "Job name"
```
fniessink, line 48: LOL.

fniessink (changes requested):

components/shared_code/src/shared_data_model/sources/jenkins.py
wkoot: the alternative was.. `_JOB_ENTITY_NAME_ATTRIBUTE_NAME` :D

components/collector/src/metric_collectors/change_failure_rate.py
wkoot: The `deployments` are sorted from newest to oldest (line 24), `issues` are only those for which deploy dt > issue dt (line 31). The loop iterates backwards so it can deal with the open ended interval `[deploy_dt -> deploy_dt | inf)`

components/shared_code/src/shared_data_model/metrics.py
wkoot: This does read more clearly, but the originally intended help text in #6722 was:
```The metric can either be configured with Azure DevOps as a source, or a combination of Jira and Jenkins or GitLab CI as source. Configuring Azure DevOps as source for the metric, means that it is used both for number of failures (numerator) and number of deployments (denominator). Azure DevOps needs to be added as source only once. Configuring Jira as source for the metric, means that it is used as the number of failures (numerator). It needs to be configured with either Jenkins or GitLab CI as source for the number of deployments (denominator).```

fniessink (approved):

PR #7409, opened 2023-11-03
Performance improvements

### LINKED ISSUE #7371: Slow loading speeds
**Describe the bug**
For our project (that contains quite a lot of metrics) the loading speeds of various pages are quite long. The main overview page almost always takes more than 10 seconds to load, which seems quite slow. It makes navigating through the pages quite frustrating. 
It seems that for the main overview (which does not show a whole lot of information) a lot of data is queried, leading to a lot of CPU usage on the database.

**Expected behaviour**
Loading times are more acceptable (at most a few seconds).

 - Running on ICTU ECO 2 (NLA project)
 - Quality-time version 5.1

fniessink:
- In the /api/v3/report endpoint, get the recent measurements for all metrics in one database call, instead of using one database call per report.
- Create a measurements index on both the start and the end field instead of just on the start field.
- Create an index on the report_uuid field of the reports to speed up the `distinct` query.

### DIFF
--- a/components/api_server/src/database/measurements.py
+++ b/components/api_server/src/database/measurements.py
@@ -37,7 +37,9 @@ def count_measurements(database: Database) -> int:
 
 
 def measurements_in_period(database: Database, min_iso_timestamp: str, max_iso_timestamp: str):
-    """Return recent measurements for the specified metrics, without source details and issue status."""
+    """Return recent measurements within the specified period, without source details and issue status."""
+    # Return measurements that partially or completely overlap with the period, meaning they have their start before
+    # the end of the period (max_iso_timestamp) and their end after the start of the period (min_iso_timestamp):
     measurement_filter = {"start": {"$lte": max_iso_timestamp or iso_timestamp()}, "end": {"$gte": min_iso_timestamp}}
     return list(database.measurements.find(measurement_filter, sort=START_ASCENDING, projection=NO_MEASUREMENT_DETAILS))
 
@@ -53,7 +55,7 @@ def all_metric_measurements(database: Database, metric_uuid: MetricId, max_iso_t
     all_measurements_stripped = list(
         database.measurements.find(measurement_filter, sort=START_ASCENDING, projection=NO_MEASUREMENT_DETAILS),
     )
-    return list(all_measurements_stripped)[:-1] + [latest_measurement]
+    return all_measurements_stripped[:-1] + [latest_measurement]
 
 
 def recent_measurements(
@@ -70,6 +72,7 @@ def recent_measurements(
     measurements_by_metric_uuid: dict[MetricId, list] = {}
     for measurement_dict in measurements:
         metric_uuid = measurement_dict["metric_uuid"]
+        # Only process the measurement if its metric exists:
         if metric := metrics_dict.get(metric_uuid):  # pragma: no feature-test-cover
             measurement = Measurement(metric, measurement_dict)
             measurements_by_metric_uuid.setdefault(metric_uuid, []).append(measurement)
--- a/components/api_server/src/database/measurements.py
+++ b/components/api_server/src/database/measurements.py
@@ -11,12 +11,21 @@ from shared.model.metric import Metric
 from shared.utils.functions import iso_timestamp
 from shared.utils.type import MetricId
 
+# Projection options
+NO_ID = {"_id": False}
+NO_SOURCE_DETAILS = NO_ID | {"sources.entities": False, "sources.entity_user_data": False}
+NO_MEASUREMENT_DETAILS = NO_SOURCE_DETAILS | {"issue_status": False}
+
+# Sort optins
+START_ASCENDING = [("start", pymongo.ASCENDING), ("end", pymongo.DESCENDING)]
+START_DESCENDING = [("start", pymongo.DESCENDING)]
+
 
 def changelog(database: Database, nr_changes: int, **uuids):
     """Return the changelog for the measurements belonging to the items with the specific uuids."""
     return database.measurements.find(
         filter={"delta.uuids": {"$in": list(uuids.values())}},
-        sort=[("start", pymongo.DESCENDING)],
+        sort=START_DESCENDING,
         limit=nr_changes,
         projection=["delta", "start"],
     )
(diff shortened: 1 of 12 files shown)


components/api_server/src/initialization/database.py
```diff
@@ -42,9 +42,9 @@ def set_feature_compatibility_version(admin_database: Database) -> None:
 def create_indexes(database: Database) -> None:
     """Create any indexes."""
     database.datamodels.create_index("timestamp")
-    database.reports.create_index("timestamp")
+    database.reports.create_index([("timestamp", pymongo.ASCENDING), ("report_uuid", pymongo.ASCENDING)])
     database.users.create_index("username", unique=True)
-    start_index = pymongo.IndexModel([("start", pymongo.ASCENDING)])
+    start_index = pymongo.IndexModel([("start", pymongo.ASCENDING), ("end", pymongo.DESCENDING)])
```
wkoot: Is the name `start_index` still accurate?

components/api_server/src/database/reports.py
```diff
@@ -81,7 +81,7 @@ def latest_reports(database: Database) -> list[Report]:
 
 def latest_reports_before_timestamp(database: Database, data_model: dict, max_iso_timestamp: str) -> list[Report]:
     """Return the latest, undeleted, reports in the reports collection before the max timestamp."""
-    report_filter = {"timestamp": {"$lt": max_iso_timestamp}} if max_iso_timestamp else {}
+    report_filter = {"timestamp": {"$lte": max_iso_timestamp}} if max_iso_timestamp else {}
```
wkoot, line 84: Why the addition of equal? Was the max timestamp not something very far in the future?

tests/feature_tests/src/steps/report.py
```diff
@@ -131,13 +131,12 @@ def import_failed(context: Context) -> None:
...
-def get_measurements(context: Context, has_or_had: str, expected_number: str) -> None:
+@then('the report has "{expected_number}" measurements')
+def get_measurements(context: Context, expected_number: str) -> None:
     """Get the recent measurements of a report."""
-    if has_or_had == "had":
-        context.report_date = "2020-11-17T10:00:00Z"
-        context.min_report_date = "2020-11-16T00:00:00Z"
-
+    now = datetime.now(tz=UTC).replace(microsecond=0)
+    context.report_date = (now + timedelta(days=10)).isoformat()
```
wkoot, line 138: Why is this needed?

components/api_server/src/database/measurements.py
```diff
@@ -11,12 +11,21 @@
 from shared.utils.functions import iso_timestamp
 from shared.utils.type import MetricId
 
+# Projection options
+NO_ID = {"_id": False}
+NO_SOURCE_DETAILS = NO_ID | {"sources.entities": False, "sources.entity_user_data": False}
+NO_MEASUREMENT_DETAILS = NO_SOURCE_DETAILS | {"issue_status": False}
+
+# Sort optins
+START_ASCENDING = [("start", pymongo.ASCENDING), ("end", pymongo.DESCENDING)]
```
wkoot, line 20: Why does this also contain "end"?

components/api_server/src/database/measurements.py
```diff
@@ -27,52 +36,24 @@ def count_measurements(database: Database) -> int:
...
-                "_id": False,
-                "sources.entities": False,
-                "sources.entity_user_data": False,
-                "issue_status": False,
-            },
-        ),
-    )
+def measurements_in_period(database: Database, min_iso_timestamp: str, max_iso_timestamp: str):
+    """Return recent measurements for the specified metrics, without source details and issue status."""
+    measurement_filter = {"start": {"$lte": max_iso_timestamp or iso_timestamp()}, "end": {"$gte": min_iso_timestamp}}
```
wkoot, line 43: I don't get this; start on less-than-or-equals-to MAX, end on greater-than-or-equals to MIN, but sort ascending?
I would expect either the start/end to be reversed, or the sorting to be descending. Perhaps describe/comment?

components/api_server/src/database/measurements.py
```diff
@@ -27,52 +36,24 @@ def count_measurements(database: Database) -> int:
...
     if max_iso_timestamp:
-        measurement_filter["start"] = {"$lt": max_iso_timestamp}
-    latest_measurement_complete = database.measurements.find_one(
-        measurement_filter,
-        sort=[("start", pymongo.DESCENDING)],
-        projection={"_id": False},
-    )
-    if not latest_measurement_complete:
+        measurement_filter["start"] = {"$lte": max_iso_timestamp}
+    latest_measurement = database.measurements.find_one(measurement_filter, sort=START_ASCENDING, projection=NO_ID)
```
wkoot, line 52: Why do we use the projection?

components/api_server/src/database/measurements.py
```diff
@@ -27,52 +36,24 @@ def count_measurements(database: Database) -> int:
...
+    if not latest_measurement:
         return []
-    all_measurements_stripped = measurements_by_metric(database, metric_uuid, max_iso_timestamp=max_iso_timestamp)
-    return list(all_measurements_stripped)[:-1] + [latest_measurement_complete]
+    all_measurements_stripped = list(
+        database.measurements.find(measurement_filter, sort=START_ASCENDING, projection=NO_MEASUREMENT_DETAILS),
+    )
+    return list(all_measurements_stripped)[:-1] + [latest_measurement]
```
wkoot: The `list` cast is duplicate

components/api_server/src/database/measurements.py
```diff
@@ -27,52 +36,24 @@ def count_measurements(database: Database) -> int:
...
-        projection={"_id": False},
-    )
-    if not latest_measurement_complete:
+        measurement_filter["start"] = {"$lte": max_iso_timestamp}
+    latest_measurement = database.measurements.find_one(measurement_filter, sort=START_ASCENDING, projection=NO_ID)
+    if not latest_measurement:
         return []
-    all_measurements_stripped = measurements_by_metric(database, metric_uuid, max_iso_timestamp=max_iso_timestamp)
-    return list(all_measurements_stripped)[:-1] + [latest_measurement_complete]
+    all_measurements_stripped = list(
```
wkoot, line 55: Would it be a performance improvement to yield a generator instead, i.e. if we don't iterate through the *entire* history? Although that would make more sense in desc sort

components/api_server/src/database/measurements.py
```diff
@@ -81,21 +62,15 @@ def recent_measurements(
...
-    )
+    measurement_filter: dict[str, Any] = {"start": {"$lte": max_iso_timestamp}, "end": {"$gte": min_iso_timestamp}}
+    measurements = database.measurements.find(measurement_filter, sort=START_ASCENDING, projection=NO_SOURCE_DETAILS)
     measurements_by_metric_uuid: dict[MetricId, list] = {}
     for measurement_dict in measurements:
         metric_uuid = measurement_dict["metric_uuid"]
-        metric = metrics_dict[metric_uuid]
-        measurement = Measurement(metric, measurement_dict)
-        measurements_by_metric_uuid.setdefault(metric_uuid, []).append(measurement)
+        if metric := metrics_dict.get(metric_uuid):  # pragma: no feature-test-cover
```
wkoot, line 76: All calling code can handle getting an empty result `if not metric`?

components/api_server/src/initialization/database.py
fniessink: Hmm, it's still an index on start first, but maybe `period_index` is better.

components/api_server/src/database/reports.py
fniessink: I got a failing testcase where the max_iso_timestamp was equal to the current time. This fixed it. I also think the less-than-or-equal is more intuitive in the case of timestamps.

components/api_server/src/database/measurements.py
fniessink: So that it matches the index, see first alinea here: https://www.mongodb.com/docs/manual/tutorial/sort-results-with-indexes/#std-label-sorting-with-indexes

tests/feature_tests/src/steps/report.py
fniessink: Because the endpoint now always needs a period (`min_report_date` and `report_date`). I tried to get rid of optional parameters that were basically only optional for sake of the tests.

components/api_server/src/database/measurements.py
fniessink: Comment added:

```python
    # Return measurements that partially or completely overlap with the period, meaning they have their start before
    # the end of the period (max_iso_timestamp) and their end after the start of the period (min_iso_timestamp):
```

components/api_server/src/database/measurements.py
fniessink: In general, to reduce data sent over the wire. Why the id is removed in particular I don't recall.

components/api_server/src/database/measurements.py
fniessink: I don't think it matters, all data is sent to the frontend anyway.

components/api_server/src/database/measurements.py
fniessink: The `if` is needed because there may be measurements for metrics that have been deleted. I'll add a comment.

tests/feature_tests/src/steps/report.py
wkoot: I was wondering about the +10 days in particular :-) But shouldn't be a problem

PR #7611, opened 2023-12-05
Add search endpoints to the API. Closes #7579.

### LINKED ISSUE #7579: Add an object search endpoint
It's hard to change credentials for sources via the API. One has to look up the source uuid. If more sources use the same URL/credentials they still have to be changed one by one.

The current endpoint for changing source parameters requires a source uuid:

`https://quality-time.example.org/api/v3/source/<source_uuid>/parameter/<source_attribute>`

Make it possible to look up the uuids of objects (sources, metrics, subjects, and reports) given an object type, an attribute name and an attribute value.

For example, give me all uuids of sources (=object type) with URL (=attribute name) equals "https://example.org" (=attribute value).

`https://quality-time.example.org/api/v3/<object_type>/search`

Note: All object types have attributes, but sources also have parameters. In the case of sources, search should work for both attributes and parameters. 

The attribute value is to be sent in the payload:

```json
{
    "<attribute_or_parameter_name>": "<value>"
}
```

The response is a list of uuids:

```json
{
    "object_type": "<object_type>",
    "count": 42,
    "uuids": [
    ]
}
```

**Constraints**

- The endpoint takes exactly one key/value pair. 

**Optional**

- Refactor existing code to use this endpoint (or the code behind the endpoint).

fniessink:


components/api_server/src/routes/search.py
```diff
@@ -0,0 +1,90 @@
...
+
+from database.reports import latest_reports
+from model.iterators import metrics, sources, subjects
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from model.report import Report
+
+
+DomainObjectType = Literal["metric", "report", "source", "subject"]
```
wkoot, line 20: Ha, inventive naming choice!

components/api_server/src/routes/search.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class SearchResults(TypedDict):
+    """Successful search response."""
+
+    domain_object_type: DomainObjectType
+    ok: Literal[True]
```
wkoot: Might as well set it to be default `True`,  if it's the only possible value?

components/api_server/src/routes/search.py
```diff
@@ -0,0 +1,90 @@
...
+    domain_object_type: DomainObjectType
+    error: str
+    ok: Literal[False]
+
+
+class SearchError(TypedDict):
+    """Failed search response."""
+
+    domain_object_type: DomainObjectType
+    search_query: NotRequired[dict[str, str]]
```
wkoot: Could define SearchQuery as a type and reuse?

components/api_server/src/routes/search.py
```diff
@@ -0,0 +1,90 @@
...
+    ok: Literal[False]
+
+
+def match_attribute(
+    domain_object_type: DomainObjectType, domain_object, attribute_name: str, attribute_value: str
+) -> bool:
+    """Return whether the domain object has an attribute with the specified name and value."""
+    if domain_object_type == "source":
+        parameters = domain_object["parameters"]
+        if attribute_name in parameters and parameters[attribute_name] == attribute_value:
```
wkoot, line 73: Why not use `parameters.get(attribute_name)` and skip the 'in' check?

components/api_server/src/routes/search.py
```diff
@@ -0,0 +1,90 @@
...
+        search_query = {attribute_name: attribute_value}
+    except Exception as reason:  # noqa: BLE001
+        logging.warning("Parsing search query for %s failed: %s", domain_object_type, reason)
+        return ParseError(domain_object_type=domain_object_type, error=str(reason), ok=False)
+    try:
+        reports = latest_reports(database)
+        if domain_object_type == "report":
+            domain_objects: list[Report] | Iterator[Report] = reports
+        else:
+            domain_objects = {"metric": metrics, "source": sources, "subject": subjects}[domain_object_type](*reports)
```
wkoot, line 92: Bit of a magic one liner

components/api_server/tests/routes/test_search.py
```diff
@@ -0,0 +1,138 @@
...
+            {"subjects": {SUBJECT_ID2: {"metrics": {METRIC_ID2: {"name": "Metric 2"}}}}},
+        ]
+        expected_results = {
+            "domain_object_type": "metric",
+            "ok": True,
+            "search_query": {"name": "Metric 1"},
+            "uuids": [METRIC_ID],
+        }
+        self.assertEqual(expected_results, search("metric", self.database))
+
```
wkoot, line 71: We should also check that we can find more than one uuid, when there's multiple objects matching

components/api_server/tests/routes/test_search.py
```diff
@@ -0,0 +1,138 @@
...
+        """Test that a source can be found by parameter."""
+        self.database.reports.find.return_value = [
+            {
+                "subjects": {
+                    SUBJECT_ID: {
+                        "metrics": {METRIC_ID: {"sources": {SOURCE_ID: {"parameters": {"url": "https://example.org"}}}}}
+                    }
+                }
+            },
+            {"subjects": {SUBJECT_ID2: {"metrics": {METRIC_ID2: {"sources": {SOURCE_ID2: {"parameters": {}}}}}}}},
```
wkoot: Perhaps also include the parameter url in source2, but with unmatching value

docs/src/software.md
```diff
@@ -69,7 +69,7 @@ The frontend uses the following environment variables:
 
 The API of the API-server is versioned. The version is not changed when backwards compatible changes are made, such as the addition of new endpoints.
```
wkoot:
```suggestion
```

tests/feature_tests/src/environment.py
```diff
@@ -6,6 +6,7 @@
 from behave.runner import Context
 
 JSON_CONTENT_TYPE = "application/json"
+JSON = dict | list
```
wkoot, line 9: Bit of a generic name

tests/feature_tests/src/features/search.feature
```diff
@@ -0,0 +1,31 @@
...
+    When the client searches a report without query parameters
+    Then the search results contain an error message
+
+  Scenario: search source
+    Given an existing subject
+    And an existing metric
+    And an existing source with name "Source 1" and parameter url "https://source1"
+    And an existing source with name "Source 2" and parameter url "https://source2"
+    When the client searches a source with url "https://source2"
+    Then the search results contain the uuid of the source
```
wkoot, line 31: I guess it's acceptable that we don't check the value, as this is already covered by unittests. Feels weird to read the testcase like this though, perhaps we can assert that there is only one uuid in the result?

tests/feature_tests/src/steps/search.py
```diff
@@ -0,0 +1,37 @@
...
+@when("the client searches a report without query parameters")
+def search_invalid(context: Context) -> None:
+    """Search with invalid query parameters."""
+    context.post("report/search", json={})
+
+
+@then("the search results contain the uuid of the {domain_object_type}")
+def check_search_results(context: Context, domain_object_type: str) -> None:
+    """Check the search results."""
+    assert_in(context.uuid[domain_object_type], context.response.json()["uuids"])
```
wkoot, line 23: Hmmmmm, but what's happening when there are multiple objects of the same type? This looks like a mutable global

docs/src/api.md
```diff
@@ -0,0 +1,201 @@
...
+
+For example:
+
+```json
+{
+    "title": "My application"
+}
+```
+
+The search endpoints can search for exactly one attribute at a time. The matching is done on the complete value and is case sensitive in case of strings.
```
wkoot: Aren't all parameters always strings?

docs/src/api.md
```diff
@@ -0,0 +1,201 @@
...
+    "domain_object_type": "metric",
+    "error": "error message",
+    "search_query": {"name": "Metric 1"},
+    "ok": false
+}
+```
+
+The `search_query` is not included in the error response if parsing the search query parameters fails.
+
+## Export to PDF
```
wkoot, line 69: I haven't checked the rest of this file, assuming it's all moved from usage.md

components/api_server/src/routes/search.py
fniessink: Ah yes, I looked into that, but TypedDict doesn't allow for default values. See https://stackoverflow.com/questions/76251837/default-values-for-typeddict (and the link to https://github.com/pydantic/pydantic/discussions/6517).

components/api_server/src/routes/search.py
fniessink: I've removed the duplication.

components/api_server/src/routes/search.py
fniessink: Because then we need a sentinel value or we can't search for values that are None.

tests/feature_tests/src/features/search.feature
fniessink: We do check that the domain object added last is in the search results:

```python
@then("the search results contain the uuid of the {domain_object_type}")
def check_search_results(context: Context, domain_object_type: str) -> None:
    """Check the search results."""
    assert_in(context.uuid[domain_object_type], context.response.json()["uuids"])
```

tests/feature_tests/src/steps/search.py
fniessink: This is how the feature tests work: context.uuid["report"] always contains the uuid of the report last created.

PR #7691, opened 2023-12-18
Allow for hiding report, subject and tag cards from dashboards via th…

### LINKED ISSUE #7482: Make it possible to hide all subject cards and/or all tag cards
In the settings panel add a column: "Visible cards" with values "Subjects" and "Tags".

Either group can be hidden.

This setting does not change which subjects and/or metrics are visible.

fniessink:
…e settings panel. Closes #7482.

### DIFF
--- a/components/frontend/src/header_footer/SettingsPanel.js
+++ b/components/frontend/src/header_footer/SettingsPanel.js
@@ -4,6 +4,7 @@ import { Button, Grid, Header, Menu, Segment } from 'semantic-ui-react';
 import { Icon } from '../semantic_ui_react_wrappers';
 import {
     boolURLSearchQueryPropType,
+    hiddenCardsPropType,
     integerURLSearchQueryPropType,
     metricsToHidePropType,
     metricsToHideURLSearchQueryPropType,
@@ -35,6 +36,7 @@ export function SettingsPanel({
         disabled: oneDateColumn,
         help: "The date interval can only be changed when at least two dates are shown",
     }
+    const cardsMenuItemProps = { hiddenCards: settings.hiddenCards }
     const metricMenuItemProps = { metricsToHide: settings.metricsToHide }
     const sortColumnMenuItemProps = { sortColumn: settings.sortColumn, sortDirection: settings.sortDirection, handleSort: handleSort }
     const sortOrderMenuItemProps = {
@@ -63,6 +65,13 @@ export function SettingsPanel({
                         </Grid.Column>
                     </Grid.Row>
                 </Grid>
+            </Segment>
+            <Segment inverted color="black">
+                <Header size='small'>Visible cards</Header>
+                <Menu {...menuProps}>
+                    <VisibleCardsMenuItem cards={atReportsOverview ? "reports" : "subjects"} {...cardsMenuItemProps} />
+                    <VisibleCardsMenuItem cards="tags" {...cardsMenuItemProps} />
+                </Menu>
             </Segment>
             <Segment inverted color="black">
                 <Header size='small'>Visible metrics</Header>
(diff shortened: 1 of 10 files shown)


components/frontend/src/header_footer/SettingsPanel.js
```diff
@@ -64,6 +66,13 @@ export function SettingsPanel({
                     </Grid.Row>
                 </Grid>
             </Segment>
+            <Segment inverted color="black">
+                <Header size='small'>Visible cards</Header>
+                <Menu {...menuProps}>
+                        <VisibleCardsMenuItem cards={atReportsOverview ? "reports" : "subjects"} {...cardsMenuItemProps} />
```
wkoot: Are there no linting rules for indentation?

components/frontend/src/header_footer/SettingsPanel.js
```diff
@@ -247,6 +256,22 @@ SettingsPanel.propTypes = {
     ...ResetSettingsButton.propTypes,
 }
 
+function VisibleCardsMenuItem({ cards, hiddenCards}) {
```
wkoot:
```suggestion
function VisibleCardsMenuItem({ cards, hiddenCards }) {
```

components/frontend/src/header_footer/SettingsPanel.js
```diff
@@ -247,6 +256,22 @@ SettingsPanel.propTypes = {
...
+    return (
+        <SettingsMenuItem
+            active={!hiddenCards.includes(cards)}
+            onClick={hiddenCards.toggle}
```
wkoot, line 263: The `.toggle` suggests that this prop is a boolean, but `.includes` suggests otherwise. I don't really understand how this toggle is wired

components/frontend/src/header_footer/SettingsPanel.test.js
```diff
@@ -204,69 +173,78 @@ it("ignores a keypress if the menu item is disabled", async () => {
...
+    expect(history.location.search).toBe("?nr_dates=2&date_interval=1")
 })
 
-it("sorts the dates descending", async () => {
-    history.push("?nr_dates=2")
-    const dateOrder = renderHook(() => useDateOrderURLSearchQuery())
-    renderSettingsPanel({ dateOrder: dateOrder.result.current })
-    await act(async () => fireEvent.click(screen.getByText(/Descending/)))
-    expect(dateOrder.result.current.value).toBe("descending")
+it("sorts the dates descending", () => {
```
wkoot, line 208: Why are many tests async, but some not? I don't really see the pattern, maybe there are more async than needed?

components/frontend/src/report/ReportDashboard.js
```diff
@@ -47,40 +47,45 @@ export function ReportDashboard(
...
+        })
+    }
+    let tagCards = [];
+    if (!settings.hiddenCards.includes("tags")) {
+        const anyTagsHidden = settings.hiddenTags.value.length > 0
+        tagCards = getReportTags(report, settings.hiddenTags.value).map((tag) => {
             const summary = {}
             dates.forEach((date) => {
-                summary[date] = summarizeMetricsOnDate(metrics, measurements, date)
+                summary[date] = summarizeTagOnDate(report, measurements, tag, date)
```
wkoot, line 76: Why is `Tag` not plural, like `Metrics`, when the check is `hiddenCards.includes("tags")` ?

components/frontend/src/report/ReportDashboard.js
```diff
@@ -47,40 +47,45 @@ export function ReportDashboard(
...
             })
-            subjectCards.push(
+            return (
```
wkoot, line 78: Not sure why this part is so different compared to subjects/metrics, but looks like it works :D

components/frontend/src/header_footer/SettingsPanel.js
fniessink: Apparently not!

fniessink: I'll make a separate ticket to properly configure `eslint`.

components/frontend/src/header_footer/SettingsPanel.js
fniessink: The toggle adds or removes one specific item to the list. Which one is specified by `onClickData` on the next line.

components/frontend/src/report/ReportDashboard.js
fniessink: Because the metrics for one tag are summarised. The tag card shows the status of possibly multiple metrics (that all have that one tag) on the specified date.

components/frontend/src/report/ReportDashboard.js
fniessink: Yeah, I briefly looked into merging ReportsOverviewDashboard and ReportDashboard, but decided against it, at least as part of this issue.

components/frontend/src/header_footer/SettingsPanel.test.js
fniessink: I think it mostly depends on whether there's an useEffect that fires after the click. Many tests will work without async, but then the logging is full of messages that clicking events should be covered in act() calls.

PR #7705, opened 2023-12-20
Add the possibility to include SonarQube violations by tag…

### LINKED ISSUE #6967: Add possibility to filter issues by tags in SonarQube
Tasks

- [ ] Add a "Tags to include (regular expressions or tags)" parameter to SonarQube sources. When the tags parameter has one or more tags, the collector only includes issues that have one or more of the listed tags. Help text: "Tags to include can be specified by tag or by regular expression.
- [ ] Add a tags column to the measurement entity tables

Out of scope:
- Getting the available tags from SonarQube and providing them to the user as dropdown.
- Excluding tags

Rationale:

Currently, when using the OWASP Dependency check plugin in SonarQube, you can't differentiate the dependency vulnerabilities from vulnerabilities found in the code. However, it looks like the OWASP Dependency check plugin adds the tag 'cve' to the found vulnerable dependencies.

I'd like to be able to add tag names in Quality-Time so I can add an extra filter and only see specific vulnerabilities from SonarQube, like in this case the dependencies.

fniessink:
…. Closes #6967.

### DIFF
--- a/components/collector/tests/source_collectors/sonarqube/test_security_warnings.py
+++ b/components/collector/tests/source_collectors/sonarqube/test_security_warnings.py
@@ -7,6 +7,15 @@ class SonarQubeSecurityWarningsTest(SonarQubeTestCase):
     """Unit tests for the SonarQube security warnings collector."""
 
     METRIC_TYPE = "security_warnings"
+    SONARQUBE_URL = "https://sonarqube"
+    API_URL = f"{SONARQUBE_URL}/api"
+    LANDING_URL = f"{SONARQUBE_URL}/project"
+    BRANCH = "&branch=master"
+    DASHBOARD_URL = f"{SONARQUBE_URL}/dashboard?id=id{BRANCH}"
+    HOTSPOTS_API = f"{API_URL}/hotspots/search?projectKey=id{BRANCH}&ps=500"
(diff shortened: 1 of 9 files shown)


components/collector/tests/source_collectors/sonarqube/test_security_warnings.py
```diff
@@ -151,3 +155,15 @@ async def test_filter_security_warnings_hotspots_by_status(self):
             entities=self.hotspot_entities,
             landing_url="https://sonarqube/project/security_hotspots?id=id&branch=master",
         )
+
+    async def test_filter_security_warning_vulnerabilities_by_tag(self):
+        """Test that the security warning vulnerabilities can be filtered by tag."""
+        self.set_source_parameter("tags", ["bug"])
```
wkoot, line 181: Would it not make more sense to include by "other tag", to verify it only includes one?

fniessink: Yes, if the collector did the filtering, but in this case the filtering is done by SonarQube. We just pass the query parameter. I'll add an assert that the GET method is called with the correct query.

fniessink: And extend the other tests as well.

PR #7800, opened 2024-01-05
The violations metric can now also be used to measure accessibility v…

### LINKED ISSUE #562: Merge the accessibility violations metric with violations metric
Tasks:
- [x] Add the accessibility violation metric sources to the violations metric.
- [x] Automatically migrate the accessibility violation metric to the violation metric.
- [x] Give the old accessibility metrics a custom name "Accessibility violations" and a unit "accessibility violations" if the metric name and unit haven't been overridden before.
- [x] Update data model, including docs.

fniessink:
…iolations. Existing accessibility violation metrics have been changed to the violations metrics. Closes #562.

### DIFF
--- a/components/api_server/src/initialization/database.py
+++ b/components/api_server/src/initialization/database.py
@@ -26,6 +26,7 @@ def init_database() -> Database:  # pragma: no feature-test-cover
     import_datamodel(database)
     initialize_secrets(database)
     initialize_reports_overview(database)
+    perform_migrations(database)
     if os.environ.get("LOAD_EXAMPLE_REPORTS", "True").lower() == "true":
         import_example_reports(database)
     return database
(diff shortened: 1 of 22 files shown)


components/shared_code/src/shared_data_model/metrics.py
```diff
@@ -715,15 +701,18 @@
     ),
     "violations": Metric(
         name="Violations",
-        description="The number of violations of programming rules in the software.",
-        rationale="The more programming rules are violated, the harder it may be to understand and maintain the "
-        "software.",
+        description="The number of violations of rules in the software.",
+        rationale="""Depending on what kind of rules or guideliens the source checks, violations may pose a risk for
```
wkoot, line 705: Do we not also have "security violations"? What kind of input do we get in SARIF?

components/api_server/src/initialization/database.py
```diff
@@ -50,3 +53,33 @@ def create_indexes(database: Database) -> None:
...
+        if changed:
+            report_id = report["_id"]
+            del report["_id"]
+            database.reports.replace_one({"_id": report_id}, report)
+
+
+def change_accessibility_violations_metric_to_violations(metric: dict) -> None:  # pragma: no cover-behave
+    """Change the accessibility violations metric to violations metric."""
+    metric["type"] = "violations"
+    if not metric.get("name"):
```
wkoot, line 85: Why do we set name and unit when it is unset, shouldn't it default to a value from the data model?

components/shared_code/src/shared_data_model/metrics.py
fniessink: The metric you're thinking of is `security_warnings` I think. SARIF does indeed support both the `violations` and the `security_warnings` metric types. But since the `violations` metric itself is not changed here, the SARIF source does not need to be changed.

components/api_server/src/initialization/database.py
fniessink: If the user has not changed the accessibility metric's name or unit they currently have the (default) name "Accessibility violations" and the (default) unit "accessibility violations". The migration code changes the name and unit of these metrics so as to minimise the change from the user's perspective. Without that, the accessibility metrics would be called "Violations" and have "violations" as unit after the migration.

components/shared_code/src/shared_data_model/metrics.py
wkoot: Should these security warnings not be explicitly mentioned in the rationale?

components/api_server/src/initialization/database.py
wkoot: Ah, understandable. I thought it would make sense to have the name and unit automatically migrated :)

components/shared_code/src/shared_data_model/metrics.py
fniessink: Why would we mention security warnings in the rationale for violations? We don't mention all the other metric types either?

PR #7892, opened 2024-01-19
When measuring security warnings with Cargo audit JSON reports as sou…

### LINKED ISSUE #7889: Add Cargo audit ouput support for warnings and yanked
Add Cargo audit ouput support for "warnings" and "yanked" in addition to "vulnerabilities". See attached json for details.

**Additional context**
Add an extra column in Quality-Time to indicate the type of security warning (warnings, yanked or vulnerabilities). At the moment this is not necessary because only vulnerabilities are shown. 
Optional, make the types you want to see configurable in QT.
[cargo-audit-anonymized.json](https://github.com/ICTU/quality-time/files/13987476/cargo-audit-anonymized.json)

fniessink:
…rce, also count warnings (unsound and yanked packages) in addition to vulnerabilities. Closes #7889.


components/collector/src/source_collectors/cargo_audit/security_warnings.py
```diff
@@ -1,32 +1,78 @@
...
+    advisory: Advisory
+    package: Package
+    versions: dict[Literal["patched"], list[str]]
+    kind: Literal["vulnerability", "unsound", "yanked"]
+
+
+class CargoAuditJSON(TypedDict):
+    """Cargo audit JSON."""
+
+    vulnerabilities: dict[Literal["list"], list[CargoAuditWarning]]
```
wkoot: Weird that the source data is structured this way

components/collector/src/source_collectors/cargo_audit/security_warnings.py
```diff
@@ -1,32 +1,78 @@
...
         entities = Entities()
-        vulnerabilities = cast(JSONDict, json).get("vulnerabilities", {}).get("list", [])
-        for vulnerability in vulnerabilities:
-            package_name = vulnerability["package"]["name"]
-            package_version = vulnerability["package"]["version"]
-            advisory_id = vulnerability["advisory"]["id"]
+        for warning in self.__warnings(json):
+            package_name = warning["package"]["name"]
+            package_version = warning["package"]["version"]
+            advisory = warning.get("advisory", {}) or {}
```
wkoot: `, {}` seems useless when using `or {}` anyway

components/collector/src/source_collectors/cargo_audit/security_warnings.py
```diff
@@ -1,32 +1,78 @@
...
+                    versions_patched=", ".join(versions.get("patched", [])),
+                    warning_type=warning.get("kind", "vulnerability"),
                 ),
             )
         return entities
+
+    def __warnings(self, json: JSON) -> list[CargoAuditWarning]:
+        """Collect all warnings into one list for easier processing."""
+        json_dict = cast(CargoAuditJSON, json)
+        warnings = json_dict.get("vulnerabilities", {}).get("list", [])
```
wkoot: Why use `.get()`, shouldn't the fact that it is an instance of `CargoAuditJSON` already mean that these values exist? Or does it secretly classify fields as `Optional`?

components/collector/src/source_collectors/cargo_audit/security_warnings.py
```diff
@@ -1,32 +1,78 @@
...
+        warnings_dict = json_dict.get("warnings", {})
+        for warnings_list in warnings_dict.values():
+            warnings.extend(warnings_list)
+        return warnings
+
+    def _include_entity(self, entity: Entity) -> bool:
+        """Return whether to include the warning in the measurement."""
+        return entity["warning_type"] in self._parameter("warning_types")
```
wkoot, line 90: Perhaps it's a bit confusing to use the name `warning` here, because it applies to both vulnerabilities and warnings?

components/collector/src/source_collectors/cargo_audit/security_warnings.py
fniessink: Yeah, can't do much about that. At least the TypedDicts make it explicit.

components/collector/src/source_collectors/cargo_audit/security_warnings.py
fniessink: Good catch.

components/collector/src/source_collectors/cargo_audit/security_warnings.py
fniessink: The data is not validated against the TypedDict like when using Pydantic. It's only type hints. And there's no schema for the Cargo Audit JSON so I don't know for sure these fields are always present.

components/collector/src/source_collectors/cargo_audit/security_warnings.py
fniessink: Yeah, but the metric is "security warnings", so that's why the parameter is called warning types. I'll see if I can make it less confusing.

PR #8043, opened 2024-02-06
Add a new metric 'CI-pipeline duration'. Partially implements #6423.

### LINKED ISSUE #6423: Duration of pipelines
# User story

As a developer I want to track the duration of my pipeline so that I prevent it from taking too much time.

## Acceptance criteria

- [x] User can add a metric "CI-pipeline duration" 
- [x] Users can add Jenkins as source.
  - [x] User can select a pipeline and branch to monitor (making the metric working for jobs is nice to have). 
- [x] Users can add GitLab CI as source.
   - [x] User can select a pipeline and branch to monitor.
- [x] Users can add Azure DevOps as source.
   - [x] User can select a pipeline and branch to monitor.

## Scope

- [x] The metric can be used to measure the duration of exactly one pipeline.

## Out of scope

- [ ] Quality-time measures the duration of the complete pipeline and shows the stages as measurement entities, with the duration per stage.
- [ ] Users can ignore/include stages by name.
- [ ] #12180
- [x] #12181

# Tasks

### Shared code

- [x] Add metric `CI-pipeline duration` to the data model
- [x] Change `Jenkins` to support the new metric 
- [x] Change `Azure DevOps` to support the new metric 
- [x] Change `GitLab CI` to support the new metric 

### Collector

- [x] Add JenkinsPipelineDuration collector. For Jenkins the [`wfapi`](https://github.com/jenkinsci/pipeline-stage-view-plugin/blob/master/rest-api/README.md) can be used.
- [x] Add AzureDevopsPipelineDuration collector
- [x] Add GitLabPipelineDuration collector

## Documentation

- [x] Update changelog

fniessink:


components/collector/src/source_collectors/gitlab/source_up_to_dateness.py
```diff
@@ -80,7 +80,8 @@ class GitLabPipelineUpToDateness(TimePassedCollector, GitLabPipelineBase):
 
     async def _parse_source_response_date_time(self, response: Response) -> datetime:
         """Override to get the date and time of the pipeline."""
-        datetimes = [self._datetime(pipeline) for pipeline in await self._pipelines(response)]
+        pipelines = await self._pipelines(SourceResponses(responses=[response]))
+        datetimes = [pipeline.datetime for pipeline in pipelines]
```
wkoot, line 84: Is the `._datetime()` then still in use?

components/shared_code/src/shared_data_model/sources/quality_time.py
```diff
@@ -75,6 +75,7 @@
             values=[
                 "Average issue lead time",
                 "Change failure rate",
+                "CI-pipeline duration",
```
wkoot, line 78: How do we distinguish between "CI-pipeline" and "CI-jobs" ?

components/shared_code/src/shared_data_model/parameters.py
```diff
@@ -151,6 +151,15 @@ class Branch(StringParameter):
     metrics: list[str] = ["source_up_to_dateness"]
 
 
+class Branches(MultipleChoiceWithAdditionParameter):
+    """Branches parameter."""
```
wkoot, line 155: Is the plan to include this parameter in the other sources (jenkins, sonar, devops, etc) as well?

components/shared_code/src/shared_data_model/parameters.py
```diff
@@ -151,6 +151,15 @@ class Branch(StringParameter):
...
+
+    name: str = "Branches (regular expressions or branch names)"
```
wkoot, line 157: Does this mean it is an _inclusive_ filter parameter?

wkoot: What about this `TargetBranchesToInclude` on line 171?

components/collector/tests/source_collectors/gitlab/test_pipeline_duration.py
```diff
@@ -0,0 +1,57 @@
...
+
+    METRIC_TYPE = "pipeline_duration"
+    METRIC_ADDITION = "min"
+
+    def setUp(self):
+        """Extend to set up fixtures."""
+        super().setUp()
+        self.landing_url = "https://gitlab/project/-/pipelines/1"
+        self.pipeline_json = [
+            {
```
wkoot, line 17: Shouldn't we also test that the highest duration is included, when there are multiple pipeline runs?

components/collector/tests/source_collectors/gitlab/test_pipeline_duration.py
```diff
@@ -0,0 +1,57 @@
...
+                self.pipeline_detail_json,  # To get the pipeline landing URL for the most recent pipeline
+            ]
+        )
+        self.assert_measurement(response, value="10", landing_url=self.landing_url)
+
+    async def test_duration_when_no_match(self):
+        """Test that an error is returned when no pipelines match."""
+        self.set_source_parameter("branches", "missing")
+        response = await self.collect(get_request_json_return_value=self.pipeline_json)
+        self.assert_measurement(response, connection_error="IndexError: list index out of range")
```
wkoot: This is a kindof nondescriptive error, but I guess we usually don't format nice error hints?

components/collector/src/source_collectors/gitlab/pipeline_duration.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class GitLabPipelineDuration(GitLabPipelineBase):
+    """GitLab CI-pipeline duration collector."""
+
+    async def _get_source_responses(self, *urls: URL) -> SourceResponses:
+        """Override to get the pipeline details."""
+        responses = await super()._get_source_responses(*urls)
+        pipelines = await self._pipelines(responses)
+        api_url = await self._gitlab_api_url(f"pipelines/{pipelines[0].id}")
```
wkoot, line 16: Doesn't the high default `lookback_days` mean that this will be a very slow KPI? Should we protect users against configuring something dumb?

components/collector/src/source_collectors/gitlab/pipeline_duration.py
```diff
@@ -0,0 +1,22 @@
...
+        return await super()._get_source_responses(api_url)
```
wkoot, line 17: Do I understand it correctly to summarise that we "get all pipelines", "select the first" and then "return that pipeline data"? How then do we select the highest from X recent pipeline runs?

wkoot: And does this logic not conflict with other Gitlab pipeline metrics we offer, which consider individual stages/jobs?

components/collector/src/source_collectors/gitlab/base.py
```diff
@@ -123,15 +124,31 @@ def _build_date(job: Job) -> date:
         return parse_datetime(job.get("finished_at") or job["created_at"]).date()
 
 
-class Pipeline(TypedDict):
-    """GitLab pipeline JSON."""
+@dataclass
```
wkoot, line 128: Yay for dataclass

components/collector/src/source_collectors/gitlab/base.py
```diff
@@ -143,29 +160,28 @@ async def _api_url(self) -> URL:
...
-            and pipeline["status"] in self._parameter("pipeline_statuses_to_include")
-            and pipeline["source"] in self._parameter("pipeline_triggers_to_include")
-        )
-
-    @staticmethod
-    def _datetime(pipeline: Pipeline) -> datetime:
-        """Return the datetime of the pipeline."""
-        return parse_datetime(pipeline.get("updated_at") or pipeline["created_at"])
+        branches = self._parameter("branches")
+        matches_branches = match_string_or_regular_expression(pipeline.ref, branches) if branches else True
```
wkoot, line 187: Was there no exclude branch option?

components/collector/src/source_collectors/gitlab/source_up_to_dateness.py
fniessink: No, and it's removed.

components/shared_code/src/shared_data_model/sources/quality_time.py
fniessink: According to the user story we're mostly interested in pipeline duration, and only in the case of Jenkins measuring the duration of the jobs is mentioned. And then only as a nice to have. So I think the name is fine for now.

components/shared_code/src/shared_data_model/parameters.py
fniessink: It depends on the metric. Some metrics (still) need exactly one branch. But it may be a good idea to look at harmonising this further. I have added a story to the backlog to change the default branch name of the `Branch` parameter to `main`. I'll add a note there to investigate whether the `Branch` and `Branches` parameters can be combined.

components/shared_code/src/shared_data_model/parameters.py
fniessink: I'm not sure what you mean with inclusive filter (if a filter includes something it also excludes things, right?). Anyway, if the user does not add any branches to the `Branches` the collector considers pipelines on all branches (as signalled by the placeholder being `all branches`. As soon as the user adds one or more branches, the collector only looks at pipelines for those branches.

Yeah,`TargetBranchesToInclude` is also a `MultipleChoiceWithAdditionParameter` parameter for branches, but all attributes are different, so I don't see a way to get rid of it.

components/collector/tests/source_collectors/gitlab/test_pipeline_duration.py
fniessink: The metric does not measure the slowest pipeline, but the duration of the latest pipeline, so no.

components/collector/tests/source_collectors/gitlab/test_pipeline_duration.py
fniessink: Yes, you're right. We had a better error message for the source up-to-dateness of GitLab pipelines already. I've pulled that functionality up to `GitLabPipelineBase`.

components/collector/src/source_collectors/gitlab/pipeline_duration.py
fniessink: Yeah, good thinking. Unfortunately the GitLab `lookback_days` parameter is also used for other metrics (`change_failure_rate`, `failed_jobs`, `job_runs_within_time_period`, `source_up_to_dateness`, `unused_jobs`) where it does make sense to have a higher default value.

The Azure DevOps collector does have two separate loopback parameters, one for pipelines and one for issues. We should make the same distinction for GitLab. I'll add an tech issue to the backlog.

fniessink: Strike that. I'll change it right away. As he pipeline duration Is a new metric I can just as well introduce a new parameter now.

components/collector/src/source_collectors/gitlab/pipeline_duration.py
fniessink: As mentioned above, this metric measures the duration of the latest pipeline.

components/collector/src/source_collectors/gitlab/base.py
fniessink: Nope. Let's wait to see if users need it.

PR #8075, opened 2024-02-13
When measuring the size of source code, also report the relative size…

### LINKED ISSUE #7247: See percentage in metric 'Size (LOC)'
I'm always frustrated when I need to report the composition of code and I can only see the number of code lines.

In the entities tab, I would like the numbers per language also show the percentage of the total.
Additionally, I would like Quality-time to use a period or a space as a thousands separator to make it more readable. 
Read [this](https://en.wikipedia.org/wiki/Decimal_separator#:~:text=Space%2C%20the%20internationally%20recommended,in%20most%20English%2Dspeaking%20countries.) before implementing.

So, this will result in the following. 
total lines of code = 2.200 lines
CSS = 220 lines (10%)
Java = 1.540 lines (70%)
JavaScript = 440 lines (20%)

**Additional context**
image

fniessink:
… of each programming language in the measurement details. Closes #7247.

### DIFF
--- a/components/collector/src/source_collectors/sonarqube/loc.py
+++ b/components/collector/src/source_collectors/sonarqube/loc.py
@@ -19,13 +19,16 @@ class SonarQubeLOC(SonarQubeMetricsBaseClass):
         "cpp": "C++",
         "cobol": "COBOL",
         "css": "CSS",
+        "docker": "Docker",
         "flex": "Flex",
         "go": "Go",
         "web": "HTML",
+        "json": "JSON",
         "jsp": "JSP",
         "java": "Java",
         "js": "JavaScript",
         "kotlin": "Kotlin",
+        "kubernetes": "Kubernetes",
         "objc": "Objective-C",
         "php": "PHP",
         "plsql": "PL/SQL",
@@ -33,10 +36,12 @@ class SonarQubeLOC(SonarQubeMetricsBaseClass):
         "ruby": "Ruby",
         "scala": "Scala",
         "swift": "Swift",
-        "tsql": "T-SQL",
+        "text": "Text",
         "ts": "TypeScript",
+        "tsql": "T-SQL",
         "vbnet": "VB.NET",
         "xml": "XML",
+        "yaml": "YAML",
     }  # https://sonarcloud.io/api/languages/list
 
     def _value_key(self) -> str:
@@ -55,7 +60,7 @@ class SonarQubeLOC(SonarQubeMetricsBaseClass):
         if self._value_key() == "ncloc":
             # Our user picked non-commented lines of code (ncloc), so we can sum the ncloc per language, skipping
             # languages the user wants to ignore
-            return str(sum(int(ncloc) for _, ncloc in self.__language_ncloc(metrics)))
+            return str(self.__total(metrics))
         return super()._value(metrics)
 
     async def _entities(self, metrics: dict[str, str]) -> Entities:
@@ -63,18 +68,28 @@ class SonarQubeLOC(SonarQubeMetricsBaseClass):
         if self._value_key() == "ncloc":
             # Our user picked non-commented lines of code (ncloc), so we can show the ncloc per language, skipping
             # languages the user wants to ignore
+            total = self.__total(metrics)
             return Entities(
-                Entity(key=language, language=self.LANGUAGES.get(language, language), ncloc=ncloc)
+                Entity(
+                    key=language,
+                    language=self.LANGUAGES.get(language, language),
+                    ncloc=str(ncloc),
+                    ncloc_percentage=str(round(int(ncloc) / total * 100)),
+                )
                 for language, ncloc in self.__language_ncloc(metrics)
             )
         return await super()._entities(metrics)
 
-    def __language_ncloc(self, metrics: dict[str, str]) -> list[list[str]]:
+    def __language_ncloc(self, metrics: dict[str, str]) -> list[tuple[str, int]]:
         """Return the languages and non-commented lines of code per language, ignoring languages if so specified."""
         languages_to_ignore = [language.lower() for language in self._parameter("languages_to_ignore")]
         keys_to_ignore = [key for key, language in self.LANGUAGES.items() if language.lower() in languages_to_ignore]
         return [
-            language_count.split("=")
+            (language_count.split("=")[0], int(language_count.split("=")[1]))
             for language_count in metrics["ncloc_language_distribution"].split(";")
             if not match_string_or_regular_expression(language_count.split("=")[0], keys_to_ignore)
         ]
+
+    def __total(self, metrics: dict[str, str]) -> int:
+        """Return the total LOC for the selected languages."""
+        return sum(ncloc for _, ncloc in self.__language_ncloc(metrics))
--- a/components/collector/tests/source_collectors/cloc/test_loc.py
+++ b/components/collector/tests/source_collectors/cloc/test_loc.py
@@ -25,7 +25,15 @@ class ClocLOCTest(SourceCollectorTestCase):
             "production_file": {"blank": 2, "comment": 0, "code": 30, "language": "JavaScript"},
         }
         self.expected_entities = [
-            {"key": "Python", "language": "Python", "nr_files": "1", "blank": "5", "comment": "10", "code": "60"},
+            {
+                "key": "Python",
+                "language": "Python",
+                "nr_files": "1",
+                "blank": "5",
+                "comment": "10",
+                "code": "60",
+                "code_percentage": "67",  # round(60 / (60 + 30))
+            },
             {
                 "key": "JavaScript",
                 "language": "JavaScript",
@@ -33,6 +41,7 @@ class ClocLOCTest(SourceCollectorTestCase):
                 "blank": "2",
                 "comment": "0",
                 "code": "30",
+                "code_percentage": "33",  # round(30 / (60 + 30))
             },
         ]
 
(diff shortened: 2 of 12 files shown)


components/collector/src/source_collectors/sonarqube/loc.py
```diff
@@ -78,3 +89,7 @@ def __language_ncloc(self, metrics: dict[str, str]) -> list[list[str]]:
             for language_count in metrics["ncloc_language_distribution"].split(";")
             if not match_string_or_regular_expression(language_count.split("=")[0], keys_to_ignore)
         ]
+
+    def __total(self, metrics: dict[str, str]) -> int:
+        """Return the total LOC for the selected languages."""
```
wkoot, line 94: But in cloc we `Count the total LOC for all files`, doesn't it then make sense to do the same for sonar?

components/collector/src/source_collectors/sonarqube/loc.py
```diff
@@ -78,3 +89,7 @@ def __language_ncloc(self, metrics: dict[str, str]) -> list[list[str]]:
...
+        return sum(int(ncloc) for _, ncloc in self.__language_ncloc(metrics))
```
wkoot: Why not cast the `ncloc` returned by `__language_ncloc` to int, before returning?

components/collector/tests/source_collectors/cloc/test_loc.py
```diff
@@ -25,14 +25,23 @@ def setUp(self):
...
         self.expected_entities = [
-            {"key": "Python", "language": "Python", "nr_files": "1", "blank": "5", "comment": "10", "code": "60"},
+            {
+                "key": "Python",
+                "language": "Python",
+                "nr_files": "1",
+                "blank": "5",
+                "comment": "10",
+                "code": "60",
+                "code_percentage": "67",
```
wkoot: Perhaps add inline comment to point out the percentage follows from `60/(60+30)`

components/collector/tests/source_collectors/cloc/test_loc.py
```diff
@@ -56,4 +67,6 @@ async def test_loc_by_file_include_files(self):
         """Test that the number of lines is returned and that the languages are returned as entities."""
         self.set_source_parameter("files_to_include", ["test.*"])
         response = await self.collect(get_request_json_return_value=self.cloc_by_file_json)
-        self.assert_measurement(response, value="60", total="90", entities=self.expected_entities[:1])
+        expected_entities = self.expected_entities[:1]
+        expected_entities[0]["code_percentage"] = "100"
+        self.assert_measurement(response, value="60", total="90", entities=expected_entities)
```
wkoot, line 72: Why is the total "90" instead of "60"?

components/frontend/src/source/SourceEntities.test.js
```diff
@@ -134,6 +138,15 @@ it('sorts the entities by integer', async () => {
     assertOrder(["B", "A", "C"])
 })
 
+it('sorts the entities by integer percentage', async () => {
```
wkoot, line 141: Perhaps use the numbers "1", "42" and "9", to make sure it doesn't secretly sort alnum

components/shared_code/src/shared_data_model/meta/entity.py
```diff
@@ -27,6 +27,7 @@ class EntityAttributeType(StrEnum):
     DATETIME = "datetime"
     FLOAT = "float"
     INTEGER = "integer"
+    INTEGER_PERCENTAGE = "integer_percentage"
```
wkoot, line 30: Should we do any type checking on these attributes, i.e. `0 <= integer_percentage <= 100` ?

components/shared_code/src/shared_data_model/sources/sonarqube.py
```diff
@@ -424,7 +424,10 @@ def violation_entity_attributes(
             measured_attribute="ncloc",
             attributes=[
                 EntityAttribute(name="Language"),
-                EntityAttribute(name="Number of lines with code", key="ncloc", type=EntityAttributeType.INTEGER),
+                EntityAttribute(name="Number of lines", key="ncloc", type=EntityAttributeType.INTEGER),
+                EntityAttribute(
+                    name="Percentage of lines", key="ncloc_percentage", type=EntityAttributeType.INTEGER_PERCENTAGE
```
wkoot, line 429: Why the difference with cloc?

components/frontend/src/source/SourceEntityAttribute.js
```diff
@@ -5,10 +5,12 @@ import { TimeAgoWithDate } from '../widgets/TimeAgoWithDate';
 
 export function SourceEntityAttribute({ entity, entity_attribute }) {
     let cell_contents = entity[entity_attribute.key] ?? "";
-    cell_contents = cell_contents && entity_attribute.type === "datetime" ? <TimeAgoWithDate dateFirst date={cell_contents} /> : cell_contents;
-    cell_contents = cell_contents && entity_attribute.type === "date" ? <TimeAgoWithDate dateFirst noTime date={cell_contents} /> : cell_contents;
-    cell_contents = cell_contents && entity_attribute.type === "integer" ? Math.round(cell_contents) : cell_contents;
-    cell_contents = cell_contents && entity_attribute.type === "status" ? <StatusIcon status={cell_contents} /> : cell_contents;
+    const type = entity_attribute.type ?? "";
```
wkoot, line 8: Is `type` not some protected name like in python?

components/frontend/src/source/SourceEntityAttribute.js
```diff
@@ -5,10 +5,12 @@ import { TimeAgoWithDate } from '../widgets/TimeAgoWithDate';
...
+    cell_contents = cell_contents && type === "datetime" ? <TimeAgoWithDate dateFirst date={cell_contents} /> : cell_contents;
+    cell_contents = cell_contents && type === "date" ? <TimeAgoWithDate dateFirst noTime date={cell_contents} /> : cell_contents;
+    cell_contents = cell_contents && type.includes("integer") ? Math.round(cell_contents).toString() : cell_contents;
```
wkoot, line 11: Why `Math.round`, can't we assume that we've been given an int?

components/frontend/src/source/SourceEntityAttribute.js
```diff
@@ -5,10 +5,12 @@ import { TimeAgoWithDate } from '../widgets/TimeAgoWithDate';
...
+    cell_contents = cell_contents && type.includes("percentage") ? cell_contents + "%" : cell_contents;
```
wkoot, line 12: Do lines 11 and 12 indeed both get triggered for type "integer_percentage" ? I'm not familiar with how the `&&` operator makes the object interact with itself here

components/collector/src/source_collectors/sonarqube/loc.py
fniessink: I'm not sure I understand. Sonar also gets the total for all files (it has no possibility to filter by filename currently).

components/collector/src/source_collectors/sonarqube/loc.py
fniessink: That would make sense, yes. Changed.

components/collector/src/source_collectors/sonarqube/loc.py
wkoot: Oh, I was mislead by the comment into thinking that it would only count the selection :)
I.e. if selected languages are python and javascript; `py/(py+js)`, but if there's a third language then it'd be omitted from the total instead of `py/(py+js+third)`

components/collector/tests/source_collectors/cloc/test_loc.py
fniessink: Filtered languages are removed from the total, filtered files are not. This is so that users can calculated the percentage of test code compared to the total number of loc.

components/shared_code/src/shared_data_model/meta/entity.py
fniessink: Nope. The percentage is provided by the collector, which should be a trusted source.

components/shared_code/src/shared_data_model/sources/sonarqube.py
fniessink: cloc can provide information per source file, the SonarQube metrics endpoint does not.

components/frontend/src/source/SourceEntityAttribute.js
fniessink: Nope, `typeof` is, but not `type`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords.

components/frontend/src/source/SourceEntityAttribute.js
fniessink: Yeah, but the `round` was already there and I'm not entirely sure all collectors do indeed round themselves.

components/frontend/src/source/SourceEntityAttribute.js
fniessink: Not sure what "interact with itself" means? Anyway, `&&` has a higher priority than `x ? y : z`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence

components/collector/src/source_collectors/sonarqube/loc.py
fniessink: But it is giving the total for the selected languages. The doc string is correct. 

Maybe you got confused by the fact that cloc can *also* filter by file. SonarQube doesn't allow for that. 

When filtering by file, the cloc collector includes the size of the filtered files in the total (but not the filtered languages), so the percentage of e.g. test files can be reported.

PR #8131, opened 2024-02-22
When showing measurements on multiple dates, also allow for having co…

### LINKED ISSUE #7039: Add delta column
# User story

As a quality manager I want to see the difference between measurements made at two or more dates.

## Acceptance criteria

- [ ] User can turn delta's on and off via the settings panel
- [ ] The delta setting is included in the URL parameter
- [ ] When showing multiple dates a delta column is added between each pair of dates, showing the delta between each pair of dates
- [ ] The delta column shows the measurement difference as:
  - [ ]  "+&lt;number&gt;" or "-&lt;number&gt;" in case of count and percentage scale (note that the percentage delta is not a percentage)
  - [ ]  "lower", "equal", "higher" in case of version numbers
- [ ] The delta column display the delta using a green label if the measurement is better and a red label if it's worse, no label if it's equal. 
- [ ] The delta column header is "Change"

## Nice to have

- [ ] When showing one date, a delta column is added, showing the delta with yesterday's latest measurements

## Assumptions

- No changes needed to the API.

## Open issues

- What do we need to ...

# Tasks

### Frontend

- [ ] Add columns
- [ ] Add setting

## Documentation

- [ ] Update changelog
- [ ] Update screenshots

fniessink:
…lumns that show the delta between dates. The delta columns can be turned on and off via the Settings panel. Closes #7039.

### DIFF
--- a/components/frontend/src/sharedPropTypes.js
+++ b/components/frontend/src/sharedPropTypes.js
@@ -126,13 +126,16 @@ export const subjectPropType = PropTypes.shape({
     type: PropTypes.string
 })
 
+export const scalePropType = PropTypes.oneOf(["count", "percentage", "version_number"])
+
 export const metricPropType = PropTypes.shape({
     accept_debt: PropTypes.bool,
     debt_end_date: PropTypes.string,
     evaluate_targets: PropTypes.bool,
     issue_ids: stringsPropType,
     issue_status: PropTypes.arrayOf(issueStatusPropType),
-    tags: stringsPropType
+    scale: scalePropType,
+    tags: stringsPropType,
 })
 
 export const metricsPropType = PropTypes.arrayOf(metricPropType)
--- a/components/frontend/src/subject/SubjectTableHeader.js
+++ b/components/frontend/src/subject/SubjectTableHeader.js
@@ -225,6 +225,31 @@ const tagsHelp = <>
     </p>
 </>
 
+function MeasurementHeaderCells({ columnDates, showDeltaColumns }) {
+    const cells = []
+    columnDates.forEach((date, index) => {
+        if (showDeltaColumns && index > 0) {
+            cells.push(
+                <UnsortableTableHeaderCell
+                    key={`delta-${date}`}
+                    help="The difference between the measurement values on the previous and next date.
+                    A plus (+) sign indicates that the newer value is higher, a minus (-) sign that it is lower.
+                    A green outline indicates that the newer value is better, a red outline that it is worse."
+                    label="𝚫"
+                    textAlign="right"
+                />
+            )
+        }
+        cells.push(<UnsortableTableHeaderCell key={date} textAlign="right" label={date.toLocaleDateString()} />)
+    })
+    return cells
+}
+MeasurementHeaderCells.propTypes = {
+    columnDates: datesPropType,
+    showDeltaColumns: PropTypes.bool,
+}
+
+
 export function SubjectTableHeader(
     {
         columnDates,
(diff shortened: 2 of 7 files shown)


components/frontend/src/sharedPropTypes.js
```diff
@@ -110,6 +110,8 @@ export const measurementSourceType = PropTypes.shape({
     parse_error: PropTypes.string,
 })
 
+export const measurementValueType = PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```
wkoot: Why can it be number, doesn't the api always return strings? Or am I confusing this with collector measurement tests?

components/frontend/src/subject/SubjectTableHeader.js
```diff
@@ -225,6 +225,31 @@ const tagsHelp = <>
...
+    const cells = []
+    columnDates.forEach((date, index) => {
+        if (showDeltaColumns && index > 0) {
+            cells.push(
+                <UnsortableTableHeaderCell
+                    key={`delta-${date}`}
+                    help="The difference between the measurement values on the previous and next date.
+                    A plus (+) sign indicates that the newer value is higher, a minus (-) sign that it is lower.
+                    A green outline indicates that the newer value is better, a red outline that it is worse."
+                    label="𝚫"
```
wkoot, line 238: Is this the ASCII capital delta, or a unicode char?

wkoot: The ASCII version should be a Δ, which shows different from 𝚫

components/frontend/src/subject/SubjectTableRow.js
```diff
@@ -25,28 +26,97 @@ import {
...
+        const improved = didValueImprove(increased, direction)
+        const evaluateTarget = metric.evaluate_targets ?? true
+        let alt = "The measurement value changed";
+        let color = "blue";
+         if (evaluateTarget && improved) {
+            alt = "The measurement value improved";
+            color = "green"
+        }
+        if (evaluateTarget && !improved) {
+            alt = "The measurement value worsened";
```
wkoot, line 70: Is it okay to not address the case where the value stayed the same? Do we skip that?

components/frontend/src/subject/SubjectTableRow.js
```diff
@@ -25,28 +26,97 @@ import {
...
+            color = "red"
+        }
+        let delta = increased ? "+" : "-"
```
wkoot, line 73: This one also omits the case of value staying the same

components/frontend/src/subject/SubjectTableRow.test.js
```diff
@@ -0,0 +1,105 @@
...
+import { SubjectTableRow } from './SubjectTableRow';
+import { createTestableSettings, datamodel, report } from "../__fixtures__/fixtures";
+
+beforeEach(() => {
+    history.push("")
+})
+
+function renderSubjectTableRow(
+    {
+        direction = "<",
```
wkoot, line 14: Seems useful to also test the opposite direction

components/frontend/src/subject/SubjectTableRow.js
```diff
@@ -25,28 +26,97 @@ import {
...
+        if (metric.scale !== "version_number") {
+            delta += `${Math.abs(metricValue - previousValue)}`
+            alt += ` with ${delta}`
```
wkoot:
```suggestion
            alt += ` by ${delta}`
```

wkoot: I think "worsened by", "improved by" and "changed by" are more accurate when dealing with numeric values

components/frontend/src/sharedPropTypes.js
fniessink: Good catch. I was assuming number values had already been converted, but I was hallucinating like a LLM.

components/frontend/src/subject/SubjectTableHeader.js
fniessink: It's a unicode char. Should work fine.

components/frontend/src/subject/SubjectTableRow.js
fniessink: Yes, in that case the delta cell is empty.

components/frontend/src/subject/SubjectTableRow.js
fniessink: As designed, I'll add a comment.

components/frontend/src/subject/SubjectTableRow.test.js
fniessink: That is done, that's why it's a parameter here.

wkoot (approved):

PR #8162, opened 2024-02-26
Add eslint to quality checks. Closes #7693.

### LINKED ISSUE #7693: Properly configure eslint and make it fail the build if code is not conformant

fniessink:
This PR configures ESLint and fixes ESLint errors and warnings, most notably it adds propTypes to every React component.

### DIFF
--- a/components/frontend/.eslintrc.json
+++ b/components/frontend/.eslintrc.json
@@ -1,0 +1,46 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "jest": true,
+        "node": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:react/recommended",
+        "plugin:react/jsx-runtime"
+    ],
+    "parserOptions": {
+        "ecmaVersion": "latest",
+        "sourceType": "module"
+    },
+    "plugins": [
+        "react"
+    ],
+    "rules": {
+        "no-unused-vars": [
+            "error",
+            {
+                "argsIgnorePattern": "^_"
+            }
+        ]
+    },
+    "overrides": [
+        {
+            "files": [
+                "*.test.js"
+            ],
+            "rules": {
+                "no-import-assign": "off"
+            }
+        }
+    ],
+    "settings": {
+        "react": {
+            "version": "detect"
+        }
+    },
+    "globals": {
+        "global": "readonly"
+    }
+}
--- a/components/frontend/package.json
+++ b/components/frontend/package.json
@@ -24,14 +24,16 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
+    "cov": "react-scripts test --coverage --watchAll --transformIgnorePatterns 'node_modules/(?!history)/'",
     "test": "react-scripts test --maxWorkers 2 --transformIgnorePatterns \"node_modules/(?!history)/\"",
     "eject": "react-scripts eject"
   },
-  "eslintConfig": {
-    "extends": "react-app"
-  },
   "jest": {
-    "resetMocks": false
+    "resetMocks": false,
+    "coveragePathIgnorePatterns": [
+        "<rootDir>/node_modules/",
+        "<rootDir>/src/index.js"
+    ]
   },
   "browserslist": [
     ">0.2%",
@@ -44,6 +46,12 @@
     "@testing-library/jest-dom": "^6.4.1",
     "@testing-library/react": "^14.2.1",
     "@testing-library/user-event": "^14.5.2",
+    "eslint": "^8.57.0",
+    "eslint-config-standard": "^17.1.0",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-n": "^16.6.2",
+    "eslint-plugin-promise": "^6.1.1",
+    "eslint-plugin-react": "^7.33.2",
     "react-scripts": "5.0.1",
     "react-test-renderer": "^18.2.0"
   }
--- a/components/frontend/src/issue/IssuesRows.js
+++ b/components/frontend/src/issue/IssuesRows.js
@@ -1,5 +1,5 @@
 import { useState } from 'react';
-import PropTypes from 'prop-types';
+import { bool, func, node, string } from 'prop-types';
 import { Grid } from 'semantic-ui-react';
 import { set_metric_attribute, add_metric_issue } from '../api/metric';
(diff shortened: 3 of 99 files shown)


components/frontend/.eslintrc.json
```diff
@@ -0,0 +1,46 @@
...
+                "*.test.js"
+            ],
+            "rules": {
+                "no-import-assign": "off"
+            }
+        }
+    ],
+    "settings": {
+        "react": {
+            "version": "detect"
```
wkoot, line 40: Is it not possible to pin this to the version we've set in package.json, or is that not something easily updated by the auto update PR bots?

components/frontend/.eslintrc.json
```diff
@@ -0,0 +1,46 @@
...
+        "sourceType": "module"
+    },
+    "plugins": [
+        "react"
+    ],
+    "rules": {
+        "no-unused-vars": [
+            "error",
+            {
+                "argsIgnorePattern": "^_"
```
wkoot, line 24: Why ignore all vars starting with underscore, is it then not better to use `^_$`?

components/frontend/package.json
```diff
@@ -24,14 +24,16 @@
...
   },
-  "eslintConfig": {
-    "extends": "react-app"
-  },
   "jest": {
-    "resetMocks": false
+    "resetMocks": false,
+    "coveragePathIgnorePatterns": [
+        "<rootDir>/node_modules/",
+        "<rootDir>/src/index.js"
```
wkoot, line 35: Why do we ignore this, is it a generated file?

components/frontend/.eslintrc.json
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    "settings": {
+        "react": {
+            "version": "detect"
+        }
+    },
+    "globals": {
+        "global": "readonly"
+    }
+}
```
wkoot:
```suggestion
}

```

components/frontend/src/issue/IssuesRows.js
```diff
@@ -68,19 +68,19 @@ function IssueIdentifiers({ entityKey, issueTrackerInstruction, metric, metric_u
...
+    report_uuid: string,
+    target: string,
+    reload: func
 }
 
 export function IssuesRows({ metric, metric_uuid, reload, report, target }) {
     const parameters = report?.issue_tracker?.parameters;
     const issueTrackerConfigured = Boolean(report?.issue_tracker?.type && parameters?.url && parameters?.project_key && parameters?.issue_type);
-    const issueTrackerInstruction = issueTrackerConfigured ? null : <p>Please configure an issue tracker by expanding the report title, selecting the 'Issue tracker' tab, and configuring an issue tracker.</p>;
+    const issueTrackerInstruction = issueTrackerConfigured ? null : <p>Please configure an issue tracker by expanding the report title, selecting the &apos;Issue tracker&apos; tab, and configuring an issue tracker.</p>;
```
wkoot, line 83: Interesting that this string literal doesn't need to be quoted, is it not something worth adding explicitly?

components/frontend/src/issue/IssuesRows.js
```diff
@@ -150,10 +150,10 @@ export function IssuesRows({ metric, metric_uuid, reload, report, target }) {
...
+    entityKey: string,
     metric: metricPropType,
-    metric_uuid: PropTypes.string,
-    reload: PropTypes.func,
+    metric_uuid: string,
+    reload: func,
     report: reportPropType,
-    target: PropTypes.string,
+    target: string,
 }
```
wkoot:
```suggestion
}

```

components/frontend/src/measurement/Overrun.js
```diff
@@ -65,3 +67,10 @@ export function Overrun({ metric_uuid, metric, report, measurements, dates }) {
         />
     )
 }
+Overrun.propTypes = {
+    dates: datesPropType,
+    measurements: measurementsPropType,
+    metric: metricPropType,
+    metric_uuid: string,
+    report: reportPropType,
+}
```
wkoot:
```suggestion
}

```

components/frontend/src/metric/MetricType.js
```diff
@@ -1,4 +1,5 @@
 import { useContext } from 'react';
+import { func, string} from 'prop-types';
```
wkoot:
```suggestion
import { func, string } from 'prop-types';
```

components/frontend/src/utils.js
```diff
@@ -271,7 +271,7 @@ export function nice_number(number) {
         }
         rounded_numbers = rounded_numbers.map((value) => { return value * 10 });
     }
-    while (true);
+    while (true);  // eslint-disable-line no-constant-condition
```
wkoot, line 274: What would be the eslint acceptable implementation for this? Perhaps worth doing in a separate issue?

components/frontend/.eslintrc.json
fniessink: I don't know whether Dependabot can pick up a version number in this config file, but why would we want to pin the React version, assuming the detection works of course?

components/frontend/.eslintrc.json
fniessink: This doesn't ignore files, it ignores function arguments that start with an underscore. For example,

```javascript
function eventHandler(_event, data) {
   // Do something with data, but ignore the event
}
```

components/frontend/package.json
fniessink: I don't see how it can be covered from the unit tests... ☹️

components/frontend/src/issue/IssuesRows.js
fniessink: It's not a string literal, it's JSX I guess.

components/frontend/src/measurement/Overrun.js
fniessink: I've changed my VS Code settings to prevent this in the future. Should've done that before of course.

components/frontend/src/metric/MetricType.js
fniessink: I want to also add a style checker to check/fix issues like this. That's for another PR.

components/frontend/src/utils.js
fniessink: I guess a boolean flag would get rid of the warning, but I'm not sure the implementation would be better.

PR #8163, opened 2024-02-26
Make internal endpoints explicitly internal instead of versioned

### LINKED ISSUE #5791: Make internal endpoints explicitly internal instead of versioned
Tasks:
- Remove versioning from the frontend API, e.g. `/api/v3/nr_measurements` -> `/api/internal/nr_measurements`
- Retain versioning for the public endpoints as documented in `api.md`
- Update docs.

wkoot:
Closes #5791

### DIFF
--- a/components/api_server/src/routes/auth.py
+++ b/components/api_server/src/routes/auth.py
@@ -127,6 +127,7 @@ def verify_user(database: Database, username: str, password: str) -> User:
     return user
 
 
+@bottle.post("/api/internal/login", authentication_required=False)
 @bottle.post("/api/v3/login", authentication_required=False)
 def login(database: Database) -> dict[str, bool | str]:
     """Log the user in. Add credentials as JSON payload, e.g. {username: 'user', password: 'pass'}."""
@@ -145,6 +146,7 @@ def login(database: Database) -> dict[str, bool | str]:
     }
 
 
+@bottle.post("/api/internal/logout", authentication_required=True)
 @bottle.post("/api/v3/logout", authentication_required=True)
 def logout(database: Database) -> dict[str, bool]:
     """Log the user out."""
@@ -152,6 +154,7 @@ def logout(database: Database) -> dict[str, bool]:
     return {"ok": True}
 
 
+@bottle.get("/api/internal/public_key", authentication_required=False)
 @bottle.get("/api/v3/public_key", authentication_required=False)
 def get_public_key(database: Database) -> dict:
     """Return a serialized version of the public key."""
--- a/components/api_server/src/routes/auth.py
+++ b/components/api_server/src/routes/auth.py
@@ -127,6 +127,7 @@ def verify_user(database: Database, username: str, password: str) -> User:
     return user
 
 
+@bottle.post("/api/internal/login", authentication_required=False)
 @bottle.post("/api/v3/login", authentication_required=False)
 def login(database: Database) -> dict[str, bool | str]:
     """Log the user in. Add credentials as JSON payload, e.g. {username: 'user', password: 'pass'}."""
@@ -145,6 +146,7 @@ def login(database: Database) -> dict[str, bool | str]:
     }
 
 
+@bottle.post("/api/internal/logout", authentication_required=True)
 @bottle.post("/api/v3/logout", authentication_required=True)
 def logout(database: Database) -> dict[str, bool]:
     """Log the user out."""
--- a/components/api_server/src/routes/auth.py
+++ b/components/api_server/src/routes/auth.py
@@ -127,7 +127,6 @@ def verify_user(database: Database, username: str, password: str) -> User:
     return user
 
 
-@bottle.post("/api/internal/login", authentication_required=False)
 @bottle.post("/api/v3/login", authentication_required=False)
 def login(database: Database) -> dict[str, bool | str]:
     """Log the user in. Add credentials as JSON payload, e.g. {username: 'user', password: 'pass'}."""
@@ -146,7 +145,6 @@ def login(database: Database) -> dict[str, bool | str]:
     }
 
 
-@bottle.post("/api/internal/logout", authentication_required=True)
 @bottle.post("/api/v3/logout", authentication_required=True)
 def logout(database: Database) -> dict[str, bool]:
     """Log the user out."""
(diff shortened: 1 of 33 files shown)


components/api_server/src/routes/metric.py
```diff
@@ -4,22 +4,20 @@
...
 from shared.database.measurements import insert_new_measurement, latest_measurement, latest_successful_measurement
 from shared.model.metric import Metric
 from shared.utils.type import MetricId, SubjectId, Value
 from shared_data_model import DATA_MODEL
-
-from database.reports import insert_new_report, latest_report_for_uuids, latest_reports
-from model.actions import copy_metric, move_item
-from model.defaults import default_metric_attributes
 from utils.functions import sanitize_html, uuid
-
```
fniessink: Why are the imports reorganised? And why didn't ruff complain?

components/api_server/src/routes/report.py
```diff
@@ -76,6 +76,7 @@ def get_report(database: Database, report_uuid: ReportId | None = None):
     return {"ok": True, "reports": summarized_reports}
 
 
+@bottle.get("/api/internal/report/<report_uuid>/metric_status_summary", authentication_required=False)
 @bottle.get("/api/v3/report/<report_uuid>/metric_status_summary", authentication_required=False)
```
fniessink: Why is this endpoint internal? It's not used by the frontend I think?

components/api_server/src/routes/report.py
```diff
@@ -85,6 +86,7 @@ def get_report_metric_status_summary(database: Database, report: Report, report_
     return {"report_uuid": report_uuid, "title": summarized_report["title"], **summarized_report["summary"]}
 
 
+@bottle.post("/api/internal/report/import", permissions_required=[EDIT_REPORT_PERMISSION])
```
fniessink: Why is this endpoint internal? It's not used by the frontend I think?

components/api_server/src/routes/report.py
```diff
@@ -136,12 +138,14 @@ def post_report_copy(database: Database, report: Report, report_uuid: ReportId):
...
 
 
+@bottle.get("/api/internal/report/<report_uuid>/pdf", authentication_required=False)
 @bottle.get("/api/v3/report/<report_uuid>/pdf", authentication_required=False)
 def export_report_as_pdf(report_uuid: ReportId):
     """Download the report as PDF."""
     return export_as_pdf(report_uuid)
 
 
+@bottle.get("/api/internal/report/<report_uuid>/json", authentication_required=True)
```
fniessink: This endpoint is not used by the frontend, so why make it internal?

components/api_server/src/routes/search.py
```diff
@@ -75,6 +75,7 @@ def match_attribute(
     return attribute_name in domain_object and domain_object[attribute_name] == attribute_value
 
 
+@bottle.post("/api/internal/<domain_object_type>/search", authentication_required=False)
```
fniessink: Why is this endpoint internal? It's not used by the frontend?

components/api_server/src/routes/server.py
```diff
@@ -6,6 +6,7 @@
 QUALITY_TIME_VERSION = "5.8.0"
 
 
+@bottle.get("/api/internal/server", authentication_required=False)
```
fniessink: Why is this endpoint internal? It's not used by the frontend I think?

components/collector/tests/source_collectors/quality_time/test_metrics.py
```diff
@@ -13,7 +13,7 @@ class QualityTimeMetricsTest(QualityTimeTestCase):
     def setUp(self):
         """Set up test data."""
         super().setUp()
-        self.api_url = f"{self.url}/api/v3/report"
+        self.api_url = f"{self.url}/api/internal/report"
```
fniessink: I think the Quality-time collector should use the public API. In theory the source Quality-time could be a different version so it's better to go via a versioned endpoint.

components/collector/tests/source_collectors/quality_time/test_source_up_to_dateness.py
```diff
@@ -18,7 +18,7 @@ class QualityTimeSourceUpToDatenessTest(QualityTimeTestCase):
     def setUp(self):
         """Set up test data."""
         super().setUp()
-        self.api_url = f"{self.url}/api/v3/reports"
+        self.api_url = f"{self.url}/api/internal/reports"
```
fniessink: I think the Quality-time collector should use the public API. In theory the source Quality-time could be a different version so it's better to go via a versioned endpoint.

docs/src/api.md
fniessink, line 1: The file is reformatted, but not changed? I can't tell. Please move the reformatting to a different PR.

components/api_server/src/routes/auth.py
```diff
@@ -127,6 +127,7 @@ def verify_user(database: Database, username: str, password: str) -> User:
     return user
 
 
+@bottle.post("/api/internal/login", authentication_required=False)
 @bottle.post("/api/v3/login", authentication_required=False)
```
fniessink, line 131: Why both internal and v3?

components/api_server/src/routes/auth.py
```diff
@@ -145,13 +146,15 @@ def login(database: Database) -> dict[str, bool | str]:
     }
 
 
+@bottle.post("/api/internal/logout", authentication_required=True)
 @bottle.post("/api/v3/logout", authentication_required=True)
```
fniessink, line 150: Why both internal and v3?

components/api_server/src/routes/auth.py
```diff
@@ -145,13 +146,15 @@ def login(database: Database) -> dict[str, bool | str]:
...
 def logout(database: Database) -> dict[str, bool]:
     """Log the user out."""
     delete_session(database)
     return {"ok": True}
 
 
+@bottle.get("/api/internal/public_key", authentication_required=False)
 @bottle.get("/api/v3/public_key", authentication_required=False)
```
fniessink: This endpoint is not used by the frontend, so why internal?

fniessink (changes requested):
I think we need to consider more carefully which endpoints are internal, which are public, and which are both. I see many endpoints marked as internal that I think are not used by the frontend. I also see endpoints marked as public that I think are only internal.

Also, since we're changing the API, I think these changes should be added to the changelog.

components/api_server/src/routes/metric.py
wkoot: Oh, that's weird! Probably incorrect IDE autoformatting, or perhaps it was even ruff reformatting it when I put it through `--fix`

wkoot: Just checked, ruff doesn't raise any objections in either way the imports are organised. Perhaps it thinks they might be external package imports also?

fniessink (approved):

PR #8239, opened 2024-03-07
Improve the frontend measurement scale utility functions

### LINKED ISSUE #8164: Harmonise getting the scale from a metric in the frontend
`utils.js` provides a `getMetricScale` method, but it's not used everywhere it should be.

fniessink:


components/frontend/src/metric/TrendGraph.test.js
```diff
@@ -28,21 +28,26 @@ it('renders the time axis', () => {
...
 
 it('renders the measurements', () => {
-    renderTrendgraph([{ count: { value: "1" }, start: "2019-09-29", end: "2019-09-30" }]);
+    renderTrendgraph({ measurements: [{ count: { value: "1" }, start: "2019-09-29", end: "2019-09-30" }] });
     expect(screen.getAllByText(/Time/).length).toBe(1);
 });
 
 it('renders the measurements in dark mode', () => {
-    renderTrendgraph([{ count: { value: "1" }, start: "2019-09-29", end: "2019-09-30" }], true);
+    renderTrendgraph({ measurements: [{ count: { value: "1" }, start: "2019-09-29", end: "2019-09-30" }], darkMode: true });
```
wkoot, line 36: Why do we explicitly test for dark mode, do we have custom logic for that? The test only asserts that the text is there, so it doesn't really verify anything particular regarding dark mode right?

wkoot (approved):

components/frontend/src/metric/TrendGraph.test.js
fniessink: Correct, but the `TrendGraph` sets some colours based on dark/light mode so in order to get higher coverage we need to render it in dark mode as well. It's not interesting to check the exact colours, but it is useful to check whether rendering in general works in dark mode too.

PR #8317, opened 2024-03-19
Group source parameters so they are easier to distinguish. Groups are…

### LINKED ISSUE #7489: Split source configuration into two parts visually: location and configuration
# User story

As a user I want to distinguish between configuring the location of a source and other source parameters so not all configuration parameters are displayed together.

## Acceptance criteria

- [ ] Split the configuration tab of sources into a location panel and a configuration panel. The location panel contains the location parameters: URLs, credentials, etc (these depend on the source type).
- [ ] Panels are lightweight: no borders and no title. Location panel is the bottom panel.

## Assumptions

- The report structure does not change.

## Out of scope

- Splitting the source into generic and specific parts, see #3127.

# Tasks

### Frontend

- [ ] Add two separate panels for location and configuration, based on what type parameters are in the data model.

### Shared code

- [ ] Move the location parameters (access parameters) in the data model to a separate `location_parameters` dict.

## Documentation

- [ ] Update changelog
- [ ] Update screenshots

Reported by @Sebastiaan127001

fniessink:
… "Source location and credentials", "Source filters", and "Manual source data". Note that not all parameter groups are displayed for each source as the applicability of parameters depends on the combination of metric and source type. Closes #7489.


components/frontend/src/source/Source.js
```diff
@@ -68,15 +68,17 @@ function Parameters({ changed_fields, config_error, connection_error, metric, pa
                     />
                 </Grid.Column>
             </Grid.Row>
-            <Grid.Row columns={2}>
```
wkoot, line 71: Was it incorrectly setting 2 columns before?

components/frontend/src/source/SourceParameters.js
```diff
@@ -1,51 +1,87 @@
...
+                    warning={changed_param_keys?.indexOf(parameterKey) !== -1}
+                    reload={reload}
+                    index={index}
+                />
+            </div>
+        ));
+        return (
+            <Grid.Column key={parameterGroup.name}>
+                <Segment>
+                    <Header as="h5" color="grey">{parameterGroup.name}</Header>
```
wkoot, line 78: Does this `grey` also work in dark mode?

components/frontend/src/source/SourceParameters.js
```diff
@@ -1,51 +1,87 @@
...
+                    {parameters}
+                </Segment>
+            </Grid.Column>
+        )
+    })
+    return <Grid><Grid.Row columns={2}>{groups}</Grid.Row></Grid>
```
wkoot, line 84: Why 2 columns when there is only one `Grid.Column` returned?

components/frontend/src/source/SourceParameters.test.js
```diff
@@ -16,19 +16,36 @@ function renderSourceParameters({placeholder = "", type = "string", source_param
...
-                report={{subjects: {}}}
-                metric={{type: "violations"}}
-                source={{ type: "source_type", parameters: {parameter_key: source_parameter_value} }}
+                report={{ subjects: {} }}
+                metric={{ type: "violations" }}
+                source={{ type: "source_type", parameters: { parameter_key: source_parameter_value } }}
                 changed_param_keys={changed_param_keys}
             />
-        </DataModel.Provider>
+        </DataModel.Provider >
```
wkoot: Why the additional space?

components/frontend/src/source/SourceParameters.js
fniessink: Yup, works fine.

components/frontend/src/source/Source.js
fniessink: No, it was correct. In the previous version, the parameters would be laid out in two columns because of this setting. In the current version, the `SourceParameters` component does the layout across columns.

components/frontend/src/source/SourceParameters.test.js
fniessink: Hmm, inserted by accident apparently and it stayed there because formatter in VS Code does not remove it.

components/frontend/src/source/SourceParameters.js
fniessink: There are possibly multiple columns returned by `Object.values(parameterLayout).map()`.

PR #8335, opened 2024-03-24
7404 use an opinionated code formatter for javascript

### LINKED ISSUE #7404: Use an opinionated code formatter for Javascript
For example:
- Prettier: https://prettier.io

fniessink:
Add Prettier code formatter and ESLint import plugin to sort imports.

This leads to a lot of changes obviously:
- Reorganized/resorted imports
- Semicolumns removed where possible
- Trailing commas added
- Double quotes instead of single quotes
- Ternary operators, chained functions calls (`a().b().c()`), React components (`<Parent><Child/></Parent>`) and anonymous functions are distributed across multiple lines
- One space for inline comments instead of two (sadly this is not configurable)

Note:
- The prettifier width is set to 120 but this a desired width so lines can occasionally be longer, see  https://prettier.io/docs/en/options#print-width.
- The frontend component now also has `ci/unittest.sh` and `ci/quality.sh` scripts.
- SonarQube complains about duplication and test coverage, but that's not really due to this change of course.


components/frontend/src/__fixtures__/fixtures.js
```diff
@@ -79,16 +88,22 @@ export function createTestableSettings() {
         hiddenTags: renderHook(() => useHiddenTagsURLSearchQuery()).result.current,
         metricsToHide: renderHook(() => useMetricsToHideURLSearchQuery()).result.current,
         nrDates: renderHook(() => useNrDatesURLSearchQuery()).result.current,
-        showIssueCreationDate: renderHook(() => useShowIssueCreationDateURLSearchQuery()).result.current,
+        showIssueCreationDate: renderHook(() => useShowIssueCreationDateURLSearchQuery()).result
```
wkoot: Why the line breaks here, max line length? I'd say the previous line was more clear

components/frontend/src/setupTests.js
```diff
@@ -1 +1 @@
-import '@testing-library/jest-dom'
+import "@testing-library/jest-dom"
```
wkoot, line 1: Do we exclude this file and the other .test. files when building/pushing to dockerhub?

components/frontend/src/sharedPropTypes.js
```diff
@@ -180,7 +207,7 @@ export const reportsOverviewPropType = shape({
     comment: string,
     permissions: object,
     title: string,
-    subtitle: string
+    subtitle: string,
```
wkoot, line 204: Interesting that sometimes we have trailing slashes in js lists, but most times we don't!

components/frontend/src/utils.js
```diff
@@ -1,77 +1,93 @@
...
-    "near_target_met": `${STATUS_NAME.near_target_met}: the measurement value is close to the target value.`,
-    "target_not_met": `${STATUS_NAME.target_not_met}: the measurement value does not meet the target value.`,
-    "debt_target_met": <>{`${STATUS_NAME.debt_target_met}: the measurement value does not meet the\ntarget value, but this is accepted as `}<HyperLink url="https://en.wikipedia.org/wiki/Technical_debt">technical debt</HyperLink>{". The measurement\nvalue does meet the technical debt target."}</>,
-    "unknown": `${STATUS_NAME.unknown}: the status could not be determined because no sources have\nbeen configured for the metric yet or the measurement data could not\nbe collected.`
+    informative: `${STATUS_NAME.informative}: the measurement value is not evaluated against a target value.`,
+    target_met: `${STATUS_NAME.target_met}: the measurement value meets the target value.`,
+    near_target_met: `${STATUS_NAME.near_target_met}: the measurement value is close to the target value.`,
+    target_not_met: `${STATUS_NAME.target_not_met}: the measurement value does not meet the target value.`,
+    debt_target_met: (
+        <>
```
wkoot, line 42: Weird notation, what does this do?

components/frontend/src/__fixtures__/fixtures.js
fniessink: I've solved it by getting rid of the duplication.

components/frontend/src/setupTests.js
fniessink: That's what CreateReactApp, building on Webpack, does for us.

components/frontend/src/sharedPropTypes.js
fniessink: Trailing comma's you mean?

components/frontend/src/utils.js
fniessink: The `<>` is a React fragment. It can be used if there are multiple react components, you have to return one, but you don't want to put them into a div or other container.

PR #8345, opened 2024-03-26
Add support for using Dependency-Track as source for the source versi…

### LINKED ISSUE #7171: Add Dependency Track as source for security warnings and dependencies metrics
Acceptance criteria:
- [x] Security warnings metric
- [x] (outdated) Dependencies metric
- [x] Source version
- [x] Source up-to-dateness (Last BOM Import)

Blocked by #8503 due to the amount of pagination needed to get data from DT.

fniessink:
…on metric. Closes #7171.


components/shared_code/src/shared_data_model/sources/dependency_track.py
```diff
@@ -0,0 +1,88 @@
...
+            ],
+        ),
+        "security_warnings": Entity(
+            name="security warning",
+            attributes=[
+                EntityAttribute(name="Project", url="project_landing_url"),
+                EntityAttribute(name="Component", url="component_landing_url"),
+                EntityAttribute(name="Identifier"),
+                EntityAttribute(name="Description"),
+                EntityAttribute(name="Severity", color={"Critical": Color.NEGATIVE, "High": Color.WARNING}),
```
wkoot, line 77: Why are only the two highest severities mapped to colours, shouldn't it be up to the user to classify threshold?

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -0,0 +1,45 @@
...
+        if api_key := str(self._parameter("private_token")):
+            headers["X-Api-Key"] = api_key
+        return headers
+
+    async def _get_source_responses(self, *urls: URL) -> SourceResponses:
+        """Extend to load multiple pages, if necessary."""
+        responses = SourceResponses()
+        for url in urls:
+            page_nr = 1  # Page numbers start at 1
+            total_count = self.PAGE_SIZE + 1  # Total count is still unknown, make it so big we get at least one page
```
wkoot, line 32: Ouch, stupid api :(

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -0,0 +1,45 @@
...
+        headers = super()._headers()
+        if api_key := str(self._parameter("private_token")):
+            headers["X-Api-Key"] = api_key
+        return headers
+
+    async def _get_source_responses(self, *urls: URL) -> SourceResponses:
+        """Extend to load multiple pages, if necessary."""
+        responses = SourceResponses()
+        for url in urls:
+            page_nr = 1  # Page numbers start at 1
```
wkoot, line 31: If you start `page_nr` at 0 and increment it at the start of the while loop, then you don't need to add `PAGE_SIZE` to the `total_count` because 0 < 1

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -0,0 +1,45 @@
...
+            total_count = self.PAGE_SIZE + 1  # Total count is still unknown, make it so big we get at least one page
+            while page_nr * self.PAGE_SIZE < total_count:
+                offsetted_url = URL(f"{url}{"&" if "?" in url else "?"}pageSize={self.PAGE_SIZE}&pageNumber={page_nr}")
+                response = (await super()._get_source_responses(offsetted_url))[0]
+                total_count = int(response.headers.get("X-Total-Count", 0))
```
wkoot, line 36: It sucks that the API does this, because the total count could change while we're paginating. If the most recent data is at the *start* of pagination (i.e. age desc), we would fetch outdated results multiple times :( Wouldn't it make sense to abort (and start over or requeue) the fetch if total_count changes (aside from the first time) during the loop?

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -0,0 +1,45 @@
...
+                responses.append(response)
+                page_nr += 1
+        return responses
+
+    async def _get_project_uuids(self) -> dict[str, str]:
+        """Return a mapping of project UUIDs to project names."""
+        projects_api = URL(await self._api_url() + "/project")
+        project_uuids = {}
+        for response in await DependencyTrackBase._get_source_responses(self, projects_api):
```
wkoot, line 45: I think we'd really benefit from memoization here, especially when paginating is troublesome and might time out

components/collector/src/source_collectors/dependency_track/dependencies.py
```diff
@@ -0,0 +1,81 @@
...
+        project_components_urls = [URL(f"{api_url}/component/project/{uuid}") for uuid in project_uuids]
+        return await super()._get_source_responses(*project_components_urls)
+
+    async def _parse_entities(self, responses: SourceResponses) -> Entities:
+        """Parse the entities from the responses."""
+        entities = Entities()
+        for response in responses:
+            components = await response.json(content_type=None)
+            entities.extend([self._create_entity(component) for component in components])
+        return entities
```
wkoot, line 50:
```suggestion
        entities = [self._create_entity(component) for component in await response.json(content_type=None) for response in responses]
        return Entities(entities)
```

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
```diff
@@ -0,0 +1,54 @@
...
+                    project=project["name"],
+                    project_landing_url=f"{landing_url}/projects/{uuid}",
+                )
+                entities.append(entity)
+        return entities
+
+    @staticmethod
+    def _last_bom_import_datetime(project: DependencyTrackProject) -> datetime:
+        """Return the last BOM import datetime for the project."""
+        return datetime_fromtimestamp(int(project["lastBomImport"]) / 1000.0)
```
wkoot: Don't we have some helper method for this? I think we often have the millis case

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
```diff
@@ -0,0 +1,72 @@
...
+        self.set_source_parameter("severities", ["High", "Critical"])
+        response = await self.collect(get_request_json_side_effect=[self.projects, self.vulnerabilities])
+        self.assert_measurement(response, value="0", entities=[])
+
+    async def test_api_key(self):
+        """Test that the API key is passed as header."""
+        self.set_source_parameter("private_token", "API key")
+        _, get, _ = await self.collect(get_request_json_return_value=[], return_mocks=True)
+        get.assert_called_once_with(
+            "https://dependency_track/api/v1/project?pageSize=100&pageNumber=1",
```
wkoot: How is the api_url determined to be `https://dependency_track`, I don't see that in the test code?

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
```diff
@@ -0,0 +1,75 @@
...
+    def setUp(self) -> None:
+        """Extend to set up Dependency-Track projects."""
+        super().setUp()
+        self.projects = [{"uuid": "project uuid", "name": "project name"}]
+
+    def create_dependencies(self, latest_version: str) -> list[DependencyTrackComponent]:
+        """Create a list of dependencies as returned by Dependency-Track."""
+        return [
+            {
+                "repositoryMeta": {"latestVersion": latest_version} if latest_version else {},
```
wkoot: I thought the key `repositoryMeta` was not present in the return data when it didn't have this information (i.e. we're doing a `.get("repositoryMeta", {})`)

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
```diff
@@ -0,0 +1,72 @@
...
+                    "description": "vulnerability description",
+                    "severity": "UNASSIGNED",
+                    "vulnId": "CVE-123",
+                },
+            },
+        ]
+        self.entities = [
+            {
+                "component": "component name",
+                "component_landing_url": "/components/component uuid",
```
wkoot: Perhaps better to use a dash instead of space in the fake uuids

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
```diff
@@ -0,0 +1,47 @@
+"""Unit tests for the Dependency-Track source up-to-dateness collector."""
+
+from datetime import datetime, timedelta
+
+from dateutil.tz import tzlocal
+
+from tests.source_collectors.source_collector_test_case import SourceCollectorTestCase
+
+
+class DependencyTrackSourceVersionTest(SourceCollectorTestCase):
```
wkoot:
```suggestion
class DependencyTrackUpToDatenessTest(SourceCollectorTestCase):
```

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
```diff
@@ -0,0 +1,47 @@
...
+
+    async def test_source_up_to_dateness(self):
+        """Test that the source up-to-datentess can be measured."""
+        now = datetime.now(tz=tzlocal()).replace(microsecond=0)
+        yesterday = now - timedelta(days=1)
+        yesterday_timestamp = yesterday.timestamp() * 1000
+        last_week = now - timedelta(days=7)
+        last_week_timestamp = last_week.timestamp() * 1000
+        projects = [
+            {"name": "Project 1", "lastBomImport": yesterday_timestamp, "uuid": "p1"},
```
wkoot, line 25: Isn't the timestamp a string of an int? Or is the code doing an int cast for typing check?

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
```diff
@@ -0,0 +1,47 @@
...
+            {"name": "Project 2", "lastBomImport": last_week_timestamp, "uuid": "p2"},
+        ]
+        response = await self.collect(get_request_json_return_value=projects)
+        self.assert_measurement(
+            response,
+            value="7",
```
wkoot, line 31: Didn't we want to be able to distinguish between the age of the vulnerabilities db (and age of analysis, I guess) and age of sbom? With this logic any DT installation which happens to have some old project listed, would seem to not be up to date. If there is no option to exclude certain projects from this metric, the user would be forced to delete old projects from DT

components/shared_code/src/shared_data_model/sources/dependency_track.py
fniessink: Semantic UI defines only a few styles for warnings (https://react.semantic-ui.com/collections/table/#states-positive-negative), hence the limited options here.

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Yup.

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Yeah, but that way it's not very clear that the page numbers start at 1 and a future maintainer may think the page number is increased in the wrong location. Since it's surprising it starts at 1 I wanted it to be very clear.

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Hmm, I don't think it's worth it. I guess the main risk is that we fetch the same items multiple times, but multiple entities with the same key are already filtered by the Entities class.

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Subclasses that need the project uuids/names already keep a reference to it. Unfortunately, `functools.cache` does not work with async methods. Not sure introducing another dependency is worth it.

components/collector/src/source_collectors/dependency_track/dependencies.py
fniessink: I'm a fan of list comprehensions, but nested list comprehensions not so much.

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
fniessink: I'll add one.

fniessink: Turns out all callers of datetime_fromtimestamp pass a timestamp in milliseconds so I could used move the `/ 1000.0` to the method :-)

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
fniessink: In `SourceCollectorTestCase.setUp`: 

```python
    def setUp(self) -> None:
        """Extend to set up the source and metric under test."""
        self.sources = {"source_id": {"type": self.SOURCE_TYPE, "parameters": {"url": f"https://{self.SOURCE_TYPE}"}}}
```

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
fniessink: Yeah, you're right. I'll add that to the test.

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
fniessink: I'll add both cases to the unit tests.

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
fniessink: I've changed the test.

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
fniessink: There doesn't seem to be information about the age of the vulnerabilities db and analysis available in the API.

components/shared_code/src/shared_data_model/sources/dependency_track.py
wkoot: But what if the user finds "Medium,High,Critical" important to know about? Or only "Critical"?

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
wkoot: Perhaps only divide by 1000 when the int has more than 10 (or 11, if you want to be future proof after 2286) digits?

PR #8442, opened 2024-04-10
Fix time ago labels.

### LINKED ISSUE #8440: Time ago label does not have a correct value any more
**Describe the bug**
Since the last release, the time ago label does not show the right value any more.

**To Reproduce**
1. Go to technical debt of a metric
2. see the label at the 'Technical debt end date' field

**Screenshots**
image

fniessink:
The time ago or to go, shown in the labels of date fields, would be incorrect.

Fixes #8440.

### DIFF
--- a/components/frontend/src/metric/MetricDebtParameters.js
+++ b/components/frontend/src/metric/MetricDebtParameters.js
@@ -74,7 +74,7 @@ function TechnicalDebtEndDate({ metric, metric_uuid, reload }) {
             requiredPermissions={[EDIT_REPORT_PERMISSION]}
             label={
                 <LabelWithDate
-                    date={Date(metric.debt_end_date)}
+                    date={metric.debt_end_date}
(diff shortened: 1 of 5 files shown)


components/frontend/src/metric/MetricDebtParameters.js
```diff
@@ -74,7 +74,7 @@ function TechnicalDebtEndDate({ metric, metric_uuid, reload }) {
             requiredPermissions={[EDIT_REPORT_PERMISSION]}
             label={
                 <LabelWithDate
-                    date={Date(metric.debt_end_date)}
+                    date={metric.debt_end_date}
```
wkoot, line 77: I'm surprised that there's no change in test code needed?

fniessink: Good point. I'll add a unit test.

PR #8451, opened 2024-04-11
Move filtering failed CI-jobs to GitLab.

### LINKED ISSUE #8444: Pass failure types as scope parameter to GitLab jobs API
When measuring failed CI-jobs with GitLab as source, pass the failure types as scope parameter to the jobs endpoints so Quality-time retrieves less data and does not have to do the filtering itself.

CC: @denniebouman

fniessink:
When measuring failed CI-jobs with GitLab as source, don't download all jobs and then filter them by failure status (canceled, failed, and/or skipped), but rather pass the failure states to the GitLab API endpoint so the filtering is done by GitLab.

Fixes #8444.

### DIFF
--- a/components/collector/src/source_collectors/gitlab/base.py
+++ b/components/collector/src/source_collectors/gitlab/base.py
@@ -72,11 +72,10 @@ class GitLabJobsBase(GitLabProjectBase):
         # created or by date run, so we're going to assume that descending order of IDs is roughly equal to descending
         # order of date created and date run. As soon as all jobs on a page have a build date that is outside the
         # lookback period we stop the pagination.
-        today = now().date()
-        lookback_date = today - timedelta(days=int(cast(str, self._parameter("lookback_days"))))
+        lookback_date = self._lookback_date()
         for response in responses:
             for job in await response.json():
-                if self._build_date(job) > lookback_date:
+                if self._build_date(job) >= lookback_date:
                     return await super()._next_urls(responses)
         return []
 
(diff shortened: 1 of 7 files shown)


components/collector/tests/source_collectors/gitlab/test_job_runs_within_time_period.py
```diff
@@ -57,55 +57,3 @@ async def test_job_lookback_days(self):
             },
         ]
         self.assert_measurement(response, value="1", entities=expected_entities)
-
-    async def test_jobs_not_deduplicated(self):
```
fniessink: I didn't understand what this was supposed to be testing. It seems to test the same thing as the first unit tests, but with the default loopback period. Removing it doesn't change the code coverage.

wkoot: The `GitLabJobsBase._jobs()` method only returns the newest job, `GitLabJobRunsWithinTimePeriod` overrides this. The test is meant to verify that this actually works. Changing the lookback period in the other test and checking that both jobs are returned is also possible

components/collector/src/source_collectors/gitlab/base.py
```diff
@@ -114,16 +113,26 @@ def newer(job1: Job, job2: Job) -> Job:
...
+        )
 
     @staticmethod
     def _build_date(job: Job) -> date:
         """Return the build date of the job."""
         return parse_datetime(job.get("finished_at") or job["created_at"]).date()
 
+    def _lookback_date(self) -> date:
+        """Return the lookback cut-off date."""
+        today = now().date()
```
wkoot, line 133: I see that `now()` comes from `shared.utils.date_time`, while the `parse_datetime()` comes from `collector_utilities.date_time`. I guess it's fine that we have separate locations for this code to live, but the `collector_utilities.date_time` is actually reimplementing `shared.utils.date_time.now` in `days_ago` and `days_to_go`. Perhaps good to clean this up, so we don't need to do shotgun surgery at some point

components/collector/src/source_collectors/gitlab/failed_jobs.py
```diff
@@ -8,6 +9,12 @@
 class GitLabFailedJobs(GitLabJobsBase):
     """Collector class to get failed job counts from GitLab."""
 
+    async def _api_url(self) -> URL:
+        """Extend to add the scope parameter of the jobs API."""
+        failure_types = list(self._parameter("failure_type"))
+        scopes = "".join([f"&scope[]={failure_type}" for failure_type in failure_types])
+        return URL(await super()._api_url() + scopes)
```
wkoot:
```suggestion
        scopes = "&".join([f"scope[]={failure_type}" for failure_type in failure_types])
        return URL(f"{await super()._api_url()}&{scopes}")
```

components/collector/tests/source_collectors/gitlab/test_job_runs_within_time_period.py
fniessink: Test fixed.

components/collector/src/source_collectors/gitlab/base.py
fniessink: Alas, replacing `difference = (datetime.now(tz=date_time.tzinfo) - date_time).days` with `difference = (now() - date_time).days` is not equivalent.

Issue #8356, opened 2024-03-27
Modify Proxy-pod to run Nginx as non Root user 


jeroenma:
The current configuration for the Proxy-pod in quality-time runs as root user. In some Kubernetes environments it is not allowed to run it as a root user. 

Request is to make it configurable to run it under a non-root user. The TCP-port probably has to change to a port number above 1024. 

So for backwards compatibility it has to be configurable of default it to 80.

jeroenma:
I fixed the issues local by using standard Nginx image and move LISTEN to Quality Time config. 
No changes required to this image.

wkoot:
@jeroenma would the changes as being drafted in https://github.com/ICTU/quality-time/pull/8526/files also work for your use case?

PR #8498, opened 2024-04-15
 Modify Proxy-pod to run Nginx as non Root user #8356

jeroenma:
The listening port of Nginx is now configurable using de environment variable NGINX_LISTEN. Default is still port 80, for 
backwards compatiblity. By pushing port to 8080, Nginx proces can be without root privileges.

Removed non functional config files and modified the main config file.
Using the envsubst command is a workaround because using variables in Nginx configuration is restricted. 

For security reasons it would be preferable to move to port 8080.


fniessink:
@jeroenma Is it ok to rebase this PR? Chromium was updated, failing the build.

jeroenma:
updated and yes sure

components/proxy/Dockerfile
```diff
@@ -5,9 +5,23 @@ FROM $IMAGE_NAME:$IMAGE_VERSION
 LABEL maintainer="Quality-time team <quality-time@ictu.nl>"
 LABEL description="Quality-time proxy"
 
+ENV NGINX_LISTEN 80
```
wkoot, line 8: How does this relate to the `PROXY_PORT` [env var](https://quality-time.readthedocs.io/en/latest/software.html#id10); do they both need to be set separately, or is the alternate proxy port rerouted by ingress?

components/proxy/qualitytime.conf.template
```diff
@@ -0,0 +1,35 @@
+server {
+    listen       ${NGINX_LISTEN};
+    server_name  localhost;
```
wkoot, line 3: This file seems to be a copy of `default.conf.template` (with these lines added) but the `access_log      off;` directive has been dropped. It would be preferable to template all functionalities within a single file

components/proxy/qualitytime.conf.template
```diff
@@ -0,0 +1,35 @@
...
+
+    add_header      Strict-Transport-Security "max-age=31536000;";
+    add_header      X-XSS-Protection "1; mode=block";
+    add_header      X-Content-Type-Options "nosniff";
+    add_header      X-Frame-Options "DENY";
+    gzip            on;
+    server_tokens   off;
+
+    location /api/v3/nr_measurements {
```
wkoot, line 12: All these `/api/v3` paths are now `/api/internal`

components/proxy/nginx.conf
```diff
@@ -0,0 +1,32 @@
+user  nginx;
```
wkoot, line 1: Does this file relate to [openshift/dockerfiles/nginx.conf](https://github.com/ICTU/quality-time/blob/master/openshift/dockerfiles/nginx.conf)? Can they be combined?

components/proxy/Dockerfile
```diff
@@ -5,9 +5,23 @@ FROM $IMAGE_NAME:$IMAGE_VERSION
...
+
 ENV API_SERVER_HOST api_server
 ENV API_SERVER_PORT 5001
 
-COPY *.template /etc/nginx/templates/
+RUN rm -f /etc/nginx/conf.d/default.conf
+RUN rm -f /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
+RUN rm -f /docker-entrypoint.d/20-envsubst-on-templates.sh
+
+COPY qualitytime.conf.template /srv/qualitytime.conf.template
```
wkoot, line 20: If the other `*.template` files are no longer used, they can be deleted in this PR

components/proxy/Dockerfile
```diff
@@ -5,9 +5,23 @@ FROM $IMAGE_NAME:$IMAGE_VERSION
...
 ENV FRONTEND_HOST frontend
 ENV FRONTEND_PORT 5000
+
 ENV API_SERVER_HOST api_server
 ENV API_SERVER_PORT 5001
 
-COPY *.template /etc/nginx/templates/
+RUN rm -f /etc/nginx/conf.d/default.conf
+RUN rm -f /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
+RUN rm -f /docker-entrypoint.d/20-envsubst-on-templates.sh
```
wkoot, line 18: Can this pre-existing script not provide what `90_fix_listen.sh` does?

components/proxy/Dockerfile
```diff
@@ -5,9 +5,23 @@ FROM $IMAGE_NAME:$IMAGE_VERSION
...
+
+COPY qualitytime.conf.template /srv/qualitytime.conf.template
+
+COPY 90_fix_listen.sh /docker-entrypoint.d/90_fix_listen.sh
+RUN chmod 0755 /docker-entrypoint.d/90_fix_listen.sh
+
+COPY nginx.conf /etc/nginx/nginx.conf
```
wkoot, line 25: Looks like [the 20-envsubst-on-templates.sh](https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/20-envsubst-on-templates.sh#L14) provides ways to modify `nginx.conf` directly, this might be easier to maintain

wkoot (commented):
Would the procedure detailed in [nginx docker docs](https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration-new-in-119) suffice?

components/proxy/Dockerfile
jeroenma: This file can be removed if we switch the 'LISTEN 80' statement to 'LISTEN 8080', we can remove also remove the 'user nginx' statement. Because user switching is only supported when Nginx is running with Root permissions.

components/proxy/Dockerfile
jeroenma: The variable NGINX_LISTEN is configuration of Nginx in the docker container, the PROXY_PORT can be used as input for for this variable.

components/proxy/nginx.conf
jeroenma: probably.

components/proxy/Dockerfile
jeroenma: I will check it out!

wkoot: Perhaps https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration-new-in-119 is easiest

jeroenma:
> Would the procedure detailed in [nginx docker docs](https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration-new-in-119) suffice?

That is a way better option, I fixed the issues local by using standard Nginx image and move LISTEN to Quality Time config. 
No changes needed to this image. So no changes needed, branch can be removed imho.

PR #8504, opened 2024-04-15
Prevent timeouts when accessing paginated APIs.

### LINKED ISSUE #8503: Allow metric collection to take longer than the measurement interval
By default, the collector wakes up every 20 seconds to collect the data for up to 30 metrics. If this takes more than 20 seconds a timeout occurs for the metrics that have not been finished. This ensures that the next batch starts on time. However, when the data collection requires accessing a paginated API, 20 seconds can be insufficient time to collect all the data.

Change the collector to allow more than 20 seconds for collecting the data for a metric while still starting a new batch every 20 seconds. This should be possible by using a larger timeout in `SourceCollector.__safely_get_source_responses()` combined with using a fire-and-forget approach in `Collector.collect_metrics`. Note that currently a new session is created every 20 seconds that is garbage collected when the reference disappears, closing any open connections. Test whether we can use one session that lives indefinitely. 

This issue blocks #7171.

fniessink:
When accessing a source takes more than the collection interval (20 seconds by default), a timeout would occur and the metric could not be measured. This happened mostly with sources that have a paginated API and a lot of data. Prevent the timeout by allowing data collection to take more than the collection interval.

Closes #8503.


components/collector/tests/base_collectors/test_collector.py
```diff
@@ -62,6 +63,8 @@ async def _fetch_measurements(self, mock_async_get_request, number=1, side_effec
             async with aiohttp.ClientSession() as session:
                 for _ in range(number):
                     await self.collector.collect_metrics(session)
+                    # Wait for the created tasks to finish
+                    await asyncio.gather(*asyncio.all_tasks() - {asyncio.current_task()})
```
wkoot: This looks magically cool, I guess it's fine within test code but you break it you buy it :D

components/collector/tests/base_collectors/test_collector.py
```diff
@@ -178,26 +181,22 @@ async def test_fetch_with_post_error(self):
...
         ):
             await quality_time_collector.collect()
-        post.assert_called_once_with(
-            self.database,
-            {"has_error": False, "sources": [self._source()], "metric_uuid": "metric_uuid", "report_uuid": "report1"},
-        )
+        # Wait for the created tasks to finish
+        await asyncio.gather(*asyncio.all_tasks() - {asyncio.current_task()})
+        measurement = next(self.database["measurements"].find({"metric_uuid": "metric_uuid"}))
+        self.assertEqual("metric_uuid", measurement["metric_uuid"])
```
wkoot: Why assert that the found object is as queried, shouldn't `next()` already `raise StopIteration` if it doesn't match?

components/collector/src/base_collectors/collector.py
```diff
@@ -43,41 +40,37 @@ def record_health() -> None:
...
-                collection_timer.duration,
-                sleep_duration,
-            )
-            await asyncio.sleep(sleep_duration)
+        # The TCPConnector has limit 0, meaning unlimited, because aiohttp only closes connections when the response
+        # is closed. But due to the architecture of the collector (first collect all the responses, then parse them)
+        # the responses are kept around relatively long, and hence the connections too. To prevent time-outs while
+        # aiohttp waits for connections to become available we don't limit the connection pool size.
+        async with aiohttp.ClientSession(
+            connector=aiohttp.TCPConnector(limit=0, ssl=False),
```
wkoot, line 48: What does it mean to set `ssl=False` here, is that just a process internal?

components/collector/src/base_collectors/collector.py
```diff
@@ -43,41 +40,37 @@ def record_health() -> None:
...
+            raise_for_status=True,
+            timeout=aiohttp.ClientTimeout(total=120),
```
wkoot: Shouldn't this 120 be defined as a constant (also for usage in tests), or even an env var?

components/collector/src/base_collectors/collector.py
```diff
@@ -43,41 +40,37 @@ def record_health() -> None:
...
+            trust_env=True,
+        ) as session:
+            while True:
+                self.__log_tasks("waking up")
+                self.record_health()
```
wkoot, line 55: Does this health check still make sense in this location, as the actual processing is now deferred?

components/collector/src/base_collectors/collector.py
```diff
@@ -43,41 +40,37 @@ def record_health() -> None:
...
         """Collect measurements for metrics, prioritizing edited metrics."""
         reports = get_reports(self.database)
         metrics = get_metrics_from_reports(reports)
         next_fetch = datetime.now(tz=UTC) + timedelta(seconds=config.MEASUREMENT_FREQUENCY)
-        tasks: list[Coroutine] = []
+        created_tasks = 0
         for metric_uuid, metric in self.__sorted_by_edit_status(cast(JSONDict, metrics)):
-            if len(tasks) >= config.MEASUREMENT_LIMIT:
-                break
             if self.__should_collect(metric_uuid, metric):
```
wkoot: Why not change this into `if not should collect: continue` and skip a level of indentation?

components/collector/src/base_collectors/collector.py
```diff
@@ -43,41 +40,37 @@ def record_health() -> None:
...
-                tasks.append(self.collect_metric(session, metric_uuid, metric, next_fetch))
-        await asyncio.gather(*tasks)
+                task = asyncio.create_task(self.collect_metric(session, metric_uuid, metric, next_fetch))
+                self.tasks.add(task)
+                created_tasks += 1
+                task.add_done_callback(self.tasks.discard)
```
wkoot: Does this not also discard errors and stack traces?

components/collector/src/base_collectors/collector.py
```diff
@@ -108,3 +101,8 @@ def __should_collect(self, metric_uuid: str, metric: dict) -> bool:
         metric_edited = self.__previous_metrics.get(metric_uuid) != metric
         metric_due = self.next_fetch.get(metric_uuid, datetime.min.replace(tzinfo=UTC)) <= datetime.now(tz=UTC)
         return metric_edited or metric_due
+
+    def __log_tasks(self, event: str, warning_threshold: int = 0) -> None:
+        """Log the number of running tasks."""
+        level = logging.WARNING if len(self.tasks) > warning_threshold else logging.INFO
+        logging.log(level, "Collector %s: %d task(s) currently running", event, len(self.tasks))
```
wkoot: Can we perhaps stick this in a separate logger, so it doesn't clutter the regular logs?

components/collector/tests/base_collectors/test_collector.py
fniessink: It's what I found when looking for ways to wait for tasks that we have no reference to (created with `asyncio.create_task`) to finish. But after that I read elsewhere that we have to keep a strong reference to async tasks in any case, because there are rare circumstances where they could be garbage collected if we don't. So I added a set of tasks to the `Collector` class to keep track of running tasks. I guess I could use that set here instead of this magic.

components/collector/tests/base_collectors/test_collector.py
fniessink: Correct. I'll change the assert to check the connection error that happens due to the session being closed immediately.

components/collector/src/base_collectors/collector.py
fniessink: It turns off strict checking of SSL-certificates, see https://docs.aiohttp.org/en/stable/client_advanced.html#ssl-control-for-tcp-sockets.

components/collector/src/base_collectors/collector.py
fniessink: I'll make it an environment variable.

components/collector/src/base_collectors/collector.py
fniessink: Yes, because as long as the collector wakes up every 20 seconds it should be healthy. 

However, we could consider reporting the collector unhealthy if the number of running tasks exceeds a certain threshold. I'll open an issue for that.

components/collector/src/base_collectors/collector.py
fniessink: 👍 Will do.

components/collector/src/base_collectors/collector.py
fniessink: No, it just calls `self.tasks.discard(task)` when the task is done.

components/collector/src/base_collectors/collector.py
fniessink: We could, but that seems outside the scope of this issue.

PR #8532, opened 2024-04-23
Add metrics not supported by subject type.

### LINKED ISSUE #8176: Add a metric that is not listed as supported by the subject type
# User story

As a user I want to add a metric type to a subject that is not listed as supported by the subject type.

## Acceptance criteria

- [ ] When adding a metric using the "Add metric" button, the user can indicate that they want to see all metric types 
- [ ] After indicating that the user wants to add an unlisted metric, the user can choose all metric types.

## Assumptions

- Quality-time will not check that added metrics are listed as supported by the subject type (this is already not the case).

# Tasks

### Frontend

- [ ] Add a checkbox "Show all metric types" to the "Add metric dropdown menu"

## Documentation

- [ ] Update changelog

# Related issues

- #7992.

fniessink:
When adding a metric to a subject, add an option to choose from all metric types in addition to the metric types officially supported by the subject type.

Closes #8176.

### DIFF
--- a/components/frontend/src/semantic_ui_react_wrappers/Checkbox.js
+++ b/components/frontend/src/semantic_ui_react_wrappers/Checkbox.js
@@ -1,0 +1,9 @@
+import { useContext } from "react"
+import { Checkbox as SemanticUICheckbox } from "semantic-ui-react"
+
+import { DarkMode } from "../context/DarkMode"
+import { addInvertedClassNameWhenInDarkMode } from "./dark_mode"
+
+export function Checkbox(props) {
+    return <SemanticUICheckbox {...addInvertedClassNameWhenInDarkMode(props, useContext(DarkMode))} />
+}
--- a/components/frontend/src/widgets/Button.js
+++ b/components/frontend/src/widgets/Button.js
@@ -2,7 +2,7 @@ import { array, bool, func, string } from "prop-types"
 import { useRef, useState } from "react"
 import { Icon, Input } from "semantic-ui-react"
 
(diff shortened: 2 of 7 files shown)


components/frontend/src/semantic_ui_react_wrappers/Checkbox.js
```diff
@@ -0,0 +1,9 @@
+import { useContext } from "react"
+import { Checkbox as SemanticUICheckbox } from "semantic-ui-react"
+
+import { DarkMode } from "../context/DarkMode"
+import { addInvertedClassNameWhenInDarkMode } from "./dark_mode"
+
+export function Checkbox(props) {
+    return <SemanticUICheckbox {...addInvertedClassNameWhenInDarkMode(props, useContext(DarkMode))} />
```
wkoot, line 8: Interesting that we didn't use checkboxes at all yet, I didn't realise!

components/frontend/src/widgets/Button.js
```diff
@@ -99,28 +101,38 @@ export function AddDropdownButton({ itemSubtypes, itemType, onClick }) {
...
-                                    value={query}
-                                />
-                            </>
+                        <Dropdown.Divider />
+                        <Input
+                            className="search"
+                            focus
+                            icon="search"
+                            iconPosition="left"
+                            onBlur={(event) => event.stopPropagation()}
```
wkoot: Does this mean that the filter option is blurred when selecting from all? What if you want to search through all the possible metrics?

fniessink: It's to prevent the input from being cleared when the user tabs to the checkbox.

PR #8554, opened 2024-04-24
Show issues in the dashboard.

### LINKED ISSUE #8222: Show unresolved issues in the dashboard
Acceptance criteria:

- [x] Add an issue card to the dashboard showing the number of issues per issue status (open, in progress, done) as a table ~donut chart~.
- [x] ~When showing multiple dates, use bar charts.~
- [x] Make the card hideable in the Settings panel.
- [x] When clicking the card, only show metrics with issues.
- [x] Only show the card when the report has a configured issue tracker.
 
Changes to original idea:
- A table instead of a donut chart so the difference with metrics is more ovbious
- No bar charts because the historic information about number of issues per status is not stored as measurements. In theory we can retrieve old reports to collect that info, but this seemed a bit too much work for this story.
- Also make the Legend card hideable.

fniessink:
- When a report has a configured issue tracker, show a card in the dasbboard with the number of issues per issue status (open, in progress, done). The card can be hidden via the Settings panel.
- Allow for hiding the legend card via the Settings panel.

Closes #8222.

### DIFF
--- a/components/frontend/src/dashboard/IssuesCard.js
+++ b/components/frontend/src/dashboard/IssuesCard.js
@@ -1,0 +1,63 @@
+import "./IssuesCard.css"
+
+import { bool, func } from "prop-types"
+
+import { Card, Header, Label, Table } from "../semantic_ui_react_wrappers"
+import { reportPropType } from "../sharedPropTypes"
+import { capitalize, ISSUE_STATUS_COLORS } from "../utils"
+
+function issueStatuses(report) {
+    // The issue status is unknown when the issue was added recently and the status hasn't been collected yet
+    // or when collecting the issue status from the issue tracker failed
+    const statuses = { todo: 0, doing: 0, done: 0, unknown: 0 }
+    Object.values(report.subjects).forEach((subject) => {
+        Object.values(subject.metrics).forEach((metric) => {
+            let nrIssuesWithKnownStatus = 0
+            const issueStatus = metric.issue_status ?? []
+            issueStatus.forEach((issue) => {
+                if (issue.status_category) {
+                    statuses[issue.status_category] += 1
+                    nrIssuesWithKnownStatus += 1
+                }
+            })
+            const nrIssues = metric.issue_ids?.length ?? 0
+            statuses["unknown"] += nrIssues - nrIssuesWithKnownStatus
+        })
+    })
+    return statuses
+}
+issueStatuses.propTypes = {
+    report: reportPropType,
+}
+
+export function IssuesCard({ onClick, report, selected }) {
+    const statuses = issueStatuses(report)
+    const tableRows = Object.keys(statuses).map((status) => (
+        <Table.Row key={status}>
+            <Table.Cell>{capitalize(status)}</Table.Cell>
+            <Table.Cell textAlign="right">
+                <Label size="small" color={ISSUE_STATUS_COLORS[status]}>
+                    {statuses[status]}
+                </Label>
+            </Table.Cell>
+        </Table.Row>
+    ))
+    const color = selected ? "blue" : null
+    return (
+        <Card className="issues" color={color} onClick={onClick} onKeyPress={onClick} tabIndex="0">
+            <Card.Content>
+                <Header as="h3" color={color} textAlign="center">
+                    {"Issues"}
+                </Header>
+                <Table basic="very" compact="very" size="small">
+                    <Table.Body>{tableRows}</Table.Body>
+                </Table>
+            </Card.Content>
+        </Card>
+    )
+}
+IssuesCard.propTypes = {
+    onClick: func,
+    report: reportPropType,
+    selected: bool,
+}
--- a/components/frontend/src/dashboard/IssuesCard.test.js
+++ b/components/frontend/src/dashboard/IssuesCard.test.js
@@ -1,0 +1,42 @@
+import { render, screen } from "@testing-library/react"
+
+import { IssuesCard } from "./IssuesCard"
+
+const report = {
+    subjects: {
+        subject_uuid: {
+            metrics: {
+                metric_uuid: {
+                    issue_ids: ["ID-1"],
+                    issue_status: [{ status_category: "doing" }],
+                },
+                another_metric_uuid: {
+                    issue_ids: ["ID-2", "ID-3"],
+                    issue_status: [{ connection_error: "oops" }],
+                },
+            },
+        },
+    },
+}
+
+function renderIssuesCard({ selected = false } = {}) {
+    render(<IssuesCard report={report} selected={selected} />)
+}
+
+it("shows the correct title", () => {
+    renderIssuesCard()
+    expect(screen.getByText(/Issues/)).toBeInTheDocument()
+})
+
+it("shows the title in blue when selected", () => {
+    renderIssuesCard({ selected: true })
+    expect(screen.getByText(/Issues/)).toHaveClass("blue")
+})
+
+it("shows the number of issues", () => {
+    renderIssuesCard()
+    expect(screen.getByRole("row", { name: "Todo 0" })).toBeInTheDocument()
+    expect(screen.getByRole("row", { name: "Doing 1" })).toBeInTheDocument()
+    expect(screen.getByRole("row", { name: "Done 0" })).toBeInTheDocument()
+    expect(screen.getByRole("row", { name: "Unknown 2" })).toBeInTheDocument()
+})
--- a/components/frontend/src/issue/IssueStatus.js
+++ b/components/frontend/src/issue/IssueStatus.js
@@ -3,7 +3,7 @@ import TimeAgo from "react-timeago"
 
 import { Label, Popup } from "../semantic_ui_react_wrappers"
 import { issueStatusPropType, metricPropType, settingsPropType, stringsPropType } from "../sharedPropTypes"
-import { getMetricIssueIds } from "../utils"
+import { getMetricIssueIds, ISSUE_STATUS_COLORS } from "../utils"
 import { HyperLink } from "../widgets/HyperLink"
 import { TimeAgoWithDate } from "../widgets/TimeAgoWithDate"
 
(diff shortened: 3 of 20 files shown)


components/frontend/src/issue/IssueStatus.test.js
```diff
@@ -87,7 +87,7 @@ it("displays the status category done", () => {
 
 it("displays a missing status category as todo", () => {
     renderIssueStatus()
-    expect(screen.getByText(/123/).className).toContain("grey")
+    expect(screen.getByText(/123/).className).toContain("basic")
```
wkoot: But in `ISSUE_STATUS_COLORS`, the `todo: "grey"` value remained; shouldn't that mean this doesn't change? The test title says that it is selecting category todo

components/frontend/src/report/Report.js
```diff
@@ -33,8 +33,11 @@ export function Report({
 }) {
     function navigate_to_subject(event, subject_uuid) {
         event.preventDefault()
-        document.getElementById(subject_uuid).scrollIntoView()
-        window.scrollBy(0, 163) // Correct for menubar and subject title margin
+        const element = document.getElementById(subject_uuid)
+        if (element) {
+            element.scrollIntoView()
+            window.scrollBy(0, 163) // Correct for menubar and subject title margin
```
wkoot, line 39: Yikes that's a nasty hardcoded offset, are the menubar and title margins also exactly set by pixel count?

components/frontend/src/report/Report.test.js
```diff
@@ -164,3 +164,26 @@ it("hides subjects if empty", async () => {
...
+
+it("navigates to subject", async () => {
+    const mockScroll = jest.fn()
+    window.HTMLElement.prototype.scrollIntoView = mockScroll
+    const mockScrollBy = jest.fn()
+    window.scrollBy = mockScrollBy
+    renderReport({ reportToRender: report })
+    fireEvent.click(screen.getAllByText(/Subject title/)[0])
+    expect(mockScroll).toHaveBeenCalledWith()
+    expect(mockScrollBy).toHaveBeenCalledWith(0, 163)
```
wkoot, line 176: hahaha, well it works!

components/frontend/src/dashboard/IssuesCard.js
```diff
@@ -0,0 +1,61 @@
...
+            let nrIssuesWithKnownStatus = 0
+            const issueStatus = metric.issue_status ?? []
+            issueStatus.forEach((issue) => {
+                if (issue.status_category) {
+                    statuses[issue.status_category] += 1
+                    nrIssuesWithKnownStatus += 1
+                }
+            })
+            const nrIssues = metric.issue_ids?.length ?? 0
+            statuses["unknown"] += nrIssues - nrIssuesWithKnownStatus
```
wkoot, line 24: Nice

components/frontend/src/dashboard/IssuesCard.js
```diff
@@ -0,0 +1,61 @@
...
+        <Table.Row key={status}>
+            <Table.Cell>{capitalize(status)}</Table.Cell>
+            <Table.Cell textAlign="right">
+                <Label size="small" color={ISSUE_STATUS_COLORS[status]}>
+                    {statuses[status]}
+                </Label>
+            </Table.Cell>
+        </Table.Row>
+    ))
+    const color = selected ? "blue" : null
```
wkoot, line 45: Do we not have a 'selected color' preset, i.e. for tweaking dark/light mode differences? Or does this work for both

components/frontend/src/dashboard/IssuesCard.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+it("shows the title in blue when selected", () => {
+    renderIssuesCard({ selected: true })
+    expect(screen.getByText(/Issues/)).toHaveClass("blue")
+})
+
+it("shows the number of issues", () => {
+    renderIssuesCard()
+    expect(screen.getByText(/1/)).toBeInTheDocument() // 1 doing
+    expect(screen.getByText(/2/)).toBeInTheDocument() // 2 Unknown
```
wkoot: Hmm, isn't it a bit too optimistic to only query for the number?

components/frontend/src/report/ReportDashboard.js
```diff
@@ -78,14 +79,30 @@ export function ReportDashboard({ dates, measurements, onClick, onClickTag, relo
...
+        extraCards.push(
+            <IssuesCard
+                key="issues"
+                onClick={() => settings.metricsToHide.set(selected ? "none" : "no_issues")}
+                report={report}
+                selected={selected}
+            />,
+        )
+    }
+    if (!settings.hiddenCards.includes("legend")) {
```
wkoot: Ew... "if hidden does not include". But it is what it is I guess

components/frontend/src/report/ReportDashboard.js
```diff
@@ -78,14 +79,30 @@ export function ReportDashboard({ dates, measurements, onClick, onClickTag, relo
...
+        extraCards.push(<LegendCard key="legend" />)
+    }
     return (
         <CardDashboard
-            cards={subjectCards.concat(tagCards.concat([<LegendCard key="legend" />]))}
+            cards={subjectCards.concat(tagCards.concat(extraCards))}
```
wkoot, line 105: I'm surprised that there is no change needed to the `CardDashboard.cards` prop. It's undesirable to specify more specific typing than `array` ?

components/frontend/src/issue/IssueStatus.js
```diff
@@ -119,7 +119,7 @@ prefixName.propType = {
 }
 
 function issueLabel(issueStatus, settings, error) {
-    const color = error ? "red" : { todo: "grey", doing: "blue", done: "green" }[issueStatus.status_category ?? "todo"]
+    const color = error ? "red" : ISSUE_STATUS_COLORS[issueStatus.status_category ?? "unknown"]
```
wkoot, line 123: What does colour "unknown" mean? Why aren't unknown and error part of the `ISSUE_STATUS_COLORS`?

components/frontend/src/utils.js
```diff
@@ -193,28 +202,39 @@ sortWithLocaleCompare.propTypes = {
...
+function hideMetric(metric, metricsToHide, hiddenTags) {
+    const hideBecauseNoActionNeeded =
+        metricsToHide === "no_action_needed" && ["target_met", "debt_target_met", "informative"].includes(metric.status)
+    const hideBecauseNoIssues = metricsToHide === "no_issues" && !metric.issue_ids
+    const hideBecauseTagIsHidden =
+        hiddenTags?.length > 0 &&
+        hiddenTags?.filter((hiddenTag) => metric.tags?.includes(hiddenTag)).length >= metric.tags?.length
+    return hideBecauseNoActionNeeded || hideBecauseNoIssues || hideBecauseTagIsHidden
+}
+visibleMetrics.propTypes = {
```
wkoot: This propType still needs to be changed :)

components/frontend/src/utils.js
```diff
@@ -193,28 +202,39 @@ sortWithLocaleCompare.propTypes = {
     strings: stringsPropType,
 }
 
+function hideMetric(metric, metricsToHide, hiddenTags) {
+    const hideBecauseNoActionNeeded =
+        metricsToHide === "no_action_needed" && ["target_met", "debt_target_met", "informative"].includes(metric.status)
+    const hideBecauseNoIssues = metricsToHide === "no_issues" && !metric.issue_ids
+    const hideBecauseTagIsHidden =
+        hiddenTags?.length > 0 &&
+        hiddenTags?.filter((hiddenTag) => metric.tags?.includes(hiddenTag)).length >= metric.tags?.length
```
wkoot, line 211: Okay, so filtering tags is additive and not combinative. Think it makes sense, perhaps good to document or tooltip?

components/frontend/src/issue/IssueStatus.test.js
fniessink: I forgot to change the test title. Also, the Label doesn't get a color when its status is missing, I'll improve the assert to make that more clear.

components/frontend/src/report/Report.js
fniessink: Apparently, because it seems to work. Would be nice to be able to ask the menubar and title for their height, but I don't know how to do that. That rabbit hole is out of scope for this issue :-)

components/frontend/src/dashboard/IssuesCard.js
fniessink: Blue isn't actually a real color, but a class that gets changed by Semantic UI React to different shades of blue depending on the dark/light mode setting. I guess I could change `const color = ...` to `const colorClassName = ...`, but Semantic UI React calls these props `color` everywhere so that I don't think that's much of an improvement.

components/frontend/src/dashboard/IssuesCard.test.js
fniessink: Yeah, should really query for a table row. Let me see if I can improve this.

fniessink: This is better:

```js
it("shows the number of issues", () => {
    renderIssuesCard()
    expect(screen.getByRole("row", {name: "Todo 0"})).toBeInTheDocument()
    expect(screen.getByRole("row", {name: "Doing 1"})).toBeInTheDocument()
    expect(screen.getByRole("row", {name: "Done 0"})).toBeInTheDocument()
    expect(screen.getByRole("row", {name: "Unknown 2"})).toBeInTheDocument()
})
```

components/frontend/src/report/ReportDashboard.js
fniessink: I'll add an `excludes()` to ` useArrayURLSearchQuery`, that's just a oneliner and should help with readability.

components/frontend/src/report/ReportDashboard.js
fniessink: I've changed the prop type to `arrayOf(Card)`.

components/frontend/src/issue/IssueStatus.js
fniessink: `ISSUE_STATUS_COLORS` does include "unknown" now. Unknown means the issue status is unknown, because the status hasn't been collected yet or the connection with the issue tracker failed. I'll add a comment.

components/frontend/src/utils.js
fniessink: Oops. Weird I didn't see any error.

components/frontend/src/utils.js
fniessink: It works quite intuitively in the UI I think. Not sure it's necessary to try and explain it to the users.

PR #8586, opened 2024-04-28
Fix API-server documentation

### LINKED ISSUE #8553: Add to documentation that API-server needs internet connection
change text and diagram

fniessink:
Include outside connections made by the API-server in the software documentation.

Fixes #8553.


docs/src/components-dark.dot
```diff
@@ -1,38 +1,46 @@
-digraph {
-  graph [bgcolor=transparant, fontname="SFMono-Regular"]
+ digraph {
```
wkoot: Was this extra space at the start intentional? The `components-light.dot` does not have it! (this is about the only thing I can say about this PR, I guess the generated image is fine ;))

fniessink: Not intentional! Thanks!

PR #8588, opened 2024-04-28
Change default value of branch parameters to main

### LINKED ISSUE #8045: Change default value for the branch parameter to 'main'
Tasks:

- [x] Add migration script to change branches without a value to 'master'. Old sources have branch parameters without value. Apparently, the default value of parameters has not always been inserted into reports. 
- [x] Also investigate whether the `Branch` and `Branches` parameter can be combined. In that case no default value would be needed.

**Combining the `Branch` and `Branches` parameters**

The `Branch` parameter is only used by `GitLab` and `AzureDevOps`, both for the metric: `source_up_to_dateness`.

The `Branches` parameters is only used by `GitLab` for the metric: `pipeline_duration`.

Note that `SonarQube` also has a branch parameter, but it's not derived from `Branch`, but rather a separate `StringParameter`. The analysis below only covers the use of `Branch` by `GitLab` and `AzureDevOps`.

To decide whether `Branch` and `Branches` can be combined it's important to look at the default values of the parameters. `Branches` has as default value "all branches". `Branch` has `master` (after this issue is completed: `main`) as default value. 

When configuring `GitLab` or `Azure DevOps` as source for the `source_up_to_dateness` metric, users can measure the up-to-dateness of files and of pipelines. The `Branch` parameter is only used in case of the files option: users can specify for which branch the up-to-dateness of the file should be measured. 

If we would replace the `Branch` parameter with the `Branches` parameter, *Quality-time* is basically saying that by default users will want to track the up-to-dateness of files across all branches. We haven't had that request from users.

Alternatively, we could change the default value of the `Branches` parameter to `main`. That would impact only the `pipeline_duration` metric when used with `GitLab` as source. Users that want to measure the duration of all pipelines regardless of branch would have to delete `main` from the parameter value. It seems more likely that users would want to measure the duration of all pipelines instead of just the main branch.

Getting rid of the `Branch` parameter would save five lines of code in the data model.

Conclusion: merging the two parameters would worsen user experience in return for a very small reduction of code size. It's not worth it.

fniessink:
Make the default value of branch parameters `main` instead of `master`.

Note: no migration code is needed because Quality-time only uses the default value of the branch parameter upon creation of a source. Changing the default value of the branch parameter does not impact existing sources.

Closes #8045.

### DIFF
--- a/components/collector/tests/source_collectors/azure_devops/test_source_up_to_dateness.py
+++ b/components/collector/tests/source_collectors/azure_devops/test_source_up_to_dateness.py
@@ -37,7 +37,7 @@ class AzureDevopsSourceUpToDatenessTest(AzureDevopsTestCase):
         self.assert_measurement(
             response,
             value=self.expected_age,
-            landing_url=f"{self.url}/_git/repo?path=README.md&version=GBmaster",
+            landing_url=f"{self.url}/_git/repo?path=README.md&version=GBmain",
         )
 
     async def test_age_of_pipeline(self):
--- a/components/collector/tests/source_collectors/sonarqube/base.py
+++ b/components/collector/tests/source_collectors/sonarqube/base.py
@@ -16,8 +16,8 @@ class SonarQubeTestCase(SourceCollectorTestCase):
         """Extend to set up the SonarQube source fixture and some URLs."""
         super().setUp()
         self.set_source_parameter("component", "id")
-        self.issues_landing_url = "https://sonarqube/project/issues?id=id&branch=master&resolved=false"
-        self.metric_landing_url = "https://sonarqube/component_measures?id=id&branch=master&metric={0}"
+        self.issues_landing_url = "https://sonarqube/project/issues?id=id&branch=main&resolved=false"
+        self.metric_landing_url = "https://sonarqube/component_measures?id=id&branch=main&metric={0}"
 
     @staticmethod
     def entity(  # noqa: PLR0913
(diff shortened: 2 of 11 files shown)


components/collector/tests/source_collectors/azure_devops/test_source_up_to_dateness.py
```diff
@@ -37,7 +37,7 @@ async def test_age_of_file(self):
         self.assert_measurement(
             response,
             value=self.expected_age,
-            landing_url=f"{self.url}/_git/repo?path=README.md&version=GBmaster",
+            landing_url=f"{self.url}/_git/repo?path=README.md&version=GBmain",
```
wkoot, line 40: Does this come from the change in `shared_data_model.parameters.Branch`; is this the only instance of that class?

fniessink: Yes and yes :-)

Azure DevOps only uses the branch parameter for the source-up-to-dateness metric. There are other branch parameters used by Azure DevOps, namely branches to ignore and target branches to include, but they are lists.

wkoot: So, can the second part of #8045 be achieved?

fniessink: Oops. Forgot about that part. I'll have a look!

fniessink: I don't think it's worth it. See analysis in #8045. Agree?

wkoot: Yep makes sense! At best we could make `SonarQube` use `Branch`.. in a different story ;) Let's merge this?

PR #8641, opened 2024-05-07
Replace bump2version with bump-my-version

### LINKED ISSUE #4619: Replace bump2version with bump-my-version
Bump2version is no longer maintained.

See https://github.com/c4urself/bump2version

Recommended alternative: https://pypi.org/project/bump-my-version/
https://github.com/callowayproject/bump-my-version

fniessink:
Replace the unmaintained bump2version with bump-my-version.

Closes #4619.


release/requirements/requirements-dev.txt
```diff
@@ -32,16 +51,148 @@ pip-tools==7.4.1 \
...
+    --hash=sha256:3e163f254bef5a03b146397d7c1963bd3e2812f0964bb9a24e6ec761fd28db63 \
+    --hash=sha256:aa64ad242a462c5ff0363a7b9cfe696c20d55d9fc60c11fd8e632d064804d305
+    # via questionary
+pydantic==2.7.1 \
+    --hash=sha256:e029badca45266732a9a79898a15ae2e8b14840b1eabbb25844be28f0b33f3d5 \
+    --hash=sha256:e9dbb5eada8abe4d9ae5f46b9939aead650cd2b68f249bb3a8139dbe125803cc
+    # via
+    #   bump-my-version
+    #   pydantic-settings
+pydantic-core==2.18.2 \
```
wkoot, line 64: Interesting that so many new requirements are needed for release, is it all due to `bump-my-version` or was this just not recently updated?

release/release.py
```diff
@@ -11,20 +11,21 @@
...
     """Return the current version."""
-    command = f"bump2version --list --dry-run --allow-dirty --no-configured-files {PART} .bumpversion.cfg"
-    try:
-        output = subprocess.check_output(command.split(" "), stderr=subprocess.STDOUT, text=True)
-    except (subprocess.CalledProcessError, FileNotFoundError):
-        return "<unknown>"
-    return [line for line in output.split("\n") if line.startswith("current_version")][0].split("=")[1]
+    root = get_release_folder().parent
+    repo = git.Repo(root)
+    latest_tag = sorted(repo.tags, key=lambda tag: tag.commit.committed_datetime)[-1]
```
wkoot: This doesn't look like a safe assumption, perhaps only filter on tag names that resemble a semantic version or those which start with the letter "v"?

release/release.py
```diff
@@ -100,37 +101,18 @@ def failed_preconditions_changelog(bump: str, root: pathlib.Path) -> list[str]:
...
-                ".bumpversion-rc.cfg",
-                "--message",
-                message,
-                "--tag-message",
-                message,
-                "--allow-dirty",
-                rc_bump,
-            ]
-        )
+        bump = bump.split("-")[1]  # Create a patch, minor, or major release candidate
```
wkoot: This would be problematic when the version includes dashes, why not `bump = bump[3:]` since we just want to exclude the prefix "rc-"?

release/release.py
```diff
@@ -100,37 +101,18 @@ def failed_preconditions_changelog(bump: str, root: pathlib.Path) -> list[str]:
...
     elif bump == "drop-rc":
-        new_version = current_version.split("-")[0]
-        commands.append(["bump2version", "--new-version", new_version, PART])
-    else:
-        commands.append(["bump2version", bump])
-    for command in commands:
-        subprocess.run(tuple(command), check=True)
+        bump = "pre_release_label"  # Bump the pre-release label from "rc" to "final" (which is optional and omitted)
```
wkoot, line 120: I don't understand the comment; in order to delete the release candidate, we make the rc final? Or does it mean that the process doesn't do anything when the value "final" is passed; if so, how is the rc cleaned up, and/or can't we just short-circuit `return` here?

release/pyproject.toml
```diff
@@ -11,3 +11,67 @@ dev = [
...
+serialize = [
+    "{major}.{minor}.{patch}-{pre_release_label}.{pre_release_number}",
+    "{major}.{minor}.{patch}",
+]
+commit = true
+tag = true
+
+[tool.bumpversion.parts.pre_release_label]
+values = ["rc", "final"]
+optional_value = "final"
```
wkoot, line 37: Does "optional" mean that it's the default?

release/requirements/requirements-dev.txt
fniessink: They are new indirect dependencies of bump-my-version.

release/release.py
fniessink: Yeah, unfortunately bump-my-version has no easy way to return the latest version like bump2version had. So that's why I needed another way to get the latest version. I'll add a few more safe guards.

release/release.py
fniessink: `bump` isn't a version number here, but the bump argument specified on the command line to, for example, `minor` or `rc-minor`. But, I'll make it more robust just in case we add more bump options such as `rc-minor-include-kubernetes`.

release/release.py
fniessink: All releases without `rc` are `final`, but the final part of the version number is optional. Apparently this is how bump-my-version implements parts of the version number that are not always present.

fniessink: I'll add a link to the bump-my-version docs.

release/pyproject.toml
fniessink: See previous comment.

PR #8660, opened 2024-05-10
Show metrics needing measurement in the UI.

### LINKED ISSUE #3134: Show whether the latest measurement of metric has been done with the current configuration of its source(s)
When the source(s) of a metric have been changed, but the new measurement is not available yet:
- [x] Add a `hourglass` icon to the measurement 
- [x] Add a yellow label to the measurement
- [x] Add a warning popup: 
"The source configuration of this metric has been changed.
Collecting the new measurement data may take a few moments.
The metric was last measured 2 minutes ago (23-12-2023 13:10)
The configurations updated 20 seconds ago (22-12-2023 09:58)"

How to determine whether the metric was measured after it was edited? Add timestamps to metrics and sources in the report when they are changed. Compare the timestamps with the end of the most recent measurement.

fniessink:
In the UI, show a metric as needing a measurement because the source parameters were recently changed.

Implemented by:
- In the collector: adding a hash of the source parameters to new measurements (and removing it from old measurements).
- In the API-server: compare the hash of the latest measurement with the hash of the current configuration when adding recent measurements to a report and mark the metric as outdated.
- In the UI: show metrics as needing a measurement if the last measurement is outdated.

Closes #3134.


components/shared_code/src/shared/model/measurement.py
```diff
@@ -294,3 +294,11 @@ def summarize(self) -> dict[str, str | dict[str, str | None]]:
             "start": self["start"],
             "end": self["end"],
         }
+
+    def summarize_latest(self) -> Self:
+        """Return a summary of this measurement, given that it is the latest measurement."""
+        if parameter_hash := self.get("source_parameter_hash"):  # pragma: no feature-test-cover
+            if parameter_hash != self.metric.source_parameter_hash():
+                self["outdated"] = True
```
wkoot, line 302: Does it not make more sense to make `outdated` a property, and unset `source_parameter_hash` in the setter method of the prop? This in order to avoid implementing deletion of the `source_parameter_hash` in various places later

components/shared_code/src/shared/model/metric.py
```diff
@@ -159,9 +160,14 @@ def summarize(self, measurements: list[Measurement], **kwargs) -> dict:
...
+        summary["latest_measurement"] = latest_measurement.summarize_latest() if latest_measurement else None
         summary["recent_measurements"] = [measurement.summarize() for measurement in measurements]
         if latest_measurement:
             summary["issue_status"] = self.issue_statuses(latest_measurement)
         summary.update(kwargs)
         return summary
+
+    def source_parameter_hash(self) -> str:  # pragma: no feature-test-cover
+        """Return a hash of the source parameters that can be used to check for changed parameters."""
+        source_parameters = [source.get("parameters") for source in self.sources]
```
wkoot: Is the assumption that these objects are always listed in the same order, safe?

components/shared_code/src/shared/database/measurements.py
```diff
@@ -27,6 +27,9 @@ def latest_successful_measurement(
 
 def insert_new_measurement(database: Database, measurement: Measurement) -> Measurement:
     """Insert a new measurement."""
+    if latest := latest_measurement(database, measurement.metric):  # pragma: no feature-test-cover
```
wkoot, line 30: The name `latest` was a bit confusing, although I'm not sure that naming it `previous` would make things better :(

components/shared_code/tests/shared/model/test_metric.py
```diff
@@ -25,91 +25,91 @@ def create_metric(self, **kwargs) -> Metric:
...
-        """Test that a minimal metric returns a summary."""
+    def test_summarize_metric_with_an_outdated_measurement(self):
+        """Test that the metric is marked outdated when the source parameters have changed."""
         metric = self.create_metric()
         measurement_timestamp = iso_timestamp()
         measurement = Measurement(
             metric,
             count={"value": 1, "start": measurement_timestamp},
             status="target_met",
+            source_parameter_hash="old hash",
```
wkoot, line 102: That's not actually changing the source parameters, but a sneaky shortcut :)))

components/shared_code/src/shared/model/metric.py
fniessink: Good question. I think the order may change if the user reorders sources. I'll add sorting by source UUID before generating the hash.

components/shared_code/src/shared/database/measurements.py
fniessink: `lastest_measurement_soon_to_be_the_previous_measurement` :-)

components/shared_code/tests/shared/model/test_metric.py
fniessink: Yup :-)

components/shared_code/src/shared/model/measurement.py
fniessink: I don't think that's needed. There's only one location where the hash is deleted and that is when a new one is added.

PR #8669, opened 2024-05-13
Allow specifying partial database connection parameters

### LINKED ISSUE #8668: Allow specifying partial database connection parameters, instead of only the full URL
Allow for separately setting `DATABASE_USERNAME`, `DATABASE_PASSWORD`, `DATABASE_HOST`, `DATABASE_PORT`, instead of only `DATABASE_URL`.

wkoot:
Closes #8668


components/shared_code/src/shared/initialization/database.py
```diff
@@ -5,12 +5,17 @@
...
 def client() -> pymongo.MongoClient:  # pragma: no feature-test-cover
     """Return a pymongo client."""
-    database_url = os.environ.get("DATABASE_URL", DEFAULT_DATABASE_URL)
+    database_url = os.environ.get("DATABASE_URL")
+    if not database_url:
+        _db_user = os.environ.get("DATABASE_USERNAME", "root")
+        _db_pass = os.environ.get("DATABASE_PASSWORD", "root")
+        _db_host = os.environ.get("DATABASE_HOST", "localhost")
+        _db_port = os.environ.get("DATABASE_PORT", "27017")
+        database_url = f"mongodb://{_db_user}:{_db_pass}@{_db_host}:{_db_port}"
```
fniessink: Why the leading underscores?

components/shared_code/tests/shared/database/test_connection_params.py
```diff
@@ -0,0 +1,48 @@
...
+                case "DATABASE_USERNAME":
+                    return "user"
+                case "DATABASE_PASSWORD":
+                    return "pass"
+                case "DATABASE_HOST":
+                    return "host"
+                case "DATABASE_PORT":
+                    return 4242
+                case _:
+                    return default
```
fniessink: Wouldn't a dict lookup be much shorter? Also, it's a bit confusing that the parameter is called "value", but it's really the key.

fniessink (approved):

PR #8693, opened 2024-05-16
Add database migration for #8045.

### LINKED ISSUE #8045: Change default value for the branch parameter to 'main'

fniessink:


components/api_server/tests/initialization/test_database.py
```diff
@@ -137,23 +173,113 @@ def test_change_accessibility_violations_to_violations_when_report_has_accessibi
...
+                        "metrics": {
+                            METRIC_ID: {"type": "issues"},
+                            METRIC_ID2: {"type": "loc"},
+                        },
+                    },
+                },
+            },
+        ]
+        perform_migrations(self.database)
+        self.database.reports.replace_one.assert_not_called()
```
wkoot, line 223: Hmm, is it enough to assume the migrations work correctly when perform_migrations does not throw an exception?

components/api_server/src/initialization/database.py
```diff
@@ -72,17 +73,70 @@ def change_accessibility_violation_metrics_to_violations(database: Database) ->
...
+    "commented_out_code": ["sonarqube"],
+    "complex_units": ["sonarqube"],
+    "duplicated_lines": ["sonarqube"],
+    "loc": ["sonarqube"],
+    "long_units": ["sonarqube"],
+    "many_parameters": ["sonarqube"],
+    "remediation_effort": ["sonarqube"],
+    "security_warnings": ["sonarqube"],
+    "software_version": ["sonarqube"],
+    "source_up_to_dateness": ["azure_devops", "gitlab", "sonarqube"],
```
wkoot, line 118: Is this really the only place for devops and gitlab? I would have expected a lot more, what about `BRANCHES`?

fniessink: The BRANCHES parameter's default value is an empty list, so nothing to migrate there.

components/api_server/tests/initialization/test_database.py
fniessink: I guess we could test that the database isn't changed by using mongomock, but since this is temporary code we accept the risk.

PR #8698, opened 2024-05-17
Allow for filtering Dependency-Track projects.

### LINKED ISSUE #8686: Filter on project for Dependency Track metrics
Acceptance criteria:
- [x] Add multiple choice with addition parameter to select projects by name, called "Project (regular expressions or project names)". Default value when the parameter is empty is all projects. Placeholder "All projects".
- [x]  Add multiple choice with addition parameter to select projects by version, called "Version (regular expressions or project versions)". Default value when the parameter is empty is all versions. Placeholder "All versions".
- [x] Security warnings metric uses the parameters to filter warnings
- [x] Dependencies metric uses the parameters to filter dependencies 
- [x] Source up-to-dateness metric uses the parameters to filter SBOMs

fniessink:
When using Dependency-Track as source for dependencies, security warnings, or source-up-to-dateness, allow for filtering by project name or identifier.

Closes #8686.

### DIFF
--- a/components/collector/src/source_collectors/dependency_track/base.py
+++ b/components/collector/src/source_collectors/dependency_track/base.py
@@ -1,8 +1,23 @@
 """Dependency-Track base collector."""
 
+from collections.abc import AsyncIterator
+from typing import TypedDict, cast
+
 from base_collectors import SourceCollector
-from collector_utilities.type import URL
+from collector_utilities.functions import match_string_or_regular_expression
+from collector_utilities.type import URL, Response
 from model import SourceResponses
+
+
+class DependencyTrackProject(TypedDict):
+    """Project as returned by Dependency-Track."""
+
+    # Last BOM import is a Unix timestamp, despite the Dependency-Tracker Swagger docs saying it's a datetime string
+    # See https://github.com/DependencyTrack/dependency-track/issues/840
+    lastBomImport: int
+    name: str
+    uuid: str
+    version: str
 
 
 class DependencyTrackBase(SourceCollector):
@@ -40,9 +55,28 @@ class DependencyTrackBase(SourceCollector):
 
     async def _get_project_uuids(self) -> dict[str, str]:
         """Return a mapping of project UUIDs to project names."""
+        return {project["uuid"]: project["name"] async for project in self._get_projects()}
+
+    async def _get_projects(self) -> AsyncIterator[DependencyTrackProject]:
+        """Return the Dependency-Track projects."""
         projects_api = URL(await self._api_url() + "/project")
-        project_uuids = {}
         for response in await DependencyTrackBase._get_source_responses(self, projects_api):
-            projects = await response.json(content_type=None)
-            project_uuids.update({project["uuid"]: project["name"] for project in projects})
-        return project_uuids
+            # We need an async for-loop and yield projects one by one because Python has no `async yield from`,
+            # see https://peps.python.org/pep-0525/#asynchronous-yield-from
+            async for project in self._get_projects_from_response(response):
+                yield project
+
+    async def _get_projects_from_response(self, response: Response) -> AsyncIterator[DependencyTrackProject]:
+        """Return the projects from the response that match the configured project names and versions."""
+        project_names = cast(list, self._parameter("project_names"))
+        project_versions = cast(list, self._parameter("project_versions"))
+        for project in await response.json(content_type=None):
+            if self._project_matches(project, project_names, project_versions):
+                yield project
+
+    @staticmethod
+    def _project_matches(project: DependencyTrackProject, names: list[str], versions: list[str]) -> bool:
+        """Return whether the project name matches the project names and versions."""
+        project_matches_name = match_string_or_regular_expression(project["name"], names) if names else True
+        project_matches_version = match_string_or_regular_expression(project["version"], versions) if versions else True
+        return project_matches_name and project_matches_version
--- a/components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
+++ b/components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
@@ -1,22 +1,14 @@
 """Dependency-Track source up-to-dateness collector."""
 
 from datetime import datetime
-from typing import TypedDict
 
 from base_collectors import TimePassedCollector
 from collector_utilities.date_time import datetime_from_timestamp
+from collector_utilities.exceptions import CollectorError
 from collector_utilities.type import URL, Response
 from model import Entities, Entity, SourceResponses
 
-from .base import DependencyTrackBase
-
-
-class DependencyTrackProject(TypedDict):
-    """Project as returned by Dependency-Track."""
-
-    lastBomImport: int  # Timestamp
-    name: str
-    uuid: str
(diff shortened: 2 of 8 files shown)


components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
```diff
@@ -28,16 +20,19 @@ async def _api_url(self) -> URL:
...
         """Override to parse the timestamp from the response."""
-        projects = await response.json(content_type=None)
-        datetimes = [self._last_bom_import_datetime(project) for project in projects]
-        return self.minimum(datetimes)
+        if datetimes := [
+            self._last_bom_import_datetime(project) async for project in self._get_projects_from_response(response)
+        ]:
+            return self.minimum(datetimes)
+        error_message = "No projects found"
+        raise CollectorError(error_message)
```
wkoot, line 28: Why raise an error, don't we usually return no-data when filters don't match? Is it a solution to use something like `datetime.MIN`, or should this code not be reached when the filter doesn't match?

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
```diff
@@ -50,26 +50,51 @@ async def test_no_projects(self):
...
+
+    async def test_filter_by_project_regular_expression(self):
+        """Test filtering projects by regular expression."""
+        self.set_source_parameter("project_names", ["project .*"])
+        response = await self.collect(get_request_json_side_effect=[self.projects(), self.dependencies("1.0")])
+        self.assert_measurement(response, value="1", entities=self.entities("1.0", "up-to-date"))
+
+    async def test_filter_by_project_version(self):
+        """Test filtering projects by version."""
+        self.set_source_parameter("project_versions", ["1.2", "1.3"])
```
wkoot, line 91: Sneaky version numbers containing dots are also regex

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -40,9 +55,28 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
...
+        project_names = cast(list, self._parameter("project_names"))
+        project_versions = cast(list, self._parameter("project_versions"))
+        for project in await response.json(content_type=None):
+            if self._project_matches(project, project_names, project_versions):
+                yield project
+
+    @staticmethod
+    def _project_matches(project: DependencyTrackProject, names: list[str], versions: list[str]) -> bool:
+        """Return whether the project name matches the project names and versions."""
+        project_matches_name = not names or match_string_or_regular_expression(project["name"], names)
```
wkoot: The `not names` syntax is a bit confusing, perhaps rename the args to `name_filter` and `version_filter` if you don't want to wrap it in if clauses :)

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -40,9 +55,28 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
...
+        """Return the Dependency-Track projects."""
         projects_api = URL(await self._api_url() + "/project")
-        project_uuids = {}
         for response in await DependencyTrackBase._get_source_responses(self, projects_api):
-            projects = await response.json(content_type=None)
-            project_uuids.update({project["uuid"]: project["name"] for project in projects})
-        return project_uuids
+            # We need an async for-loop and yield projects one by one because Python has no `async yield from`,
+            # see https://peps.python.org/pep-0525/#asynchronous-yield-from
+            async for project in self._get_projects_from_response(response):
```
wkoot, line 66: Dang, this was surprisingly much work

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
```diff
@@ -46,20 +60,45 @@ async def test_no_projects(self):
...
 
-    async def test_filter_vulnerabilities(self):
+    async def test_filter_by_severity(self):
         """Test that vulnerabilities can be filtered."""
         self.set_source_parameter("severities", ["High", "Critical"])
-        response = await self.collect(get_request_json_side_effect=[self.projects, self.vulnerabilities])
+        response = await self.collect(get_request_json_side_effect=[self.projects(), self.findings()])
         self.assert_measurement(response, value="0", entities=[])
 
+    async def test_filter_by_project_name(self):
```
wkoot, line 77: If we make a DependencyTrackBaseTest that includes the filter tests, they'll be inherited by all these test classes

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
fniessink: "Give me the date time the items in this empty set were last updated" is non-sensical, hence the error. There's no "no-data" when the metric is a date time. Returning `datetime.MIN` just leads to confused users.

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
fniessink: You're right. Hmm. That means "1.2.3" would match e.g. "102.3" and "1.203". I guess it's unlikely in practice, but we could change `is_regexp` to prevent interpreting semantic versions as regular expression (there's an official regex for that: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)? What do you think?

Current version of `is_regexp`:

```python
def is_regexp(string: str) -> bool:
    """Return whether the string looks like a regular expression."""
    return bool(set("$^?.+*[]") & set(string))
```

Updated version:

```python
def is_regexp(string: str) -> bool:
    """Return whether the string looks like a regular expression."""
    if matches_semantic_version(string):
        return False
    return bool(set("$^?.+*[]") & set(string))
```

wkoot: I think we should keep it as-is, but perhaps a story on the backlog to think about special chars such as dot in regex. There are other solutions such as a tooltip, to inform the user that their input might lead to unexpected results

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Changed to use ternary operators:

```python
    @staticmethod
    def _project_matches(project: DependencyTrackProject, names: list[str], versions: list[str]) -> bool:
        """Return whether the project name matches the project names and versions."""
        project_matches_name = match_string_or_regular_expression(project["name"], names) if names else True
        project_matches_version = match_string_or_regular_expression(project["version"], versions) if versions else True
        return project_matches_name and project_matches_version
```

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
fniessink: Agreed. I've opened https://github.com/ICTU/quality-time/issues/8739

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
fniessink: I'm getting some weird unit test failures in unrelated parts of the code when introducing a base module for Dependency-Track unit tests, so I'm going to open an issue for this.

PR #8747, opened 2024-05-22
Show spinner on first edit of source parameters.

### LINKED ISSUE #8736: Spinner for out of date measurement does not appear on the first edit
**Describe the bug**
Spinner for out of date measurement does not appear on the first edit

**To Reproduce**
1. Go to a metric not edited after installing v5.12.0
2. Change a source parameter
3. The spinner does not appear
4. Change a source parameter again
5. The spinner now does appear

**Expected behaviour**
Spinner should appear on first edit.

 - Quality-time version v5.12.0

fniessink:
The spinner indicating that the latest measurement of a metric is not up-to-date with the latest source configuration would not appear on the first edit of a metric after upgrading to v5.12.0.

Fixes #8736.

### DIFF
--- a/components/api_server/src/initialization/migrations.py
+++ b/components/api_server/src/initialization/migrations.py
@@ -2,32 +2,39 @@
 
 import logging
 
+import pymongo
+from pymongo.collection import Collection
 from pymongo.database import Database
+
+from shared.model.metric import Metric
 
 
 def perform_migrations(database: Database) -> None:  # pragma: no feature-test-cover
     """Perform database migrations."""
-    change_accessibility_violation_metrics_to_violations(database)
-    fix_branch_parameters_without_value(database)
+    for report in database.reports.find(filter={"last": True, "deleted": {"$exists": False}}):
+        change_accessibility_violation_metrics_to_violations(database, report)
+        fix_branch_parameters_without_value(database, report)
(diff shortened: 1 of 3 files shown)


components/api_server/src/initialization/migrations.py
```diff
@@ -87,8 +93,32 @@ def sources_with_branch_parameter(report: dict):  # pragma: no feature-test-cove
...
+                continue
+            latest_measurement["source_parameter_hash"] = source_parameter_hash(metric)
+            replace_document(database.measurements, latest_measurement)
+
+
+def source_parameter_hash(metric) -> str:  # pragma: no feature-test-cover
+    """Return a hash of the source parameters that can be used to check for changed parameters."""
+    sources = [source for _, source in sorted(metric.get("sources", {}).items())]
+    source_parameters = [source.get("parameters") for source in sources]
+    return md5_hash(str(source_parameters))
```
wkoot: Can't we import the function from where it's being set, instead of re-implementing?

fniessink: We could, but it means the migration code would get coupled to the Metric model class. Since this is (was, since this morning's discussion) long-living code I decided against that. I'll change it.

wkoot (approved):

PR #8768, opened 2024-05-24
Don't rearrange dropdown dynamically.

### LINKED ISSUE #8745: Pop-up menu changes width unexpectedly
**Describe the bug**
the pop-up menu for adding metrics changes width when one click the checkbox for hiding metrics.

**To Reproduce**
I believe this only happens for CI-environment subject.

**Expected behaviour**

**Screenshots**
image

image

fniessink:
In the dropdown of the "Add metric" button, keep the two checkboxes next to each other regardless of the width of the menu.

Fixes #8745.

### DIFF
--- a/components/frontend/src/widgets/Button.js
+++ b/components/frontend/src/widgets/Button.js
@@ -36,6 +36,47 @@ FilterCheckbox.propTypes = {
     setFilter: func,
 }
 
+function FilterCheckboxes({
+    itemType,
+    allowHidingUnsupportedItems,
+    showUnsupportedItems,
+    setShowUnsupportedItems,
+    allowHidingUsedItems,
+    hideUsedItems,
+    setHideUsedItems,
+}) {
+    if (!allowHidingUnsupportedItems && !allowHidingUsedItems) {
+        return null
+    }
+    return (
+        <span style={{ paddingRight: "12px" }}>
+            {allowHidingUnsupportedItems && (
+                <FilterCheckbox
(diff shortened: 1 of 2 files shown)


components/frontend/src/widgets/Button.js
```diff
@@ -155,20 +195,15 @@ export function AddDropdownButton({ itemSubtypes, itemType, onClick, allItemSubt
...
-                            />
-                        )}
-                        {usedItemSubtypeKeys?.length > 0 && (
-                            <FilterCheckbox
-                                label={`Hide ${itemType} types already used`}
-                                filter={hideUsedItems}
-                                setFilter={setHideUsedItems}
-                            />
-                        )}
+                        <FilterCheckboxes
```
wkoot, line 198: Nice to group the logic like this

components/frontend/src/widgets/Button.js
```diff
@@ -155,20 +195,15 @@ export function AddDropdownButton({ itemSubtypes, itemType, onClick, allItemSubt
...
+                            itemType={itemType}
+                            allowHidingUnsupportedItems={allItemSubtypes?.length > 0}
+                            showUnsupportedItems={showUnsupportedItems}
```
wkoot, line 201: Although the variable name is the same as the arg, wouldn't it be more explicit to include the word "filter" in the arg name?

wkoot: Oh, nvm. I only just saw that this is also a boolean

PR #8769, opened 2024-05-25
Refactor collector measurement model classes.

fniessink:
Make the SourceMeasurement and MetricMeasurement classes dataclasses.


components/collector/src/model/measurement.py
```diff
@@ -10,39 +12,43 @@
 from .issue_status import IssueStatus
 
 
+@dataclass
```
wkoot, line 15: Awesome

wkoot (approved):

PR #8770, opened 2024-05-25
Group digits in numbers.

### LINKED ISSUE #8076: Thousand separator
 I'm always frustrated when looking at the size(LOC) metric and I hardly see the difference between 500000 and 5000000

I would like Quality-time to use a period or a space as a thousand separator to make it more readable.
Read [this](https://en.wikipedia.org/wiki/Decimal_separator#:~:text=Space%2C%20the%20internationally%20recommended,in%20most%20English%2Dspeaking%20countries.) before implementing.

If this could be implemented in a way that all numbers (in every metric) use this thousand separator, that would be the best solution.

See also #7247

fniessink:
Group digits in groups of three in the UI.

Closes #8076.

### DIFF
--- a/components/frontend/src/source/SourceEntityAttribute.js
+++ b/components/frontend/src/source/SourceEntityAttribute.js
@@ -1,31 +1,49 @@
+import { string } from "prop-types"
+
 import { StatusIcon } from "../measurement/StatusIcon"
+import { entityPropType } from "../sharedPropTypes"
+import { formatMetricValue } from "../utils"
 import { HyperLink } from "../widgets/HyperLink"
 import { TimeAgoWithDate } from "../widgets/TimeAgoWithDate"
 
-export function SourceEntityAttribute({ entity, entity_attribute }) {
-    let cellContents = entity[entity_attribute.key] ?? ""
+export function SourceEntityAttribute({ entity, entityAttribute }) {
+    let cellContents = entity[entityAttribute.key] ?? ""
     if (typeof cellContents === "string" && cellContents.length >= 250) {
         cellContents = cellContents.slice(0, 247) + "..."
     }
-    const type = entity_attribute.type ?? ""
-    cellContents =
-        cellContents && type === "datetime" ? <TimeAgoWithDate dateFirst date={cellContents} /> : cellContents
-    cellContents =
-        cellContents && type === "date" ? <TimeAgoWithDate dateFirst noTime date={cellContents} /> : cellContents
-    cellContents = cellContents && type.includes("integer") ? Math.round(cellContents).toString() : cellContents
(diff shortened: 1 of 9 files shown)


components/frontend/src/utils.js
```diff
@@ -322,6 +322,18 @@ export function scaled_number(number) {
     return (number / Math.pow(1000, exponent)).toFixed(0) + scale[exponent]
 }
 
+export function formatMetricValue(scale, value) {
+    // metric value may be a string (version number or percentage), float, or integer
```
wkoot: The comment is a bit misleading, I guess it applies to the check on line 330?

components/frontend/src/source/SourceEntityAttribute.js
```diff
@@ -12,7 +13,8 @@ export function SourceEntityAttribute({ entity, entity_attribute }) {
         cellContents && type === "datetime" ? <TimeAgoWithDate dateFirst date={cellContents} /> : cellContents
     cellContents =
         cellContents && type === "date" ? <TimeAgoWithDate dateFirst noTime date={cellContents} /> : cellContents
-    cellContents = cellContents && type.includes("integer") ? Math.round(cellContents).toString() : cellContents
+    cellContents =
+        cellContents && type.includes("integer") ? formatMetricValue("count", Math.round(cellContents)) : cellContents
```
wkoot: Is "integer" the only case, or does that also include floats? Are version numbers always labeled as type string?

components/frontend/src/utils.js
```diff
@@ -322,6 +322,18 @@ export function scaled_number(number) {
...
+    if (value === "?") {
+        return value
+    }
+    if (scale === "count") {
+        const number = Number(value)
+        return number.toLocaleString(undefined, { useGrouping: true })
```
wkoot, line 332: The thousand separators comes from LocaleString builtin? Perhaps then it's better to rename the change to "apply locale to digits", and is this something we can add tests for?

components/frontend/src/utils.js
fniessink: Replaced the comment by propTypes.

components/frontend/src/source/SourceEntityAttribute.js
fniessink: Entity attribute types can also be floats, but so far they apparently haven't needed special formatting. I've added a reference to the data model where these types are defined.

components/frontend/src/utils.js
fniessink: I used a more generic name because I think there's more metric value formatting that could be moved to this method, like adding a percentage symbol. I'll see what can easily be moved here.

PR #8771, opened 2024-05-26
Fix measurement entity status menu items.

### LINKED ISSUE #8284: Fix the descriptions of measurement entity status menu items
I'm always frustrated when I want to manage entity statuses and the comments speak of 'undefined days' while the days for each state have been defined in the reaction times of measurement entities.

That the defined time is used in the comments.

**Additional context**
image

image

fniessink:
In the measurement entity status menu, the description of the menu items would say "undefined days" if the desired response time for the status had not been changed from its default value.

Fixes #8284.

### DIFF
--- a/components/frontend/src/source/SourceEntityDetails.js
+++ b/components/frontend/src/source/SourceEntityDetails.js
@@ -7,7 +7,7 @@ import { DateInput } from "../fields/DateInput"
 import { SingleChoiceInput } from "../fields/SingleChoiceInput"
 import { TextInput } from "../fields/TextInput"
 import { entityPropType, entityStatusPropType, reportPropType } from "../sharedPropTypes"
-import { capitalize } from "../utils"
+import { capitalize, getDesiredResponseTime } from "../utils"
 import { LabelWithDate } from "../widgets/LabelWithDate"
 import { source_entity_status_name as status_name } from "./source_entity_status"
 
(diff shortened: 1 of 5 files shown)


components/frontend/src/source/SourceEntityDetails.js
```diff
@@ -27,12 +27,11 @@ entityStatusOption.propTypes = {
...
-    const confirmedSubheader = `This ${entityType} has been reviewed and should be dealt with within ${desiredResponseTimes["confirmed"]} days.`
-    const fixedSubheader = `Ignore this ${entityType} for ${desiredResponseTimes["fixed"]} days because it has been fixed or will be fixed shortly.`
-    const falsePositiveSubheader = `Ignore this ${entityType} for ${desiredResponseTimes["false_positive"]} days because it's been incorrectly identified as ${entityType}.`
-    const wontFixSubheader = `Ignore this ${entityType} for ${desiredResponseTimes["wont_fix"]} days because it will not be fixed.`
+    const confirmedSubheader = `This ${entityType} has been reviewed and should be dealt with within ${getDesiredResponseTime(report, "confirmed")} days.`
+    const fixedSubheader = `Ignore this ${entityType} for ${getDesiredResponseTime(report, "fixed")} days because it has been fixed or will be fixed shortly.`
+    const falsePositiveSubheader = `Ignore this ${entityType} for ${getDesiredResponseTime(report, "false_positive")} days because it has been incorrectly identified as ${entityType}.`
```
wkoot: Consider rewording to `incorrectly identified as a ${entityType}` or `incorrectly identified as "${entityType}"`

components/frontend/src/source/SourceEntityDetails.js
```diff
@@ -27,12 +27,11 @@ entityStatusOption.propTypes = {
 }
 
 function entityStatusOptions(entityType, report) {
-    const desiredResponseTimes = report?.desired_response_times ?? {}
     const unconfirmedSubheader = `This ${entityType} should be reviewed to decide what to do with it.`
-    const confirmedSubheader = `This ${entityType} has been reviewed and should be dealt with within ${desiredResponseTimes["confirmed"]} days.`
-    const fixedSubheader = `Ignore this ${entityType} for ${desiredResponseTimes["fixed"]} days because it has been fixed or will be fixed shortly.`
-    const falsePositiveSubheader = `Ignore this ${entityType} for ${desiredResponseTimes["false_positive"]} days because it's been incorrectly identified as ${entityType}.`
-    const wontFixSubheader = `Ignore this ${entityType} for ${desiredResponseTimes["wont_fix"]} days because it will not be fixed.`
+    const confirmedSubheader = `This ${entityType} has been reviewed and should be dealt with within ${getDesiredResponseTime(report, "confirmed")} days.`
```
wkoot:
```suggestion
    const confirmedSubheader = `This ${entityType} has been reviewed and should be addressed within ${getDesiredResponseTime(report, "confirmed")} days.`
```

components/frontend/src/source/SourceEntityDetails.js
```diff
@@ -27,12 +27,11 @@ entityStatusOption.propTypes = {
...
+    const fixedSubheader = `Ignore this ${entityType} for ${getDesiredResponseTime(report, "fixed")} days because it has been fixed or will be fixed shortly.`
```
wkoot, line 32: I'm not sure that "shortly" is correct, perhaps "promptly" or "imminently" is more suitable?

components/frontend/src/source/SourceEntityDetails.js
fniessink: Hmm "as a" won't work for issues, for example, because it would need to be "as an".  

I think the quotes would be an improvement, but instead around the first occurrence of the entity type, since that's where you'd use air quotes if you had to say the sentence out loud:

> Ignore this "violation" for 180 days because it has been incorrectly been identified as violation.

components/frontend/src/source/SourceEntityDetails.js
fniessink: I think shortly is fine, see second example here: https://dictionary.cambridge.org/dictionary/english/shortly

PR #8772, opened 2024-05-26
Remove dead code and fix Sonarcloud issues.

fniessink:
- Remove the `editableLabel` prop of `EditableTextInput`, it's not used anywhere.
- Move the `tabIndex` from the `div` containing the Settings menu item to the menu item itself, where it belongs. 
- Instead of having the first table cell of the `TableRowWithDetails` pretending it is a button, give it a proper button. This should improve accessibility. Also add some styling to only show the button appearance when the button has focus or is hovered so that the subject table does not get cluttered with a whole column of expand/collapse buttons.

### DIFF
--- a/components/frontend/src/widgets/TableRowWithDetails.css
+++ b/components/frontend/src/widgets/TableRowWithDetails.css
@@ -1,0 +1,3 @@
+.ui.basic.button.expandcollapse:not(:hover, :focus) {
+    box-shadow: 0 0 0 0px transparent;
+}
--- a/components/frontend/src/widgets/TableRowWithDetails.js
+++ b/components/frontend/src/widgets/TableRowWithDetails.js
@@ -1,6 +1,8 @@
+import "./TableRowWithDetails.css"
+
 import { bool, func, object } from "prop-types"
 
-import { Icon, Table } from "../semantic_ui_react_wrappers"
+import { Button, Icon, Table } from "../semantic_ui_react_wrappers"
 import { childrenPropType } from "../sharedPropTypes"
 
 export function TableRowWithDetails(props) {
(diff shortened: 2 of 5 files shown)


components/frontend/src/widgets/TableRowWithDetails.css
```diff
@@ -0,0 +1,3 @@
+.ui.basic.button.expandcollapse:not(:hover, :focus) {
+    box-shadow: 0 0 0 0px transparent;
```
wkoot, line 2: Does this affect print to pdf in any way?

components/frontend/src/widgets/TableRowWithDetails.js
```diff
@@ -1,27 +1,23 @@
...
-                        onExpand(!expanded)
-                    }}
-                    tabIndex="0"
-                    textAlign="center"
-                    style={style}
-                    role="button"
-                >
-                    <Icon size="large" name={expanded ? "caret down" : "caret right"} />
+                <Table.Cell collapsing textAlign="center" style={style}>
+                    <Button
```
wkoot, line 14: Is there no functional test that requires altering due to this change?

components/frontend/src/widgets/TableRowWithDetails.css
fniessink: Good question, but no.

components/frontend/src/widgets/TableRowWithDetails.js
fniessink: No, because the Table.Cell was rendered as a button before (see https://react.semantic-ui.com/augmentation/) so all the tests keep working.

PR #8782, opened 2024-05-27
Migrate to the new SonarQube issue structure.

fniessink:
Migrate to the new SonarQube issue structure introduced in SonarQube 10.2. This means:
- The possible parameter values of the severity parameter are now "low", "medium", and "high". Previously allowed values were "critical", "blocker", "major", "minor", and "info".
- Migrate the severity parameter values to the new values according to the mapping documented by SonarSource: https://docs.sonarsource.com/sonarqube/latest/user-guide/issues/#severity-mapping.
- Drop the `types` parameter
- Add a "software quality" parameter and use the impactSoftwareQualities parameter for the api/issues/search endpoint
- Add a "clean code attribute" parameter and use the cleanCodeAttributeCategories parameter for the api/issues/search endpoint
- Change the `security_types` parameter so it does not depend on the issue type.
- Document that the lowest supported SonarQube version is 10.2. As implemented in #8786. 
- Migrate the contents of the `types` parameter to the new impacted software qualities parameter.

Closes #8354.

PR #8819, opened 2024-05-30
Allow for specifying supported source versions.

### LINKED ISSUE #8786: Allow for specifying supported versions of sources
# User story

As a user of Quality-time who configures sources I want to see which versions of the source are supported by Quality-time.

## Acceptance criteria

- [x] Quality-time shows the supported source version(s) in the source drop down for adding a source to a metric.
- [x] Quality-time shows the supported source version(s) after adding a source to a metric.
- [x] The Quality-time reference documentation contains the supported source version(s).
- [x] Documenting the supported source version is optional (otherwise we would have to document them all at once).

# Tasks

### Frontend

- [x] Show supported versions in the source drop down for adding a source to a metric, if available in the data model.
- [x] Add supported versions to the source configuration tab of metrics after adding a source, if available in the data model.

### Shared code

- [x] Add `supported_versions` to the `source` object in the data model

## Documentation

- [x] Update changelog
- [x] Update reference documentation to include supported source versions.

fniessink:
Allow for specifying supported source versions in the data model. Show the supported source version in the UI and the reference documentation.

Blocks #8782.

Closes #8786.


components/shared_code/src/shared_data_model/meta/source.py
```diff
@@ -26,6 +26,7 @@ class Source(DescribedModel):
     parameter_layout: dict[str, ParameterGroup] = DEFAULT_PARAMETER_LAYOUT
     entities: dict[str, Entity] = {}
     issue_tracker: bool | None = False
+    supported_versions: str | None = None  # The source versions that Quality-time supports, e.g. ">=10.2"
```
wkoot: The title of the attr made me suspect a (possibly empty) list of strings. I'd say it's more of a `supported_version_str`, but we're not that exact on some other fields so should be fine until we have a different usecase :-)

docs/src/create_reference_md.py
```diff
@@ -141,6 +141,10 @@ def source_section(source: Source, source_key: str, level: int) -> str:
         # Add generic documentation, meaning documentation that applies to all metrics that the source supports, here.
         # Documentation for specific metrics is added in the metric-source sections, see metric_source_section() below.
         markdown += markdown_paragraph(documentation)
+    if source.supported_versions:
+        markdown += f"```{{admonition}} Supported {source.name} versions\n:class: important\n"
```
wkoot: Should there not be a newline right before this as well? Or is it safe to assume that previous output always ends with (multiple) newline(s)?

components/shared_code/src/shared_data_model/sources/owasp_dependency_check.py
```diff
@@ -30,6 +30,7 @@
     name="OWASP Dependency-Check",
     description="OWASP Dependency-Check is a utility that identifies project dependencies and checks if there are any "
     "known, publicly disclosed, vulnerabilities.",
+    supported_versions="Versions that use v2.0 - v2.5, v3.0, v3.1, or v4.0 of the OWASP Dependency-Check XSD",
```
wkoot: It doesn't seem to be super easy to find out which version is being used from the url below, nor search engine. Is it referenced in the source output data somewhere?

components/frontend/src/widgets/Button.js
```diff
@@ -172,7 +172,7 @@ export function AddDropdownButton({ itemSubtypes, itemType, onClick, allItemSubt
                     }
                     value={null} // Without this, a selected item becomes active (shown bold in the menu) and can't be selected again
                 >
-                    <Dropdown.Menu>
+                    <Dropdown.Menu style={{ minWidth: "50em" }}>
```
wkoot, line 175: Is this another fix, or needed for this feature?

components/shared_code/src/shared_data_model/meta/source.py
fniessink: I don't like Hungarian notation, but I can make it `supported_versions_description`.

docs/src/create_reference_md.py
fniessink: Yes, the previous lines always add a paragraph so no newlines needed here.

components/shared_code/src/shared_data_model/sources/owasp_dependency_check.py
fniessink: I have changed it to `>=9.0, <=9.2`. Versions older than 9.0 are not supported by OWASP because of the source database API changes. I did not remove the old XSD versions so Quality-time will still work with older versions, but they shouldn't be used anymore.

components/frontend/src/widgets/Button.js
fniessink: I've enabled line breaks so that long source descriptions don't run off the screen. But then Semantic UI React makes the menu as wide as the trigger which is a button, so that's a bit too small.

PR #8829, opened 2024-06-02
Add more subject types.

### LINKED ISSUE #3130: More subject types
When using the QT instance itself as a metric source. The issues can accumulate well above the 100 items. If QT had more detail in subject types, this could have been avoided by creating more metrics with these detailed subject types.

Suggested subject types (bold) with accompanying metrics. Subtypes have the same metrics as their supertype, plus the additional metrics listed under the subtype.

- **Process**
  - **Continuous integration/deployment**
    - Merge requests
    - Unmerged branches
    - Pipeline duration
  - **Technical debt management**
    - Dependencies (for outdated dependencies)
    - Issues
    - Suppressed violations 
    - Violation remediation effort
  - **Defect management**
    - Issues
  - **Test management**
    - Tests
    - Test cases
    - Manual test duration
    - Manual test execution
    - (Also include test coverage metrics?)
  - **Backlog management**
    - User story points (to measure ready work)
    - Velocity (to know how much work should be ready)    
- **Software**
  - Size
  - Security warnings
  - **Ops**
    - DORA metrics
  - **Application**
    - Scalability 
    - Slow transactions
    - Test cases
  - **Component**
    - Commented out code
    - Complex units
    - Dependencies
    - Duplicated lines
    - Long units
    - Many parameters
    - Violations
  - **Tests**
    - Tests
    - Test branch coverage
    - Test line coverage
  - **Documentation**
    - Source up-to-dateness
    - Time remaining
- **Development environment**
  - **CI-tools**
    - Failed CI-jobs
    - Unused CI-jobs
  - **Performancetest environment**
    - Performancetest duration
    - Performancetest stability
  - **Development tool**
    - Source version
- **Quality report**
  - Missing metrics
  - Metrics

Add both the subject type and the subject subtype to the list of subjects. The subject type allows adding all metrics of the underlying subtypes.

Changes should be backwards compatible.

Tasks:
1. Rename the CI-environment "ci" to "development_environment" in the data model and the default name to "Development environment".
1. Write migration code to insert a new copy of each (non-deleted) report with the new subject key (if the report has one or more subjects with the key "ci").
1. Create a separate PR for 1 and 2.
1. Add the new subject subtypes to the data model. Subjects get an optional list of (sub)subjects, see https://pydantic-docs.helpmanual.io/usage/models/#recursive-models. In the database this should look something like this:
```json
subjects: {
    software: {
    name: "Software",
    description: "...",
    metrics: [...],
    subjects: {
       defect_management: {
          name: "...",
       }
    }
}
```
1. Add a validator to check that the list of metrics of a subject type contains all metrics of all subject subtypes (recursively).

Blocked by: #8176

fniessink:
Closes #3130.


docs/src/create_reference_md.py
```diff
@@ -180,7 +184,7 @@ def admonition(text: str, title: str = "", admonition: AdmonitionType = "admonit
 
 def slugify(name: str) -> str:
     """Return a slugified version of the name."""
-    return f'#{name.lower().replace(" ", "-").replace("(", "").replace(")", "")}'
+    return f'#{name.lower().replace(" ", "-").replace("(", "").replace(")", "").replace("/", "")}'
```
wkoot: Still seems ok but getting long, when should we switch to a regex replace?

components/frontend/src/subject/SubjectType.js
```diff
@@ -1,26 +1,39 @@
...
+                <Header as={headingLevel}>
+                    {bullet}
+                    <HeaderContent>
+                        {subjectType.name}
+                        <HeaderSubheader>{subjectType.description}</HeaderSubheader>
+                    </HeaderContent>
+                </Header>
+            ),
         })
+        options.push(...subjectTypes(subjectType?.subjects ?? [], level + 1))
```
wkoot, line 30: Why is it level+1 here?

components/collector/tests/source_collectors/quality_time/test_missing_metrics.py
```diff
@@ -51,19 +51,18 @@ def setUp(self):
...
-                        )
+                    self.entities.append(
+                        {
+                            "key": f"{report['report_uuid']}:{subject_uuid}:{metric_type}",
+                            "report": report["title"],
+                            "report_url": f"https://quality_time/{report['report_uuid']}",
+                            "subject": subject["name"],
+                            "subject_url": f"https://quality_time/{report['report_uuid']}#{subject_uuid}",
+                            "subject_uuid": f"{subject_uuid}",
+                            "subject_type": self.data_model["subjects"][subject["type"]]["name"],
```
wkoot: We don't need to use the prop `all_subjects` here?

components/shared_code/src/shared_data_model/subjects.py
```diff
@@ -7,35 +7,99 @@
...
+                name="Development tool",
+                description="A software development and/or maintenance tool.",
+                metrics=[
+                    "source_version",
+                    "software_version",
+                ],
+            ),
+        },
         metrics=[
-            "failed_jobs",
```
wkoot, line 11: I see 9 metrics removed here and 7 metrics added in the subjects above, is that correct?

components/shared_code/src/shared_data_model/subjects.py
```diff
@@ -7,35 +7,99 @@
...
     ),
     "process": Subject(
         name="Process",
-        description="A software development and/or maintenance process.",
+        description="A software development, maintenance, and/or operations process.",
+        subjects={
+            "process_backlog": Subject(
+                name="Backlog management",
+                description="A process to manage the product backlog.",
+                metrics=["user_story_points", "velocity"],
```
wkoot, line 39:
```suggestion
                metrics=[
                    "user_story_points",
                    "velocity",
                ],
```

docs/src/create_reference_md.py
eric-nieuwland: 1. Is het wel handig om de '#' in de ge-slugify-de naam op te nemen?
2. Alle voorkomens van karakters verwijderen/vervangen kan makkelijker, zie https://stackoverflow.com/questions/3939361/remove-specific-characters-from-a-string-in-python

fniessink: - Multiple replace: I'll use `str.translate` instead.
- Does # belong in the slugified name: I guess not, but removing it here means adding it to every URL that uses a slug. I have added a comment.

components/frontend/src/subject/SubjectType.js
fniessink: Because subjects are composite and the child subjects get a higher heading level (h4, h5 ...).

components/shared_code/src/shared_data_model/subjects.py
fniessink: Yes, with the new subject structure there's no guarantee the number of metrics removed from the parent and added to the child subjects is equal.

components/shared_code/src/shared_data_model/subjects.py
fniessink: Apparently, `ruff format` doesn't care for being consistent across lists :-(

components/collector/tests/source_collectors/quality_time/test_missing_metrics.py
fniessink: We can't, at least not easily. The missing metrics collector downloads the data model as JSON from the source-Quality-time. Instantiating the DataModel class using the JSON is possible in theory, but only if the source-Quality-time is an older version of Quality-time with an older version of the data model. We would need to specify which versions of Quality-time are supported as source for the missing metrics metric, but that is outside the scope of this story.

PR #8858, opened 2024-06-03
Use unprivileged container "nginxinc/nginx-unprivileged" as base for proxy

### LINKED ISSUE #8857: Use unprivileged nginx container
Use the unprivileged nginx container, so we don't need to allow additional capabilities
https://hub.docker.com/r/nginxinc/nginx-unprivileged/

wkoot:
Closes #8857

### DIFF
--- a/components/api_server/src/example-reports/example-report.json
+++ b/components/api_server/src/example-reports/example-report.json
@@ -107,7 +107,7 @@
                             "type": "testng",
                             "parameters": {
                                 "url": "http://testdata:8000/reports/testng/testng-results.xml",
-                                "landing_url": "http://testdata:8000/reports/testng/testng-results.xml",
+                                "landing_url": "http://localhost:8000/reports/testng/testng-results.xml",
                                 "username": "",
                                 "password": "",
                                 "private_token": "",
@@ -167,7 +167,7 @@
                             "type": "testng",
                             "parameters": {
                                 "url": "http://testdata:8000/reports/testng/testng-results.xml",
-                                "landing_url": "http://testdata:8000/reports/testng/testng-results.xml",
+                                "landing_url": "http://localhost:8000/reports/testng/testng-results.xml",
                                 "username": "",
                                 "password": "",
                                 "private_token": ""
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   www:
     image: ictu/quality-time_proxy:${QUALITY_TIME_VERSION}
     ports:
-      - "${PROXY_PORT:-80}:80"
+      - "80:8080"
     environment:
       - FRONTEND_HOST=${FRONTEND_HOST:-frontend}
       - FRONTEND_PORT=${FRONTEND_PORT:-5000}
@@ -89,7 +89,7 @@ services:
     image: ictu/quality-time_renderer:${QUALITY_TIME_VERSION}
     environment:
       - PROXY_HOST=${PROXY_HOST:-www}
-      - PROXY_PORT=${PROXY_PORT:-80}
+      - PROXY_PORT=${PROXY_PORT:-8080}
       - LC_ALL=${LC_ALL:-en_GB.UTF-8}  # Set the date format in the PDF export to DD-MM-YYYY
       - TZ=${TZ:-Europe/Amsterdam}  # Make the PDF export use the correct timezone
     volumes:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -4,8 +4,9 @@ services:
   www:
     image: ictu/quality-time_proxy:${QUALITY_TIME_VERSION}
     ports:
-      - "80:8080"
+      - "80:${PROXY_PORT:-80}"
     environment:
+      - PROXY_PORT=${PROXY_PORT:-80}
       - FRONTEND_HOST=${FRONTEND_HOST:-frontend}
       - FRONTEND_PORT=${FRONTEND_PORT:-5000}
       - API_SERVER_HOST=${API_SERVER_HOST:-api_server}
@@ -89,7 +90,7 @@ services:
     image: ictu/quality-time_renderer:${QUALITY_TIME_VERSION}
     environment:
       - PROXY_HOST=${PROXY_HOST:-www}
-      - PROXY_PORT=${PROXY_PORT:-8080}
+      - PROXY_PORT=${PROXY_PORT:-80}
       - LC_ALL=${LC_ALL:-en_GB.UTF-8}  # Set the date format in the PDF export to DD-MM-YYYY
       - TZ=${TZ:-Europe/Amsterdam}  # Make the PDF export use the correct timezone
     volumes:
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -28,6 +28,7 @@ If your currently installed *Quality-time* version is not v5.13.0, please first 
 
 ### Changed
 
+- Use unprivileged container `nginxinc/nginx-unprivileged` as base for the proxy component, so that it does not require additional capabilities. Closes [#8857](https://github.com/ICTU/quality-time/issues/8857).
 - Migrate to the new SonarQube issue structure introduced in SonarQube 10.2. See the [release 10.2 upgrade notes](https://docs.sonarsource.com/sonarqube/latest/setup-and-upgrade/release-upgrade-notes/#release-10.2-upgrade-notes). Closes [#8354](https://github.com/ICTU/quality-time/issues/8354).  Where possible, SonarQube parameters and parameter values are migrated automatically:
   - The 'severities' parameter is changed into 'impact severities' and the severity values are changed ('blocker' and 'critical' become 'high', 'major' becomes 'medium', and 'minor' and 'info' become 'low').
   - The 'types' parameter is changed into 'impacted software qualities' and the types are changed ('code smell' becomes 'maintainability', 'vulnerability' becomes 'security', and 'bug' becomes 'reliability').
--- a/docs/src/deployment.md
+++ b/docs/src/deployment.md
@@ -27,12 +27,17 @@ To deploy *Quality-time* locally, follow these steps:
 5. Go to http://localhost to access the application.
 6. To log in, use one of the default users available in the [test LDAP server](software.md#test-ldap-server).
 
-```{note}
-By default, the application listens internally on port 8080.
+Internally, the application listens on port 80 by default.
 To change this, set the `PROXY_PORT` environment variable to a different port before starting the application.
-For example: `export PROXY_PORT=8000`.
-Externally, the application listens on port 80.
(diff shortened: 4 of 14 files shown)


docs/src/changelog.md
```diff
@@ -24,6 +24,10 @@ If your currently installed *Quality-time* version is v4.10.0 or older, please r
 - In the measurement entity status menu, the description of the menu items would say "undefined days" if the desired response time for the status had not been changed from its default value. Fixes [#8284](https://github.com/ICTU/quality-time/issues/8284).
 - Allow for specifying supported source versions in the data model. Show the supported source version in the UI and the reference documentation. Closes [#8786](https://github.com/ICTU/quality-time/issues/8786).
 
+### Changed
+
+- Use unprivileged container "nginxinc/nginx-unprivileged" as base for proxy. Closes [#8857](https://github.com/ICTU/quality-time/issues/8857).
```
fniessink: Maybe add a small explanation for the change? 

```suggestion
- Use unprivileged container "nginxinc/nginx-unprivileged" as base for proxy so that .... Closes [#8857](https://github.com/ICTU/quality-time/issues/8857).
```

fniessink (approved):

components/api_server/src/example-reports/example-report.json
```diff
@@ -107,7 +107,7 @@
                             "type": "testng",
                             "parameters": {
                                 "url": "http://testdata:8000/reports/testng/testng-results.xml",
-                                "landing_url": "http://testdata:8000/reports/testng/testng-results.xml",
+                                "landing_url": "http://localhost:8000/reports/testng/testng-results.xml",
```
fniessink, line 110: This means the example report won't work when Quality-time is deployed to a Docker (or Kubernetes) host, while that is the scenario we have been supporting thus far with the example report.

components/api_server/src/example-reports/example-report.json
```diff
@@ -167,7 +167,7 @@
...
```
fniessink, line 170: See previous comment.

docker/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
   www:
     image: ictu/quality-time_proxy:${QUALITY_TIME_VERSION}
     ports:
-      - "${PROXY_PORT:-80}:80"
+      - "80:8080"
```
fniessink: If 8080 is the default proxy port, shouldn't this be:
 
```suggestion
      - "80:${PROXY_PORT:-8080}"
```

docs/src/changelog.md
```diff
@@ -24,6 +24,10 @@ If your currently installed *Quality-time* version is v4.10.0 or older, please r
 - In the measurement entity status menu, the description of the menu items would say "undefined days" if the desired response time for the status had not been changed from its default value. Fixes [#8284](https://github.com/ICTU/quality-time/issues/8284).
 - Allow for specifying supported source versions in the data model. Show the supported source version in the UI and the reference documentation. Closes [#8786](https://github.com/ICTU/quality-time/issues/8786).
 
+### Changed
+
+- Use unprivileged container `nginxinc/nginx-unprivileged` as base for proxy, so that it does not require additional capabilities. Closes [#8857](https://github.com/ICTU/quality-time/issues/8857).
```
fniessink:
```suggestion
- Use unprivileged container `nginxinc/nginx-unprivileged` as base for the proxy component, so that it does not require additional capabilities. Closes [#8857](https://github.com/ICTU/quality-time/issues/8857).
```

docs/src/deployment.md
```diff
@@ -20,7 +20,11 @@ To deploy *Quality-time* locally, follow these steps:
 6. To log in, use one of the default users available in the [test LDAP server](software.md#test-ldap-server).
 
 ```{note}
-By default, the application listens on port 80. To change this, set the `PROXY_PORT` environment variable to a different port before starting the application. For example: `export PROXY_PORT=1080`.
+By default, the application listens internally on port 8000.
```
fniessink: 8000? Did you mean 8080?

docs/src/development.md
```diff
@@ -125,6 +125,12 @@ ci/pip-install.sh
 python src/quality_time_notifier.py
 ```
 
+##### Running the proxy component
+
+The `proxy` component is mapped to the `www` service in the docker compose file.
+This container runs an unprivileged version of Nginx on port 8080, so it does not require additional capabilities.
+Because the upstream container assumes it is internally running on port 8080, this is not configurable.
```
fniessink: What's the "upstream container"? Did you mean "base image"?

fniessink (changes requested):
Users outside ICTU would need to change their compose files, right? That would mean we will have to make a major release.

components/api_server/src/example-reports/example-report.json
wkoot: I'm not sure, I brought the test data in line with all other cases. We specify `testdata` as host for "url" and `localhost` as host for "landing_url"

docker/docker-compose.yml
wkoot: We don't actually have any logic that uses the `PROXY_PORT` env var. It's possible to build but then we need to change Nginx config and change exposed port in Dockerfile. Also, we won't have dynamic port mappings for a kubernetes deploy via Helm

docs/src/development.md
wkoot: Indeed, I changed the docs

wkoot:
> Users outside ICTU would need to change their compose files, right? That would mean we will have to make a major release.

That depends on how they are using it. If they stack compose files then there will automatically be multiple port mappings active, because the port directive stacks. If they let the renderer connect to ingress, then port 80 remains as-is and they wouldn't need to change that. But they might indeed need to change the `PROXY_PORT` env var of the renderer as the default value is changed. 

Since there is definitely a need for manually checking, a major release would be a good idea. Alternatively, we could alter the default port to 80 for now and add a feature to override the port (which could be used by Helm).

docker/docker-compose.yml
wkoot: Addressed this by implementing `PROXY_PORT`.

components/api_server/src/example-reports/example-report.json
fniessink: Oh wait, you are right. Strike that comment.

fniessink:
@wkoot Both localhost:80 and localhost:8080 do not work for me after a `docker compose build; docker compose up`.

fniessink:
> @wkoot Both localhost:80 and localhost:8080 do not work for me after a `docker compose build; docker compose up`.

Not sure what was going on, but after a pull and another `docker compose build` it works.

docker/docker-compose.ci.yml
```diff
@@ -3,7 +3,8 @@
 
 services:
   www:
-    build: ../components/proxy
+    build:
+      context: ../components/proxy
```
fniessink: Is this a style issue or does it do different things? In any case, shouldn't we make the same change for the other services?

docs/src/deployment.md
```diff
@@ -28,7 +28,11 @@ To deploy *Quality-time* locally, follow these steps:
 6. To log in, use one of the default users available in the [test LDAP server](software.md#test-ldap-server).
 
 ```{note}
-By default, the application listens on port 80. To change this, set the `PROXY_PORT` environment variable to a different port before starting the application. For example: `export PROXY_PORT=1080`.
+By default, the application listens internally on port 80.
```
fniessink: To make the internal/external sentences have a consistent structure:

```suggestion
Internally, the application listens on port 80 by default.
```

docs/src/deployment.md
```diff
@@ -28,7 +28,11 @@ To deploy *Quality-time* locally, follow these steps:
...
+To change this, set the `PROXY_PORT` environment variable to a different port before starting the application.
+For example: `export PROXY_PORT=8080`.
+Externally, the application listens on port 80.
```
fniessink:
```suggestion
Externally, the application listens on port 80 by default.
```

docs/src/deployment.md
```diff
@@ -28,7 +28,11 @@ To deploy *Quality-time* locally, follow these steps:
...
+This can be changed in the default port mapping of the `www` service in the compose file.
```
fniessink:
````suggestion
To changes this, adapt the default port mapping of the `www` service in the compose file.
For example: 
```
www:
  ports:
    - "1080:${PROXY_PORT:-80}"
```
````

File: docs/src/software.md [hunk 12]
```diff
@@ -369,12 +369,13 @@ The proxy [Dockerfile](https://github.com/ICTU/quality-time/blob/master/componen
...
 
-| Name              | Default value | Description                         |
-|:------------------|:--------------|:------------------------------------|
-| `FRONTEND_HOST`   | `frontend`    | The host name of the frontend.      |
-| `FRONTEND_PORT`   | `5000`        | The port the frontend listens on.   |
-| `API_SERVER_HOST` | `api_server`  | The hostname of the API-server.     |
-| `API_SERVER_PORT` | `5001`        | The port the API-server listens on. |
+| Name              | Default value | Description                                              |
+|:------------------|:--------------|:---------------------------------------------------------|
+| `PROXY_PORT`      | `80`          | Port of the proxy, within the internal (docker) network. |
```
fniessink:
```suggestion
| `PROXY_PORT`      | `80`          | Port of the proxy, within the internal (Docker) network. |
```

docs/src/software.md
```diff
@@ -409,13 +410,13 @@ The [Dockerfile](https://github.com/ICTU/quality-time/blob/master/components/ren
...
-| `PROXY_PROTOCOL` | `http`        | Protocol of the proxy. The renderer uses this to access the reports that need to be exported to PDF.             |
-| `LC_ALL`         |               | Set the date format in the PDF export. For example, to get DD-MM-YYYY use: `en_GB.UTF-8`.                        |
-| `TZ`             |               | Make the PDF export use the correct timezone. For example, to get Central European Time use: `Europe/Amsterdam`. |
+| Name             | Default value | Description                                                                                                                            |
+|:-----------------|:--------------|:---------------------------------------------------------------------------------------------------------------------------------------|
+| `PROXY_HOST`     | `www`         | Hostname of the proxy. The renderer uses this to access the reports that need to be exported to PDF.                                   |
+| `PROXY_PORT`     | `80`          | Port of the proxy, within the internal (docker) network. The renderer uses this to access the reports that need to be exported to PDF. |
```
fniessink:
```suggestion
| `PROXY_PORT`     | `80`          | Port of the proxy, within the internal (Docker) network. The renderer uses this to access the reports that need to be exported to PDF. |
```

fniessink (approved):

docker/docker-compose.ci.yml
wkoot: Ah! This is an unneeded change, the default option is `context`. I had added a build argument, which required the separation into a dict with options

PR #8975, opened 2024-06-13
Update quality tools.

### LINKED ISSUE #8928: Add pyproject-fmt to quality tools
https://pypi.org/project/pyproject-fmt/

fniessink:
Update quality tools.

- Add [pyproject-fmt](https://pyproject-fmt.readthedocs.io/en/stable/) to quality tools
- Remove safety from quality tools (the free plan is limited to one developer).
- Remove duplication between unittest.sh scripts.
- Remove duplication between quality.sh scripts.
- Remove duplication between pip-compile.sh scripts.
- Remove duplication between pip-install.sh scripts.
- Add type and security checking to Python tests.

Closes #8928.


release/release.py
```diff
@@ -26,9 +26,9 @@ def get_version() -> str:
     with pathlib.Path(release_folder / "pyproject.toml").open(mode="rb") as py_project_toml_fp:
         py_project_toml = tomllib.load(py_project_toml_fp)
     version_re = py_project_toml["tool"]["bumpversion"]["parse"]
-    version_tags = [tag for tag in repo.tags if re.match(version_re, tag.tag.tag.strip("v"), re.MULTILINE)]
+    version_tags = [tag for tag in repo.tags if tag.tag and re.match(version_re, tag.tag.tag.strip("v"), re.MULTILINE)]
```
wkoot, line 30: Is this to get around the issue of unpushed local changes?

release/release.py
```diff
@@ -26,9 +26,9 @@ def get_version() -> str:
...
     latest_tag = sorted(version_tags, key=lambda tag: tag.commit.committed_datetime)[-1]
-    return latest_tag.tag.tag.strip("v")
+    return latest_tag.tag.tag.strip("v") if latest_tag.tag else "?"
```
wkoot: But what should we do with this, is it useful?

ci/pip-base.sh
```diff
@@ -0,0 +1,17 @@
...
+
+run_pip_compile() {
+    if [[ -f "requirements/requirements.txt" ]]; then
+        run pip-compile --output-file requirements/requirements.txt pyproject.toml
+    fi
+    run pip-compile --extra dev --output-file requirements/requirements-dev.txt pyproject.toml
+}
+
+run_pip_install() {
+    run pip install --ignore-installed --quiet --use-pep517 $@
```
wkoot, line 16: Why not `--upgrade` ?

ci/python_files_and_folders.py
```diff
@@ -0,0 +1,13 @@
+"""Determine the Python files and folders in the current directory."""
+
+from pathlib import Path
+
+def python_files_and_folders() -> list[str]:
+    """Return the Python files and folders in the current directory."""
+    python_files = [python_file.name for python_file in Path(".").glob('*.py') if not python_file.name.startswith(".")]
+    python_folders = [folder_name for folder_name in ("src", "tests") if Path(folder_name).exists()]
```
wkoot, line 8: Hmm, this is a bit sneaky perhaps

ci/requirements_files.py
```diff
@@ -0,0 +1,13 @@
+"""Determine the Python requirements files."""
+
+from pathlib import Path
+
+def requirements_files() -> list[str]:
+    """Return the Python requirements files in the requirements directory."""
+    # We never return the internal requirements file, because it does not need to be checked
+    requirements_filenames = ("requirements/requirements.txt", "requirements/requirements-dev.txt")
+    return [f"-r {filename}" for filename in requirements_filenames if Path(filename).exists()]
```
wkoot: Aren't we assuming in `pip-base.sh` that `requirements-dev.txt` exists?

ci/requirements_files.py
```diff
@@ -0,0 +1,13 @@
+"""Determine the Python requirements files."""
+
+from pathlib import Path
+
+def requirements_files() -> list[str]:
+    """Return the Python requirements files in the requirements directory."""
+    # We never return the internal requirements file, because it does not need to be checked
```
wkoot: This sounds like it's not in accordance with the docstring. We're only interested in requirements which are checked by ci?
Perhaps separate the logic to two methods, for installing from requirements and checking them?

components/notifier/pyproject.toml
```diff
@@ -1,30 +1,85 @@
...
+    "shared",
+    "shared_data_model",
 ]
+lint.isort.sections."tests" = [
+    "tests",
+]
+
+[tool.pyproject-fmt]
+indent = 4
+keep_full_version = true # Remove trailing zero's from version specifiers?
```
wkoot: I don't understand the comment

ci/quality-base.sh
```diff
@@ -0,0 +1,63 @@
...
+    run_pipx pip-audit --strict --progress-spinner=off $(python $(script_dir)/requirements_files.py)
+}
+
+run_vulture() {
+    run_pipx vulture --min-confidence 0 $PYTHON_FILES_AND_FOLDERS .vulture_ignore_list.py $@
+}
+
+run_vale() {
+    run_pipx vale sync
+    run_pipx vale --no-wrap src/*.md
```
wkoot: Are there no md files nested in other directories? Is it possible to use globbing like `**`?

ci/quality-base.sh
```diff
@@ -0,0 +1,63 @@
...
+}
+
+run_markdownlint() {
+    run ./node_modules/markdownlint-cli/markdownlint.js src/*.md
```
wkoot: Same here, this is building in an assumption that all the markdown files are located at root level of the src folder. But that's very specific to the implementation of docs, while this file lives in the ci folder.

release/release.py
fniessink: No, it's just a typing thing. Apparently `tag.tag` can be None.

ci/pip-base.sh
fniessink: It was already like this, but I think we do it like this because we install all dependencies, direct and indirect, using an exact version spec, so `--upgrade` doesn't make sense.

ci/python_files_and_folders.py
fniessink: I guess looking up the folders could be made more generic by looking for folders with python files in it, but I didn't feel it was worth it.

The same could be done in bash of course, but less readable :-)

ci/requirements_files.py
fniessink: Yeah, correct. Well spotted. What do you think. Make the same assumption here, or remove the assumption from `pip-base.sh`? Hmm, or try to remove the duplication, that would be even better of course.

ci/requirements_files.py
fniessink: Hmm, yeah. I'll have to think about this one.

components/notifier/pyproject.toml
fniessink: I'll change it, but if `keep_full_version` is `false`, pyproject-fmt removes training zero's from version numbers. For example, `requests==2.42.0` becomes `requests==2.42`.

ci/quality-base.sh
fniessink: Good points, I'll have to check.

fniessink: No nested markdown files in `docs/src` but `vale` indeed has a `--glob` command line option:

```bash
run_vale() {
    run_pipx vale sync
    run_pipx vale --no-wrap --glob "*.md" src
}
```

There are no markdown files in other folders worth adding value for.

ci/quality-base.sh
fniessink: `markdownlint` also supports glowing:

```bash
run_markdownlint() {
    run ./node_modules/markdownlint-cli/markdownlint.js src/**/*.md
}
```

release/release.py
fniessink: Since `latest_tag.tag` can be `None` (but no idea in what circumstances) I had to do it like this to keep `mypy` happy. Maybe it's better to raise an exception, I'll do that.

fniessink: Oh wait, mypy thinks `latest_tag.tag` can be `None`, but mypy is wrong, so we can just `cast`:

```python
    version_tags = [tag for tag in repo.tags if tag.tag and re.match(version_re, tag.tag.tag.strip("v"), re.MULTILINE)]
    latest_tag = sorted(version_tags, key=lambda tag: tag.commit.committed_datetime)[-1]
    # We cast latest_tag.tag to TagObject because we know it cannot be None, given how version_tags is constructed
    return cast(git.TagObject, latest_tag.tag).tag.strip("v")
```

ci/requirements_files.py
wkoot: It feels like there's quite some duplication indeed, but I'm not sure if that's easy to solve in the bash files

release/release.py
wkoot: It can indeed be none if you have a tag locally which hasn't been pushed to upstream. I encountered this when testing the changes to the release script :-) But I think it's fine if we cast, it's easy enough to deal with the exception

ci/requirements_files.py
fniessink: I've generalised `ci/requirements_files.py` a little bit so that it takes a template parameter:

```python
"""Determine the Python requirements files.

The script returns the requirements files as space separated string:

$ requirements_files.py
requirements/requirements.txt requirements/requirements-dev.txt

The script takes an optional template argument that is used to wrap each requirements filename. For example:

$ requirements_files.py "-r %s"
-r requirements/requirements.txt -r requirements/requirements-dev.txt
"""
```

Now both `pip_compile` and `pip_audit` can use it:

```bash
run_pip_compile() {
    for requirements_file in $(python $(script_dir)/requirements_files.py); do
        extra=$([[ "$requirements_file" == *"-dev"* ]] && echo "--extra dev" || echo "")
        run pip-compile $extra --output-file $requirements_file pyproject.toml
    done
}

run_pip_audit() {
    run_pipx pip-audit --strict --progress-spinner=off $(python $(script_dir)/requirements_files.py "-r %s")
}
```

ci/requirements_files.py
fniessink: It's not needed. The internal requirements are all off the form `-e ../shared_code` so they never need to be compiled nor checked. I've added a comment.

PR #8980, opened 2024-06-14
Close renderer headless browser in finally block.

### LINKED ISSUE #8979: Renderer has zombie processes when PDF creation fails
**Describe the bug**
When PDF creation fails, the browser processes are not killed.

**To Reproduce**
1. Use curl to create multiple PDF's at the same time on a renderer component with limited resources
2. One or more render attempts will timeout
3. The browser processes will not be killed.

**Expected behaviour**
The browser processes should be killed regardless  of whether PDF creation succeeds or not.

**Solution**
Close the browser in the finally branch of a try/catch statement.

fniessink:
Prevent zombie browser processes when creating the PDF fails by moving browser.close() to a finally block of the try/catch.

Closes #8979.

### DIFF
--- a/components/renderer/src/index.js
+++ b/components/renderer/src/index.js
@@ -6,21 +6,20 @@ const app = express();
 const RENDERER_PORT = process.env.RENDERER_PORT || 9000;
 const PROXY = `${process.env.PROXY_HOST || 'www'}:${process.env.PROXY_PORT || 80}`;
 const PROTOCOL = process.env.PROXY_PROTOCOL || 'http';
+let browser
 
 app.get("/api/health", async (_req, res) => {
-    res.status(200).send("OK")
+    // This endpoint is used by the Docker health check command (that in turn uses src/healthcheck.cjs) to report the
+    // health of the renderer component
+    const healthy = browser.connected ?? false
+    res.status(healthy ? 200 : 503).send(healthy ? "OK" : "Chromium not connected")
 })
 
 app.get("/api/render", async (req, res) => {
+    let webPage
     try {
         const url = sanitizeUrl(`${PROTOCOL}://${PROXY}/${req.query.path}`);
-        const browser = await puppeteer.launch({
-            defaultViewport: { width: 1500, height: 1000 },
-            args: ['--disable-dev-shm-usage', '--no-sandbox'],
-            // Opt in to new Chrome headless implementation, see https://developer.chrome.com/articles/new-headless/:
-            headless: "new",
-        });
-        const webPage = await browser.newPage();
+        webPage = await browser.newPage();
         await webPage.goto(url, {
             waitUntil: "networkidle2",
             timeout: 60000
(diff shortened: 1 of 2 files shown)


docs/src/changelog.md
```diff
@@ -17,7 +17,8 @@ If your currently installed *Quality-time* version is not v5.13.0, please first
 ### Fixed
 
 - In the dropdown of the "Add metric" button, keep the two checkboxes next to each other regardless of the width of the menu. Fixes [#8745](https://github.com/ICTU/quality-time/issues/8745).
-- The icon of the trend graph tab would not be shown. Closes [#8822](https://github.com/ICTU/quality-time/issues/8822).
+- The icon of the trend graph tab would not be shown. Fixes [#8822](https://github.com/ICTU/quality-time/issues/8822).
+- Prevent zombie browser processes in the renderer component by starting Chromium only once and opening and closing a page for each render. Fixes [#8979](https://github.com/ICTU/quality-time/issues/8979).
```
wkoot: "and opening and closing" sounds superfluous, should be sufficient to say "and closing"

components/renderer/src/index.js
```diff
@@ -6,21 +6,18 @@ const app = express();
 const RENDERER_PORT = process.env.RENDERER_PORT || 9000;
 const PROXY = `${process.env.PROXY_HOST || 'www'}:${process.env.PROXY_PORT || 80}`;
 const PROTOCOL = process.env.PROXY_PROTOCOL || 'http';
+let browser
 
 app.get("/api/health", async (_req, res) => {
-    res.status(200).send("OK")
+    const healthy = browser.connected ?? false
+    res.status(healthy ? 200 : 503).send(healthy ? "OK" : "Chromium not connected")
```
wkoot, line 15: What happens when it is unhealthy, is this only sent to console.log?

components/renderer/src/index.js
```diff
@@ -42,15 +39,26 @@ app.get("/api/render", async (req, res) => {
             }
         });
         console.log(`URL ${url}: PDF created`);
-        await browser.close();
```
wkoot, line 45: Do we still call `browser.close` somewhere? Or does `webPage.close` also close the browser recursively?

components/renderer/src/index.js
fniessink: No, the Docker healtcheck command uses src/healthcheck.cjs that in turn calls this endpoint to report the container health. I'll add a comment mentioning this.

components/renderer/src/index.js
fniessink: The PR comment is out of date (GitHub uses the commit message of the first commit to create the PR description, but then never updates it with the message of commits that are later force pushed. Annoying!). We now start the browser once and then never stop it. Only pages/tabs are opened and closed.

docs/src/changelog.md
fniessink: It's the browser that is opened once, but for each render a page is opened and closed. But I agree it could be worded more clearly. Changed the description.

PR #8999, opened 2024-06-17
Dashboard card showing metrics requiring action.

### LINKED ISSUE #8938: Filter metrics requiring actions via the dashboard
Tasks:
- Add a "Metrics requiring action" card to the dashboard, showing the number of red, yellow, and white metrics in a little table.
- When clicking the card, only show the metrics requiring action via the existing setting.
- Make the card hideable via the settings.

fniessink:
Show a card in the dashboard with the number of metrics that require action (red, yellow, and white metrics). The card can be hidden via the Settings panel. Clicking the card or setting "Visible metrics" to "Metrics requiring actions" in the Settings panel hides metrics that do not require action.

Closes #8938.


components/frontend/src/dashboard/FilterCard.css
```diff
@@ -0,0 +1,16 @@
...
+}
+
+.ui.card.filter_card .header {
+    margin-bottom: 0.3em;
+    overflow: hidden;
+    white-space: nowrap;
+}
+
+.ui.card.filter_card .header.blue {
+    color: #2185d0;
```
wkoot: Do we not have one css file with all colour definitions?

components/frontend/src/utils.js
```diff
@@ -254,7 +255,8 @@ sortWithLocaleCompare.propTypes = {
 
 function hideMetric(metric, metricsToHide, hiddenTags) {
     const hideBecauseNoActionNeeded =
-        metricsToHide === "no_action_needed" && ["target_met", "debt_target_met", "informative"].includes(metric.status)
+        metricsToHide === "no_action_required" &&
+        ["target_met", "debt_target_met", "informative"].includes(metric.status)
```
wkoot: Perhaps it's handy to create a different const with the statuses that don't require action

components/frontend/src/dashboard/FilterCard.css
fniessink: No, but good suggestion. I'll add one.

fniessink: I've added a few color variables to `App.css`, the rest we can do later.

components/frontend/src/utils.js
fniessink: Done.

PR #9047, opened 2024-06-22
Request a metric to be measured ASAP.

### LINKED ISSUE #920: Force measurement of a metric
# User story

As a user I want to see the new measurement value for a metric as soon as possible so that I know that the changes I made to the metric or source configuration or changes in the source give the expected results.

## Acceptance criteria

- [x] A user can force Quality-time to measure a metric as soon as possible.

## Assumptions

- ~~#7760 is implemented first, otherwise this feature would not provide any feedback to the user about what is happening.~~
- #3134 will imply that they are being measured. 

## Out of scope

- Showing that Quality-time is currently measuring a metric, see #7760.
- Showing that the config has changed and the current measurement value was measured with the old config, see #3134.

# Tasks

### Frontend

- [x] Add button "Measure now" next to the metric position buttons.
- [x] When the button is clicked, the frontend tells the API-server to remove the hash from the latest measurement of the metric.
- [x] Make the button inactive while the metric is being measured.

### Collector

- [x] Insert a new measurement after a measurement was requested, even if the measurement was not changed. This lets the frontend get a new measurement count via the number of measurements server-sent events end point so it knows the requested measurement was done.

### Shared code

- [x] Include a `measurement_requested` flag in the latest measurement summary so the frontend knows a measurement is pending.

## Feature tests

- [x] Add a scenario for requesting a measurement.
 
## Documentation

- [x] Update changelog

fniessink:
Allow for requesting a metric to be measured as soon as possible.

Closes #920.


tests/feature_tests/src/features/metric.feature
```diff
@@ -127,3 +127,14 @@ Feature: metric
...
+  Scenario: request measurement
+    Given an existing metric
+    And an existing source
+    And the collector has measured "100"
+    When the client requests the metric to be measured
+    Then the metric is being measured
+    When the client waits a second
+    And the collector measures "100"
+    And the client waits a second
+    Then the metric is not being measured
```
wkoot, line 140: Isn't this secretly two test scenarios in one?

components/frontend/src/utils.js
```diff
@@ -96,6 +96,16 @@ export function getMetricUnit(metric, dataModel) {
     return metric.unit || dataModel.metrics[metric.type].unit || ""
 }
 
+export function isMeasurementRequested(metric) {
+    if (metric.measurement_requested) {
+        if (metric.latest_measurement) {
+            return new Date(metric.measurement_requested) >= new Date(metric.latest_measurement.end)
```
wkoot, line 102: Does this automatically take timezone differences between server and client into account?

components/shared_code/src/shared/model/measurement.py
```diff
@@ -302,6 +302,8 @@ def summarize_latest(self) -> Self:
             if parameter_hash != self.metric.source_parameter_hash():
                 self["outdated"] = True
             del self["source_parameter_hash"]
+        if measurement_requested := self.metric.get("measurement_requested"):
+            self["measurement_requested"] = measurement_requested >= self["end"]
```
wkoot, line 306: Should we also add an entry to the audit log for this action?

tests/feature_tests/src/features/metric.feature
fniessink: I don't think so. There's just one user action: clicking the "Measure metric" button.

components/frontend/src/utils.js
fniessink: Oh, good one. Let me check. Yes, the frontend uses `new Date().toISOString()` which converts the date to an ISO-formatted string in UTC.

components/shared_code/src/shared/model/measurement.py
fniessink: There is, the frontend uses the `/api/internal/metric/<metric_uuid>/attribute/<metric_attribute>` endpoint with `metric_attribute=measurement_requested` to record the timestamp of the request. So it's logged the same as other metric attributes.

PR #9139, opened 2024-07-03
Refactor tabs.

fniessink:


components/frontend/src/semantic_ui_react_wrappers/Icon.js
```diff
@@ -1,8 +1,21 @@
+import { string } from "prop-types"
 import { useContext } from "react"
 import { Icon as SemanticUIIcon } from "semantic-ui-react"
 
 import { DarkMode } from "../context/DarkMode"
 
 export function Icon(props) {
-    return <SemanticUIIcon inverted={useContext(DarkMode)} {...props} />
+    let { className, name, ...otherProps } = props
+    /* Using name="linegraph" results in "Invalid prop `name` of value `linegraph` supplied to `Icon`."
```
wkoot, line 9: ouch, that's a sneaky.

components/frontend/src/widgets/TabPane.js
```diff
@@ -0,0 +1,44 @@
...
+
+import { bool, element, oneOfType, string } from "prop-types"
+import { useContext } from "react"
+import { Menu } from "semantic-ui-react"
+
+import { DarkMode } from "../context/DarkMode"
+import { Icon, Label, Tab } from "../semantic_ui_react_wrappers"
+
+function FocusableTab({ error, iconName, label }) {
+    const className = useContext(DarkMode) ? "tabbutton inverted" : "tabbutton"
```
wkoot, line 11: Can't we use a "darkmode" class instead of "inverted"?

components/frontend/src/semantic_ui_react_wrappers/Icon.js
fniessink: Yeah, but it works *and* suppresses the log message, so best possible result :-)

components/frontend/src/widgets/TabPane.js
fniessink: Yeah, we could here, but `inverted` is the classname used by Semantic UI CSS, so I wanted to be consistent with that.

PR #9140, opened 2024-07-03
Remove the option to use a custom base image for the www container

wkoot:


fniessink (commented):
Looks good. Maybe add a comment to the Dockerfile explaining why we're setting the capability?

PR #9194, opened 2024-07-09
Give measurement entity tables sticky headers.

### LINKED ISSUE #8879: Measurement entities also need to have sticky headers

fniessink:
Give measurement entity tables sticky headers, like the metric tables already have.

Closes #8879.

### DIFF
--- a/components/frontend/src/source/SourceEntities.css
+++ b/components/frontend/src/source/SourceEntities.css
@@ -1,0 +1,7 @@
+.ui.sortable.table.entities.stickyHeader > thead {
+    /* Make thead sticky by positioning the th's */
+    position: sticky;
+    /* Leave room for the menu bar, the subject title, and the subject table header row */
+    top: 187px;
+    z-index: 1;
+}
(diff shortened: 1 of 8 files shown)


components/frontend/src/subject/SubjectTitle.js
```diff
@@ -100,7 +100,7 @@ export function SubjectTitle({
             item_uuid={`${subject_uuid}:${tabIndex}`}
             level="h2"
             settings={settings}
-            style={{ marginTop: 50 }}
+            style={{ marginTop: 50, height: 50 /* Make height independent of whether the subject has a subtitle */ }}
```
wkoot:
```suggestion
            style={{ marginTop: 50, height: 50 /* Height is independent of whether the subject has a subtitle */ }}
```

components/frontend/src/source/SourceEntities.css
```diff
@@ -0,0 +1,7 @@
+.ui.sortable.table.entities.stickyHeader > thead {
+    /* Make thead sticky by positioning the th's */
+    position: sticky;
+    /* Leave room for the menu bar, the subject title, and the subject table header row */
+    top: 187px;
```
wkoot, line 5: Hmm, there's so many magic numbers in various files :( I guess it can't be helped but I can imagine it could become a dark rabbit hole at some point

fniessink: Yeah, this one should be calculated based on the height of various elements, but I have no idea how to do that in CSS.

PR #9275, opened 2024-07-19
Always show measurement entity tabs.

### LINKED ISSUE #8826: Always show measurement entities tabs
Use a loader until the data has been retrieved

Show an entity tab for each source.

For sources that can't have measurement entities show a message: "This source does not support measurement details."

When a source has zero measurement entities show a message: "Zero <unit>s." (for example "Zero violations."

fniessink:
Display a loader in the tab while loading measurements. If the measurement entities cannot be displayed, explain in the tab why not.

Closes #8826.

### DIFF
--- a/components/frontend/src/metric/MetricDetails.js
+++ b/components/frontend/src/metric/MetricDetails.js
@@ -171,32 +171,24 @@ export function MetricDetails({
             { iconName: "linegraph" },
         ),
     )
-    if (measurements.length > 0) {
-        lastMeasurement.sources.forEach((source) => {
-            const reportSource = metric.sources[source.source_uuid]
-            if (!reportSource) {
-                return
-            } // source was deleted, continue
-            const nrEntities = source.entities?.length ?? 0
-            if (nrEntities === 0) {
-                return
-            } // no entities to show, continue
-            const sourceName = getSourceName(reportSource, dataModel)
-            panes.push(
-                tabPane(
-                    sourceName,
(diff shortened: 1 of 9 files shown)


components/frontend/src/metric/MetricDetails.js
```diff
@@ -171,32 +171,24 @@ export function MetricDetails({
...
-                        metric_uuid={metric_uuid}
-                        source={source}
-                        reload={measurementsReload}
-                    />,
-                    { image: <Logo logo={reportSource.type} alt={sourceName} /> },
-                ),
-            )
-        })
-    }
+    Object.entries(metric.sources).forEach(([source_uuid, source]) => {
```
wkoot, line 174: Interesting simplification, does that come from an assumption we can make regarding the input data or was it just not so optimal before?

wkoot: Or was this logic moved to SourceEntities.js ?

wkoot (approved):

components/frontend/src/metric/MetricDetails.js
fniessink: It's simpler here because MetricDetails can just unconditionally render the SourceEntities tabs and the logic for dealing with missing measurement and/or entities is moved to the  SourceEntities component. Basically in the form of guard clauses that check whether the measurements are still being loaded, loading failed, or whether there are no entities at all, and if so, render an error or warning message.

PR #9349, opened 2024-07-30
Update to Python 3.13.

fniessink:
Issues preventing update:

- [x] ~~No wheel available for lxml. Upstream issue: https://bugs.launchpad.net/lxml/+bug/2068828~~
- [x] ~~Building wheels for greenlet fails. Upstream PR: https://github.com/python-greenlet/greenlet/pull/396~~
- [x] ~~No wheel available for aiohttp. Upstream issue: https://github.com/aio-libs/aiohttp/issues/8194~~
- [x] ~~API-server unittest fails with `AttributeError: '_MainThread' object has no attribute '_tstate_lock'`. Gevent issue: https://github.com/gevent/gevent/issues/2054~~
- [x] ~~No `cimg/python` 3.13 image available: https://hub.docker.com/r/cimg/python/tags?page_size=&ordering=&name=3.13~~
- [x] ~~No `ghcr.io/astral-sh/uv:python3.13-bookworm` image available. Should be available shortly as the PR has been merged: https://github.com/astral-sh/uv/pull/8105~~
- [x] ~~ReadTheDocs does not support 3.13 yet: https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python, ticket: https://github.com/readthedocs/readthedocs.org/issues/11671~~

### DIFF
--- a/components/api_server/pyproject.toml
+++ b/components/api_server/pyproject.toml
@@ -1,10 +1,9 @@
 [project]
 name = "api-server"
 version = "5.19.0-rc.1"
-requires-python = ">=3.12"
+requires-python = ">=3.13"
 classifiers = [
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
@@ -36,7 +35,6 @@ optional-dependencies.tools = [
 ]
 
 [tool.ruff]
-target-version = "py312"
 line-length = 120
 src = [
     "src",
--- a/docs/src/conf.py
+++ b/docs/src/conf.py
@@ -25,7 +25,7 @@ create_reference_md.main()
 # -- Project information -----------------------------------------------------
 
 project = "Quality-time"
-copyright = "2021-2023, ICTU"  # noqa: A001
+copyright = "2021-2024, ICTU"  # noqa: A001
 author = "ICTU"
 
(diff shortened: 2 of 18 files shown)


components/api_server/pyproject.toml
```diff
@@ -36,7 +35,6 @@ optional-dependencies.tools = [
 ]
 
 [tool.ruff]
-target-version = "py312"
```
wkoot, line 39: Is there no added benefit to specifying a target version?

docs/src/conf.py
```diff
@@ -25,7 +25,7 @@
 # -- Project information -----------------------------------------------------
 
 project = "Quality-time"
-copyright = "2021-2023, ICTU"  # noqa: A001
+copyright = "2021-2024, ICTU"  # noqa: A001
```
wkoot, line 28: Hehe, might as well set it to 2025 now or leave a ticket for next year ;)

components/api_server/pyproject.toml
fniessink: It's (no longer?) needed: "If you're already using a pyproject.toml file, we recommend project.requires-python instead" (https://docs.astral.sh/ruff/settings/#target-version).

docs/src/conf.py
fniessink: Done: https://github.com/ICTU/quality-time/issues/10354

PR #9569, opened 2024-08-29
Don't fail on parsing npm outdated output with multiple dependents pe…

### LINKED ISSUE #9566: Parsing npm outdated output with multiple dependents per dependency fails
```python
Parse error
  File "/home/collector/base_collectors/source_collector.py", line 150, in __safely_parse_source_responses
    return await self._parse_source_responses(responses)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/collector/base_collectors/source_collector.py", line 163, in _parse_source_responses
    included_entities = [entity for entity in await self._parse_entities(responses) if self._include_entity(entity)]
                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/collector/base_collectors/file_source_collector.py", line 113, in _parse_entities
    entities.extend(self._parse_json(json, filename))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/collector/source_collectors/npm/dependencies.py", line 15, in _parse_json
    return Entities(
           ^^^^^^^^^
  File "/home/collector/model/entity.py", line 34, in __init__
    self.extend(entities)
  File "/home/collector/model/entity.py", line 52, in extend
    for entity in entities:
  File "/home/collector/source_collectors/npm/dependencies.py", line 17, in <genexpr>
    key=<redacted>'{dependency}@{versions.get("current", "?")}',
                        ^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

```json
{
"@testing-library/jest-dom": [
    {
      "current": "6.4.6",
      "wanted": "6.5.0",
      "latest": "6.5.0",
      "dependent": "foo",
      "location": "/builds/project/baz/node_modules/@testing-library/jest-dom"
    },
    {
      "current": "6.4.6",
      "wanted": "6.5.0",
      "latest": "6.5.0",
      "dependent": "bar",
      "location": "/builds/project/baznode_modules/@testing-library/jest-dom"
    }
  ]
}
```

fniessink:
…r dependency.

Fixes #9566.


components/collector/src/source_collectors/npm/dependencies.py
```diff
@@ -1,24 +1,45 @@
...
+            for version in versions:
+                entities.append(self._create_entity(dependency, version))
+        return entities
+
+    def _create_entity(self, dependency: str, version: DependencyVersionInfo) -> Entity:
+        """Create an entity for the dependency."""
+        key = f'{dependency}@{version.get("current", "?")}'
+        name = dependency
+        if "dependent" in version:
+            key += f"@{version['dependent']}"
```
wkoot, line 37: So we could potentially have a key like `package@?@dep`, is that problematic?

fniessink: I don't think so. A dependency should be uniquely identified by package name, current version ("?" if unknown), and dependent. The key is not visible to the users, so aesthetics do not come into play :-)

PR #9572, opened 2024-08-29
Show measurement entity overview in popup.

### LINKED ISSUE #7626: Show in measurement value popup how many entities are excluded from the measurement value (won't fix, false positive, fixed)
A measurement value may have excluded measurement entities (entities marked "False positive", "Won't fix" or "Will be fixed"), however, this is not visible without expanding the metric.

Add an indicator (e.g. *) next to the metric value when there are excluded entities. Make sure this indicator is not included in the PDF-export. 

When there are excluded entities, add an info message to the measurement value popup, detailing how many entities are excluded, per status.

fniessink:
Show the number of measurement entities marked as "Won't fix" and "Will be fixed" in the measurement value popup.

Closes #7626.

### DIFF
--- a/components/frontend/src/measurement/MeasurementValue.js
+++ b/components/frontend/src/measurement/MeasurementValue.js
@@ -1,37 +1,86 @@
+import "./MeasurementValue.css"
+
 import { bool, string } from "prop-types"
 import { useContext } from "react"
-import { Icon } from "semantic-ui-react"
 
 import { DataModel } from "../context/DataModel"
-import { Label, Popup } from "../semantic_ui_react_wrappers"
-import { datePropType, metricPropType } from "../sharedPropTypes"
+import { Icon, Label, Message, Popup } from "../semantic_ui_react_wrappers"
+import { datePropType, measurementPropType, metricPropType } from "../sharedPropTypes"
+import { IGNORABLE_SOURCE_ENTITY_STATUSES, SOURCE_ENTITY_STATUS_NAME } from "../source/source_entity_status"
 import {
     formatMetricValue,
     getMetricScale,
+    getMetricUnit,
     getMetricValue,
     isMeasurementRequested,
     MILLISECONDS_PER_HOUR,
+    sum,
 } from "../utils"
 import { TimeAgoWithDate } from "../widgets/TimeAgoWithDate"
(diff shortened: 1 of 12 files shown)


components/frontend/src/measurement/MeasurementValue.js
```diff
@@ -1,39 +1,91 @@
...
+            <Label color="red">
+                {ignoredEntitiesIcon} {value}
+            </Label>
+        )
     }
     if (updating) {
         return (
             <Label color="yellow">
-                <Icon loading name="hourglass" /> {value}
+                <Icon loading name="hourglass" /> {ignoredEntitiesIcon} {value}
```
wkoot: This seems a bit like code duplication, should we not create a helper method to check if it needs to wrap the content in colour yellow with hourglass icon?

components/frontend/src/measurement/MeasurementValue.js
```diff
@@ -1,39 +1,91 @@
...
+}
+ignoredEntitiesCount.propTypes = {
+    measurement: measurementPropType,
+}
+
+function ignoredEntitiesMessage(measurement, unit) {
+    const count = ignoredEntitiesCount(measurement)
+    let summary = `The measurement value excludes ${sum(count)} ${unit}.`
+    let details = ""
+    ;["false_positive", "fixed", "wont_fix"].forEach((status) => {
```
wkoot: What's this weird semicolon shorthand? Is it not a better idea to turn this list of three options into a constant which can also be used elsewhere (i.e. when assigning the value 0 to each of them in a dict)?

components/frontend/src/subject/SubjectTableHeader.js
```diff
@@ -76,6 +76,12 @@ const measurementHelp = (
             , the metric has not been measured recently. This indicates a problem with <em>Quality-time</em> itself, and
             a system administrator should be notified.
         </p>
+        <p>
+            If there is a <Icon name="hide" /> before the measurement value, it means one or more measurement entities
+            are being ignored. Hover over the measurement value to see how many entities are ignored. Expand the metric
+            (click <Icon fitted name="triangle right" />) and navigate to the entities tab to see why each entity is
```
wkoot: Perhaps "individual entities" instead of  "each entity" is more clear

PR #9573, opened 2024-08-29
When filtering metrics by clicking on the "Action required" card or t…

### LINKED ISSUE #9245: Donut charts do not update when clicking on metrics requiring action or metrics with issues
**To Reproduce**

2. Click on 'action required' card or 'issues' card an nothing happens with the other cards
3. but when clicking on the 'security' card updates all the other cards with donut charts

fniessink:
…he "Issues" card, update the donut charts in the other cards.

Fixes #9245.

### DIFF
--- a/components/frontend/src/report/Report.test.js
+++ b/components/frontend/src/report/Report.test.js
@@ -167,15 +167,3 @@ it("navigates to subject", async () => {
     expect(mockScroll).toHaveBeenCalledWith()
     expect(mockScrollBy).toHaveBeenCalledWith(0, 163)
 })
-
-it("does not navigate to a subject that is hidden", async () => {
-    history.push("?metrics_to_hide=no_issues")
-    const mockScroll = jest.fn()
-    window.HTMLElement.prototype.scrollIntoView = mockScroll
-    const mockScrollBy = jest.fn()
-    window.scrollBy = mockScrollBy
-    renderReport({ reportToRender: report })
-    fireEvent.click(screen.getAllByText(/Subject title/)[0])
-    expect(mockScroll).not.toHaveBeenCalled()
-    expect(mockScrollBy).not.toHaveBeenCalled()
-})
--- a/components/frontend/src/report/ReportDashboard.js
+++ b/components/frontend/src/report/ReportDashboard.js
@@ -8,33 +8,10 @@ import { IssuesCard } from "../dashboard/IssuesCard"
 import { LegendCard } from "../dashboard/LegendCard"
 import { MetricsRequiringActionCard } from "../dashboard/MetricsRequiringActionCard"
 import { MetricSummaryCard } from "../dashboard/MetricSummaryCard"
-import { STATUS_COLORS } from "../metric/status"
 import { datesPropType, measurementsPropType, reportPropType, settingsPropType } from "../sharedPropTypes"
-import { getMetricTags, getReportTags, getSubjectName, nrMetricsInReport, visibleMetrics } from "../utils"
+import { getReportTags, getSubjectName, nrMetricsInReport, sum, visibleMetrics } from "../utils"
 import { Tag } from "../widgets/Tag"
-import { metricStatusOnDate } from "./report_utils"
-
-function summarizeMetricsOnDate(metrics, measurements, date, dataModel) {
-    const summary = { red: 0, yellow: 0, green: 0, blue: 0, grey: 0, white: 0 }
-    Object.entries(metrics).forEach(([metric_uuid, metric]) => {
-        const status = metricStatusOnDate(metric_uuid, metric, measurements, date, dataModel)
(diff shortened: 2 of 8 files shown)


components/frontend/src/report/Report.test.js
```diff
@@ -167,15 +167,3 @@ it("navigates to subject", async () => {
...
-it("does not navigate to a subject that is hidden", async () => {
-    history.push("?metrics_to_hide=no_issues")
-    const mockScroll = jest.fn()
-    window.HTMLElement.prototype.scrollIntoView = mockScroll
-    const mockScrollBy = jest.fn()
-    window.scrollBy = mockScrollBy
-    renderReport({ reportToRender: report })
-    fireEvent.click(screen.getAllByText(/Subject title/)[0])
-    expect(mockScroll).not.toHaveBeenCalled()
-    expect(mockScrollBy).not.toHaveBeenCalled()
```
fniessink, line 180: This test is now obsolete because empty subjects are now hidden from the dashboard.

components/frontend/src/report/ReportDashboard.js
```diff
@@ -8,45 +8,22 @@ import { IssuesCard } from "../dashboard/IssuesCard"
...
-    const summary = { red: 0, yellow: 0, green: 0, blue: 0, grey: 0, white: 0 }
-    Object.values(report.subjects).forEach((subject) => {
-        Object.entries(subject.metrics).forEach(([metric_uuid, metric]) => {
-            if (getMetricTags(metric).indexOf(tag) >= 0) {
-                const status = metricStatusOnDate(metric_uuid, metric, measurements, date, dataModel)
-                summary[STATUS_COLORS[status]] += 1
-            }
-        })
-    })
-    return summary
```
fniessink, line 36: Code moved to report_utils.js.

components/frontend/src/report/ReportsOverviewDashboard.js
```diff
@@ -7,39 +7,10 @@ import { CardDashboard } from "../dashboard/CardDashboard"
...
-            Object.entries(metrics).forEach(([metric_uuid, metric]) => {
-                if (getMetricTags(metric).indexOf(tag) >= 0) {
-                    const status = metricStatusOnDate(metric_uuid, metric, measurements, date, dataModel)
-                    summary[STATUS_COLORS[status]] += 1
-                }
-            })
-        })
-    })
-    return summary
-}
```
fniessink, line 42: Code moved to report_utils.js.

components/frontend/src/report/ReportsOverviewDashboard.test.js
```diff
@@ -100,6 +100,7 @@ it("hides the required actions cards", async () => {
 })
 
 it("hides metrics not requiring action", async () => {
+    history.push("?metrics_to_hide=all")
```
fniessink, line 103: Test was working, but this change increases the test coverage.

components/frontend/src/utils.js
```diff
@@ -237,14 +237,9 @@ export function visibleMetrics(metrics, metricsToHide, hiddenTags) {
...
-    Object.entries(metrics).forEach(([metric_uuid, metric]) => {
-        if (hideMetric(metric, metricsToHide, hiddenTags)) {
-            return
-        }
-        visible[metric_uuid] = metric
-    })
-    return visible
+    return Object.fromEntries(
+        Object.entries(metrics).filter(([_, metric]) => !hideMetric(metric, metricsToHide, hiddenTags)),
+    )
```
fniessink, line 242: Refactor by using `Object.fromentries` instead of manually constructing the filtered object.

components/frontend/src/utils.js
```diff
@@ -405,6 +400,15 @@ export function slugify(name) {
     return `#${name?.toLowerCase().replaceAll(" ", "-").replaceAll("(", "").replaceAll(")", "").replaceAll("/", "")}`
 }
 
+export function addCounts(object1, object2) {
+    // Assuming object1 and object2 are objects of the form {key1: count1, key2: count2, ...}, add them together
```
wkoot, line 404: Is it correct that we're assuming that both objects have the exact same set of keys? And values of 0 are never omitted for instance?

wkoot (approved):

components/frontend/src/utils.js
fniessink: Yeah, it is currently a correct assumption, but maybe you meant that the implementation of `addCounts` should check the assumption?

wkoot: I was wondering if it would be an idea to loop over the unique `set(object1.keys + object2.keys)`

fniessink: I think it's safer to just assert that both objects have the same keys. I'll add that.

PR #9719, opened 2024-09-13
Partially replace Semantic UI React with MUI.

### LINKED ISSUE #9550: POC: Replace a small part of Semantic UI React with MUI
Quality-time uses React and Semantic UI React (SUIR) for the frontend. SUIR seems dead (https://github.com/Semantic-Org/Semantic-UI-React/issues/4230).

Obsoletes:
- #8764
- #3233 
- #7251

POC:
- Add MUI to the frontend.
- Replace a few components by MUI equivalents.

fniessink:
This PR changes a few components in the UI to use MUI instead of Semantic UI React. It also uses MUI for dark/light UI mode.


components/frontend/src/AppUI.js
```diff
@@ -42,20 +41,7 @@ export function AppUI({
     set_user,
     user,
 }) {
-    const [uiMode, setUIMode] = useLocalStorageState("ui_mode", { defaultValue: "follow_os" })
-    useEffect(() => {
```
wkoot, line 46: Nice simplify

components/frontend/src/App.js
```diff
@@ -237,27 +244,29 @@ class App extends Component {
...
-                report_uuid={this.state.report_uuid}
-                reports={this.state.reports}
-                reports_overview={this.state.reports_overview}
-                set_user={(username, email, sessionExpirationDateTime) =>
-                    this.setUserSession(username, email, sessionExpirationDateTime)
-                }
-                user={this.state.user}
-            />
+            <ThemeProvider theme={theme}>
+                <AppUI
```
wkoot, line 248: This is only indentation change right? The github viewer doesn't highlight it :(

components/frontend/src/dashboard/LegendCard.js
```diff
@@ -1,34 +1,21 @@
...
-        </List.Item>
+        <ListItem key={status} dense={true} sx={{ padding: "0px" }}>
+            <StatusIcon status={status} size="small" />
+            &nbsp;
+            <ListItemText primary={STATUS_SHORT_NAME[status]} />
+        </ListItem>
     ))
+
     return (
-        <Card tabIndex="0" className="legend">
```
wkoot, line 25: Do we no longer have the tab index logic, or is it automatic?

components/frontend/src/dashboard/MetricSummaryCard.js
```diff
@@ -90,8 +89,8 @@ export function MetricSummaryCard({ header, onClick, selected, summary, maxY })
         width: Math.max(bbWidth, 1), // Prevent "Failed prop type: Invalid prop range supplied to VictoryBar"
     }
     return (
-        <FilterCard onClick={onClick} selected={selected}>
-            <div ref={ref} style={{ height: "72%" }} aria-label={ariaChartLabel(summary)}>
+        <DashboardCard onClick={onClick} selected={selected} title={header}>
+            <div ref={ref} aria-label={ariaChartLabel(summary)} style={{ height: "72%" }}>
```
wkoot: Where does the 72% height come from, and can it be moved to css?

tests/application_tests/src/test_report.py
```diff
@@ -42,14 +42,19 @@ def __call__(self, driver):
...
     def setUp(self):
         """Override to setup the driver."""
         firefox_options = webdriver.FirefoxOptions()
         firefox_options.add_argument("--headless")
         self.driver = webdriver.Remote("http://selenium:4444", options=firefox_options)
         self.driver.implicitly_wait(10)
         self.wait = WebDriverWait(self.driver, 10)
-        self.driver.get("http://www:8080?hide_toasts=true")  # Hide toasts so they don't block other visual elements
+        # Hide things so they don't block other visual elements
+        self.driver.get("http://www:8080?hide_toasts=true&hidden_cards=legend")
```
wkoot: Why `&hidden_cards=legend`? Was that a code change in this PR that I missed, or was it already something blocking other elements?

tests/application_tests/src/test_report.py
```diff
@@ -64,14 +69,18 @@ def login(self):
...
     def test_title(self):
         """Test the title."""
         self.assertTrue(expect.title_contains("Quality-time"))
 
     def test_open_report(self):
-        """Test that the first report can be opened."""
-        report = self.driver.find_elements(By.CLASS_NAME, "card")[-1]  # The last card is a report
-        report_title = report.find_element(By.CLASS_NAME, "header")
+        """Test that a report can be opened."""
+        report = self.dashboard_cards()[-1]
```
wkoot:
```suggestion
        report = self.dashboard_cards()[-1]  # The last card is a report
```

tests/application_tests/src/test_report.py
```diff
@@ -89,32 +98,31 @@ def test_login_and_logout(self):
     def test_add_report(self):
         """Test that a logged-in user can add a report."""
         self.login()
-        nr_reports = len(self.driver.find_elements(By.CLASS_NAME, "card"))
+        nr_cards = len(self.dashboard_cards())
```
wkoot, line 100: Hmm, but these renames would suggest the test be called `test_add_card`. It's kinda sneaky because we never really refer to cards as cards in the UI, they were even read-only reports :D The "button.primary" is even sneaky because there are visually two buttons, but the second one isn't actually a html button element! Oh well, I guess functional/application tests are usually a bit dirty - making the variable naming more consistent here probably isn't doable without a refactor.

tests/application_tests/src/test_report.py
```diff
@@ -89,32 +98,31 @@ def test_login_and_logout(self):
...
+                report_file.write(human_readable_axe_report)
         except OSError:
             self.fail("Could not write axe violations report")
 
-        # If there are moe violations than expected, output the readable report data
+        # If there are more violations than expected, output the human readable report
         # Fixing the axe violations is on the backlog: https://github.com/ICTU/quality-time/issues/6354
-        current_number_of_axe_violations = 6
-        self.assertLessEqual(len(violation_results), current_number_of_axe_violations, readable_report)
+        current_number_of_axe_violations = 7
```
wkoot, line 127: Uh-oh! What's the newly introduced violation?

components/frontend/src/dashboard/DashboardCard.js
```diff
@@ -0,0 +1,43 @@
...
+        />
+    )
+    return (
+        <Card
+            elevation={2}
+            onClick={onClick}
+            sx={{
+                border: 1,
+                borderColor: color,
+                height: "100%",
```
wkoot: Are these 100%'s related to the other 72% ? Seems weird having to state 100%

components/frontend/src/App.js
fniessink: Yes, this delta only adds the MUI `ThemeProvider`.

components/frontend/src/dashboard/LegendCard.js
fniessink: If a MUI `Card` has a `CardActionArea` it gets keyboard focus automatically.

components/frontend/src/dashboard/MetricSummaryCard.js
fniessink: Because the donut charts are round and the cards are higher than wide, the chart needs to have a height that is equal to the width of the card. I guess (can't remember) the 72% is the default ratio between the height and width of the card. Anyway, I'm not setting the width to 100%, that also works.

tests/application_tests/src/test_report.py
fniessink: I had some problems that I thought this would work around. I think this change can be reverted.

tests/application_tests/src/test_report.py
fniessink: Not all cards are reports, e.g. the Legend card. I'll add a comment to explain.

tests/application_tests/src/test_report.py
fniessink: I don't know which the previous six were :-)  I think we should get rid of these after the migration to MUI is completed.

components/frontend/src/dashboard/DashboardCard.js
fniessink: No, not related. This makes sure the component fills the available space in its container. I've added a comment.

wkoot (approved):

PR #9792, opened 2024-09-18
Don't stop retrieving data from Dependency-Track after the first page.

### LINKED ISSUE #9751: Quality-time does not retrieve more than one page from Dependency Track
Tasks:
- [x] Fix pagination code
- [x] Fix async loop
- [x] Add a warning if the configured project does not exist (e.g. typo)

fniessink:
Fixes #9751.

### DIFF
--- a/components/collector/src/source_collectors/dependency_track/base.py
+++ b/components/collector/src/source_collectors/dependency_track/base.py
@@ -4,7 +4,7 @@ from collections.abc import AsyncIterator
 from typing import TypedDict, cast
 
 from base_collectors import SourceCollector
-from collector_utilities.functions import match_string_or_regular_expression
+from collector_utilities.functions import add_query, match_string_or_regular_expression
 from collector_utilities.type import URL, Response
 from model import SourceResponses
 
@@ -24,7 +24,7 @@ class DependencyTrackBase(SourceCollector):
     """Dependency-Track base class."""
 
     # Max page size is 100, see https://github.com/DependencyTrack/dependency-track/issues/209.
-    # However, we use a lower number because pages may contain a lot of data, # especially for the
+    # However, we use a lower number because pages may contain a lot of data, especially for the
     # projects and findings endpoints.
     PAGE_SIZE = 25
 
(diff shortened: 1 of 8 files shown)


components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -59,7 +59,7 @@ async def _get_project_uuids(self) -> dict[str, str]:
 
     async def _get_projects(self) -> AsyncIterator[DependencyTrackProject]:
         """Return the Dependency-Track projects."""
-        projects_api = URL(await self._api_url() + "/project")
+        projects_api = URL(await DependencyTrackBase._api_url(self) + "/project")
```
wkoot, line 62: Why was this needed?

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -44,8 +44,8 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
         responses = SourceResponses()
         for url in urls:
             page_nr = 1  # Page numbers start at 1
-            total_count = self.PAGE_SIZE + 1  # Total count is still unknown, make it so big we get at least one page
-            while page_nr * self.PAGE_SIZE < total_count:
+            total_count = 1  # Total count is still unknown, make sure we retrieve at least one page
+            while (page_nr - 1) * self.PAGE_SIZE < total_count:
                 offsetted_url = URL(f"{url}{"&" if "?" in url else "?"}pageSize={self.PAGE_SIZE}&pageNumber={page_nr}")
```
wkoot: This url string munging is a bit dirty, we should have a helper method to merge http GET params from the query.
There are a lot of helpers provided by `urllib.parse`, avoiding to worry about a `#` fragment in the url.

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Because the code was wrong :-) 

If a subclass would override the `_api_url` method to, for example, access the `/issues` endpoint, this method would also call that overridden method and try to access `/issues/project`. 

It was just a coincidence we haven't triggered this accident waiting to happen so far: only `DependencyTrackSourceUpToDateness` overrides `_api_url` at the moment, but it doesn't call `_get_projects`. The other subclasses do call `_get_projects` but don't override `_api_url`. Pfew!

components/collector/src/source_collectors/dependency_track/base.py
fniessink: I don't see a helper in `urllib.parse` that lets one add one query argument to a URL. But I agree a helper is useful. I'll add one.

PR #9795, opened 2024-09-18
Remove 'manual number' as supported source for 'test cases'.

### LINKED ISSUE #9793: Test cases metric has manual number as supported source
**Describe the bug**
Test cases metric has manual number as supported source, but this makes no sense for this metric.

**To Reproduce**
See https://quality-time.readthedocs.io/en/v5.15.0/reference.html#test-cases

**Expected behaviour**
Manual number is not listed as supported source for the test cases metric and can't be chosen as source.

 - Quality-time version v5.15.0

fniessink:
The 'test cases' metric had 'manual number' as supported source for the metric, but that could never work since the metric needs sources that contain test cases with identifiers so they can be matched.

Fixes #9793.


docs/src/changelog.md
```diff
@@ -31,6 +31,10 @@ If your currently installed *Quality-time* version is not v5.15.0, please first
 
 - Start replacing Semantic UI React with Material UI as frontend component library. This migration will probably span multiple releases. Until the migration is finished, the UI may contain inconsistent elements, such as fonts, colors, and icons. Closes [#9550](https://github.com/ICTU/quality-time/issues/9550).
 
+### Removed
+
+- The 'test cases' metric had 'manual number' as supported source for the metric, but that could never work since the metric needs sources that contain test cases with identifiers so they can be matched. Fixes [#9793](https://github.com/ICTU/quality-time/issues/9793).
```
wkoot: Since we don't include a migration, should we tell operators to address existing data before upgrading?
If this could never work, is it safe to just drop all these sources?

fniessink: I think it's safe to assume that no one is using manual number as source for the test cases metric, but I'll add a note to the change log nevertheless.

fniessink: I decided to add a migration anyway, it wasn't very much work. I'd appreciate another review though.

wkoot (commented):
lgtm

PR #9849, opened 2024-09-24
Show hour glass icon when a change triggers a measurement.

### LINKED ISSUE #9631: The hour glass icon does not appear when copying a source
### Scenarios that already work

The hour glass icon should appear when:
- copying an existing source to a metric that already has a source

The hour glass should not appear when:
- removing the last source (the measurement value becomes `?` immediately)
- removing a source with a source remaining (the measurement value is changed immediately)
- adding a new source to a metric (the source needs to be configured first)

### Scenarios to implement

The hour glass icon should appear when:
- [x] copying an existing source to a metric that has no sources
- [x] copying a metric with a source
- [x] copying a subject with one or more metrics with a source
- [x] copying a report with one or more subjects with one or more metrics with a source

fniessink:
Fixes #9631.

### DIFF
--- a/components/frontend/src/utils.js
+++ b/components/frontend/src/utils.js
@@ -4,6 +4,7 @@ import { PERMISSIONS } from "./context/Permissions"
 import { defaultDesiredResponseTimes } from "./defaults"
 import { STATUSES_NOT_REQUIRING_ACTION } from "./metric/status"
 import {
+    datePropType,
     metricPropType,
     metricsPropType,
     metricsToHidePropType,
@@ -96,6 +97,16 @@ export function getMetricUnit(metric, dataModel) {
     return metric.unit || dataModel.metrics[metric.type].unit || ""
 }
 
+export function isMeasurementOutdated(metric) {
+    if (metric.latest_measurement) {
+        return metric.latest_measurement.outdated ?? false
+    }
+    return Object.keys(metric.sources ?? {}).length > 0 // If there are sources, measurement is needed
+}
+isMeasurementOutdated.propTypes = {
+    metric: metricPropType,
+}
+
 export function isMeasurementRequested(metric) {
     if (metric.measurement_requested) {
         if (metric.latest_measurement) {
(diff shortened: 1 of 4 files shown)


components/frontend/src/utils.test.js
```diff
@@ -509,3 +511,18 @@ it("returns whether a measurement is requested for the metric", () => {
...
+    expect(isMeasurementStale({})).toBe(false)
+    expect(isMeasurementStale({ latest_measurement: { end: "2024-09-23" } })).toBe(true)
+    expect(isMeasurementStale({ latest_measurement: { end: "2024-09-23" } }, new Date("2024-09-24"))).toBe(true)
+    expect(isMeasurementStale({ latest_measurement: { end: "2024-09-24" } }, new Date("2024-09-23"))).toBe(false)
+})
+
+it("returns whether a metric's measurement is outdated", () => {
+    expect(isMeasurementOutdated({})).toBe(false)
+    expect(isMeasurementOutdated({ sources: {} })).toBe(false)
+    expect(isMeasurementOutdated({ sources: { source_uuid: {} } })).toBe(true)
```
wkoot, line 536: This test case is a bit magical without explanatory comments

components/frontend/src/utils.js
```diff
@@ -106,6 +117,19 @@ export function isMeasurementRequested(metric) {
     return false
 }
 
+export function isMeasurementStale(metric, reportDate) {
+    if (!metric.latest_measurement) {
+        return false
+    }
+    const end = new Date(metric.latest_measurement.end)
+    const now = reportDate ?? new Date()
+    return now - end > MILLISECONDS_PER_HOUR // No new measurement for more than one hour means something is wrong
```
wkoot, line 126: Does it not make more sense to implement such logic in the backend, where we for instance also know the collector sleep time? Picking an hour seems a bit random

fniessink: The collector doesn't know `now` though, i.e. when the user is looking at the measurement. I guess we should make the measurement frequency (`COLLECTOR_MEASUREMENT_FREQUENCY`) available to the frontend. 

By the way, we have documented that this is hardcoded in the frontend: https://quality-time.readthedocs.io/en/latest/deployment.html#configuring-measurement-frequency-optional

components/frontend/src/utils.test.js
fniessink: I'll add some comments!

PR #9850, opened 2024-09-24
Deploy to Kubernetes as part of pipeline

### LINKED ISSUE #8204: Deploy to Kubernetes as part of pipeline
We have a Helmchart in the repo, but don't test it. 

Add Kubernetes deployment to CircleCI and/or GitHub Actions and then run the application and feature tests against the deployed application. 

We don't want to have to install an LDAP container or the testdata container, so check whether the Quality-time demo report gets data from Sonarcloud.

wkoot:
Closes #8204

### DIFF
--- a/.github/workflows/api_server.yml
+++ b/.github/workflows/api_server.yml
@@ -1,3 +1,4 @@
+---
 name: API-server
 
 on: [push]
--- a/.github/workflows/application-tests-quality.yml
+++ b/.github/workflows/application-tests-quality.yml
@@ -1,3 +1,4 @@
+---
 name: Application tests quality
 
 on: [push]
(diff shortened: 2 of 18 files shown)


fniessink (commented):
I don't understand where/how SonarQube is deployed? We're generating passwords for SonarQube and Progress, but why if they are not used?

wkoot:
> I don't understand where/how SonarQube is deployed? 

The final step `Build and run chart` deploys to a local kubernetes instance which is ran on `minikube`.
See also the corresponding github action: https://github.com/marketplace/actions/setup-minikube :-)
After installing the helm chart, the test waits for ready checks and all deployments being marked as available.

PR #9873, opened 2024-09-26
Fix incorrect tests count for JUnit XML.

### LINKED ISSUE #9872: Incorrect test counts when JUnit XML test case names are not unique
**Describe the bug**
Quality-time reports wrong test counts if JUnit XML has test case names that are not unique

**Expected behaviour**
Use both `className` and `name` attributes of `testcase` to generate entity key.

fniessink:
The 'tests' metric with JUnit XML files as source would report incorrect results if the JUnit XML files contain test case names that are not unique across test suites.

Fixes #9872.

### DIFF
--- a/components/collector/src/source_collectors/junit/tests.py
+++ b/components/collector/src/source_collectors/junit/tests.py
@@ -39,5 +39,8 @@ class JUnitTests(XMLFileSourceCollector):
     @staticmethod
     def __entity(case_node: Element, case_result: str) -> Entity:
         """Transform a test case into a test case entity."""
+        class_name = case_node.get("classname", "")
         name = case_node.get("name", "<nameless test case>")
-        return Entity(key=name, name=name, class_name=case_node.get("classname", ""), test_result=case_result)
+        key = f"{class_name}:{name}"
+        old_key = name  # Key was changed after Quality-time 5.16.1, enable migration of user entity data
+        return Entity(key=key, old_key=old_key, name=name, class_name=class_name, test_result=case_result)
--- a/components/collector/tests/source_collectors/junit/test_tests.py
+++ b/components/collector/tests/source_collectors/junit/test_tests.py
@@ -12,11 +12,11 @@ class JUnitTestReportTest(JUnitCollectorTestCase):
         """Extend to set up JUnit test data."""
         super().setUp()
         self.expected_entities = [
-            {"key": "tc1", "name": "tc1", "class_name": "cn", "test_result": "passed"},
-            {"key": "tc2", "name": "tc2", "class_name": "cn", "test_result": "passed"},
-            {"key": "tc3", "name": "tc3", "class_name": "cn", "test_result": "failed"},
-            {"key": "tc4", "name": "tc4", "class_name": "cn", "test_result": "errored"},
-            {"key": "tc5", "name": "tc5", "class_name": "cn", "test_result": "skipped"},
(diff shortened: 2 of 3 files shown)


components/collector/tests/source_collectors/junit/test_tests.py
```diff
@@ -45,3 +45,15 @@ async def test_empty_test_suites(self):
         """Test that a JUnit XML file with an empty testsuites node works."""
         response = await self.collect(get_request_text=self.JUNIT_XML_EMPTY_TEST_SUITES)
         self.assert_measurement(response, value="0", total="0")
+
+    async def test_repeated_test_case_names(self):
+        """Test that repeated test case names are distinguished."""
+        junit_xml_repeated_test_case_name = """<testsuites>
+        <testsuite timestamp="2009-12-19T17:58:59">
+            <testcase name="tc1" classname="cn1"/>
```
wkoot, line 53: Is it not possible for JUnit XML files to contain the same combination of test name + class name across files?
E.g. `file1`: `name=1 class=1` and  `file2` having the same entry: `name=1 class=1` ?

components/collector/src/source_collectors/junit/tests.py
```diff
@@ -39,5 +39,8 @@ def _include_entity(self, entity: Entity) -> bool:
     @staticmethod
     def __entity(case_node: Element, case_result: str) -> Entity:
         """Transform a test case into a test case entity."""
+        class_name = case_node.get("classname", "")
         name = case_node.get("name", "<nameless test case>")
-        return Entity(key=name, name=name, class_name=case_node.get("classname", ""), test_result=case_result)
+        key = f"{class_name}:{name}"
+        old_key = name  # Key was changed after Quality-time 5.16.1, enable migration of user entity data
```
wkoot, line 45: How is the data migrated when it's the same value as `name`?

components/collector/tests/source_collectors/junit/test_tests.py
fniessink: I don't know. There's no official XSD for JUnit XML. Just a bunch of conventions. In the example data I received for this bug report test cases only had two attributes: `classname` and `name`, so there's not much we can do if the combination of those two would not be unique. Maybe except for throwing an error. Which we maybe should do in every collector when entity keys are not unique. I've opened #9883 for that.

components/collector/src/source_collectors/junit/tests.py
fniessink: The data is migrated by the collector when creating a measurement in https://github.com/ICTU/quality-time/blob/8536d0bd1075f042ad4b47e7e59c6a682beacc85/components/collector/src/database/measurements.py#L16 via `copy_entity_user_data`. Now sure what you mean by 'when it's the same value as `name`'?

wkoot: I don't see where `old_key` is being used, I thought that perhaps you intended to write a migration that wasn't included in this PR (yet)? But if the value of `old_key` is the same as the value of `name`, then I don't see why this migration wouldn't use `name` instead of `old_key` directly

fniessink: The `old_key` is used in the `Source.copy_entity_user_data` method which is called by `Measurement.copy_entity_user_data` which in turn is called by `create_measurement` when a new measurement is created.

PR #9884, opened 2024-09-27
Make GitLabJobs collector use datetime instead of date objects

### LINKED ISSUE #5313: Job runs metric shows jobs outside time window
**To Reproduce**
1. Set a metric 'Job runs within time period' to 7 days (in source config)
2. Click on results tab (metric entities)
3. see job runs from earlier that 7 days ago

wkoot:
Fixes #5313

### DIFF
--- a/components/collector/src/metric_collectors/change_failure_rate.py
+++ b/components/collector/src/metric_collectors/change_failure_rate.py
@@ -3,7 +3,7 @@
 from collections.abc import Sequence
 from datetime import datetime
 from itertools import pairwise
-from typing import ClassVar
+from typing import ClassVar, cast
 
 from base_collectors import MetricCollector
 from collector_utilities.date_time import MAX_DATETIME, parse_datetime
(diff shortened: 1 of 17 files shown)


fniessink (commented):
We don't we stop putting the build_date in the entity and just only use the build_datetime? And simply let frontend format the build_datetime as a date?

wkoot:
> We don't we stop putting the build_date in the entity and just only use the build_datetime? And simply let frontend format the build_datetime as a date?

The datetime object cannot be (de)serialized correctly, throwing an error. This is what I had originally built, but it didn't work :)

fniessink:
> > We don't we stop putting the build_date in the entity and just only use the build_datetime? And simply let frontend format the build_datetime as a date?
> 
> 
> 
> The datetime object cannot be (de)serialized correctly, throwing an error. This is what I had originally built, but it didn't work :)

Right, but I was thinking that we could store the isoformatted version of the build_datetime instead of the isoformatted version of the build_date.

wkoot:
> Right, but I was thinking that we could store the isoformatted version of the build_datetime instead of the isoformatted version of the build_date.

I tried this, but I ran into timing offset issues. The added test case `test_job_lookback_days_on_edge` was specifically investigating that. Even when using isoformatted datetime, sometimes the code would behave unexpectedly and the test would fail. This was particularly apparent in the first two hours before/after.
Because the current code only converts to a date, the offset was even larger. However, it would have unexpected cutoff moments that aren't at midnight or the two surrounding hours. I did not fully investigate what's going on there, because using native types was an easy solution.

PR #9932, opened 2024-10-04
Use uv.

### LINKED ISSUE #9868: Use uv
Replace the following tools with uv:
- [x] pip
- [x] pipx: `uv tool run`
- [x] pip-compile

fniessink:
Note that `uv` names the virtual env folder `.venv` while we have been using `venv` as default. Developers may have to recreate the virtual environments on their laptop.


.circleci/config.yml
```diff
@@ -24,8 +24,8 @@ jobs:
           esac
           cd $component
           mkdir -p build
-          python -m venv venv
-          . venv/bin/activate
```
wkoot, line 28: Why do we not want to work within a virtual env anymore, but only call the binary within the venv?

.circleci/config.yml
```diff
@@ -91,7 +90,7 @@ jobs:
           export COMPOSE_FILE=docker/docker-compose.yml:docker/docker-compose.ci.yml
           docker compose build && docker compose up -d
           docker ps
-          docker run -it -w `pwd` -v `pwd`:`pwd` --network=container:docker-www-1 python:3.12.5-bookworm tests/application_tests/ci/test.sh
+          docker run -it -w `pwd` -v `pwd`:`pwd` --network=container:docker-www-1 ghcr.io/astral-sh/uv:python3.12-bookworm tests/application_tests/ci/test.sh
```
wkoot, line 93: Is this a, specifically for uv, altered python container, or just an alternative registry backed by github?

.circleci/config.yml
```diff
@@ -110,6 +109,7 @@ jobs:
     steps:
       - checkout
       - run: |
+          pip install uv
```
wkoot, line 112: Is there no `uv venv` needed?

.github/workflows/api_server.yml
```diff
@@ -5,24 +5,28 @@ on: [push]
...
-      run: |
-        cd components/api_server
-        ci/quality.sh
+      - uses: actions/checkout@v4.2.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "0.4.18"
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
```
wkoot: Why do we need setup python when we already install uv, or does this action use the uv environment?

.github/workflows/api_server.yml
```diff
@@ -5,24 +5,28 @@ on: [push]
...
+        with:
+          python-version-file: "components/api_server/pyproject.toml"
+      - name: Install dependencies
```
wkoot: Why is this needed when the previous step already used the `pyproject.toml` to set up the environment?

.github/workflows/application-tests.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         run: |
           mkdir -p build
           docker compose --file docker/docker-compose.yml --file docker/docker-compose.ci.yml --project-name quality-time up --build --detach --wait
-          docker run -t -w `pwd` -v `pwd`:`pwd` --network=container:quality-time-www-1 python:3.12.5-bookworm tests/application_tests/ci/test.sh
+          docker run -t -w `pwd` -v `pwd`:`pwd` --network=container:quality-time-www-1 ghcr.io/astral-sh/uv:python3.12-bookworm tests/application_tests/ci/test.sh
```
wkoot, line 20: Do we still need debian, or can we go to alpine?

.github/workflows/notifier.yml
```diff
@@ -5,24 +5,28 @@ on: [push]
...
-        ci/quality.sh
+      - uses: actions/checkout@v4.2.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "0.4.18"
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: "components/collector/pyproject.toml"
```
wkoot:
```suggestion
          python-version-file: "components/notifier/pyproject.toml"
```

.github/workflows/release-quality.yml
```diff
@@ -5,20 +5,24 @@ on: [push]
...
-        ci/quality.sh
+      - uses: actions/checkout@v4.2.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "0.4.18"
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: "components/collector/pyproject.toml"
```
wkoot:
```suggestion
          python-version-file: "release/pyproject.toml"
```

.github/workflows/shared_code.yml
```diff
@@ -5,24 +5,28 @@ on: [push]
...
-        ci/quality.sh
+      - uses: actions/checkout@v4.2.0
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "0.4.18"
+      - name: "Set up Python"
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: "components/collector/pyproject.toml"
```
wkoot:
```suggestion
          python-version-file: "components/shared_code/pyproject.toml"
```

.github/workflows/sonarcloud.yml
```diff
@@ -18,11 +18,16 @@ jobs:
         if: env.SONAR_TOKEN != null
         with:
           fetch-depth: 0 # Disabling shallow clone is recommended for improving relevancy of reporting
-      - name: Set up Python
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        if: env.SONAR_TOKEN != null
```
wkoot, line 22: Can't we add this conditional on the entire workflow, or step 1 which aborts the workflow?

.github/workflows/sonarcloud.yml
```diff
@@ -18,11 +18,16 @@ jobs:
...
+        with:
+          version: "0.4.18"
+      - name: "Set up Python"
         uses: actions/setup-python@v5
         if: env.SONAR_TOKEN != null
         with:
-          python-version: "3.12"
+          python-version-file: "components/collector/pyproject.toml"
```
wkoot: Oh, is this really only pulling the python version from the toml!

ci/uvx-base.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+source base.sh
+
+run_uvx() {
+    # Look up the version of the command using the spec function and run the command using uvx.
+    command_spec=$(spec $1 @)
```
wkoot, line 7: I don't get it, where does the `@` come from? This produces something like `package@v1.2.3` - is that the syntax for uvx?
Also note that the pypi page for uvx links to a `uvx2` repo, which is apparently renamed to `uvenv` to avoid naming collision with a new uv command

ci/spec.py
```diff
@@ -14,8 +14,8 @@ def spec(package: str, pyproject_toml_path: Path) -> str:
         pyproject_toml = tomllib.load(pyproject_toml_file)
     tools = pyproject_toml["project"]["optional-dependencies"]["tools"]
     package_specs = [spec for spec in tools if spec.split("==")[0] == package]
-    return package_specs[0] if package_specs else ""
+    return sep.join(package_specs[0].split("==")) if package_specs else ""
```
wkoot, line 17: This method is kinda dirty, can't we comprehend the toml file directly? The `tomllib` is a std lib since python 3.11

tests/feature_tests/ci/test.sh
```diff
@@ -11,34 +11,31 @@ export PROXY_PORT=8080
...
 cd tests/feature_tests
-python3 -m venv venv
-. venv/bin/activate
+uv venv
 ci/pip-install.sh
 cd ../..
 sleep 5  # Give components time to start up
-coverage erase
-coverage run -m behave --format pretty "${1:-tests/feature_tests/src/features}"
+tests/feature_tests/.venv/bin/coverage erase
```
wkoot, line 28: Why do we `cd ../..` above, when everything is then calling the previous path? If it's only for the combine step, perhaps that one can be prefixed?

ci/uvx-base.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+source base.sh
+
+run_uvx() {
```
wkoot, line 5: Note that the `uvx` project has been renamed to `uvenv` due to naming collision

ci/quality-base.sh
```diff
@@ -1,55 +1,54 @@
...
-    run_pipx fixit lint $(python_files_and_folders)
+    run_uvx fixit lint $(python_files_and_folders)
 }
 
 run_mypy() {
     PYTHON_FILES_AND_FOLDERS=$(python_files_and_folders)
     # Run mypy with or without pydantic plugin depending on whether pydantic is listed as dependency in the tools
     # section of the optional dependencies in the pyproject.toml file.
-    pydantic_spec=$(spec pydantic)
+    pydantic_spec=$(spec pydantic ==)
```
wkoot, line 23: This seems to be the only place where spec is called with a `==` separator, why the distinction?

ci/quality-base.sh
```diff
@@ -1,55 +1,54 @@
...
     if [[ "$pydantic_spec" == "" ]]; then
-        run_pipx mypy --python-executable=$(which python) $PYTHON_FILES_AND_FOLDERS
+        run_uvx mypy --python-executable=.venv/bin/python $PYTHON_FILES_AND_FOLDERS
     else
-        # To use the pydantic plugin, we need to first install mypy and then inject pydantic
-        run pipx install --force $(spec mypy) # --force works around this bug: https://github.com/pypa/pipx/issues/795
-        run pipx inject mypy $pydantic_spec
-        run $PIPX_BIN_DIR/mypy --python-executable=$(which python) $PYTHON_FILES_AND_FOLDERS
+        run uvx --with $pydantic_spec $(spec mypy @) --python-executable=.venv/bin/python $PYTHON_FILES_AND_FOLDERS
```
wkoot, line 27: I don't get it, doesn't this produce something like `--with pydantic==1.2.3 mypy@4.5.6` ?

.circleci/config.yml
fniessink: `uv venv` creates the virtual env.

.circleci/config.yml
fniessink: This is a Python 3.12 container with uv preinstalled. We could use a vanilla Python container, but then the `test.sh` script would have to install uv.

.circleci/config.yml
fniessink: No, the `test.sh` script creates the virtual env.

.github/workflows/api_server.yml
fniessink: Good question. I'll test without it and see if uv correctly downloads Python.

fniessink: That seems to work. Nice.

.github/workflows/api_server.yml
fniessink: Indeed not needed.

.github/workflows/application-tests.yml
fniessink: I'll test.

fniessink: Alas: `exec tests/application_tests/ci/test.sh: no such file or directory`.

.github/workflows/sonarcloud.yml
fniessink: Testing...

ci/uvx-base.sh
fniessink: uvx is a bit inconsistent with the package spec, it uses both @ and ==: `uvx --with package1==a.b.c package2@x.y.z`. Hence the need to pass the separator to the `spec` function.

Note that `uvx` is not a separate package but an alias for `uv tool run`.

ci/uvx-base.sh
fniessink: We don't use the uvx project.

ci/spec.py
fniessink: We use tomllib to read the toml file, but the package specs are strings of the form `"package==x.y.z"`. tomllib doesn't help with that.

tests/feature_tests/ci/test.sh
fniessink: I don't remember exactly but it's probably to make sure the coverage files don't end up in `tests/feature_tests`.

ci/quality-base.sh
fniessink: See above.

ci/quality-base.sh
fniessink: Yup!

.github/workflows/sonarcloud.yml
fniessink: Nope. Workflow level environment variables aren't available in the job: https://github.com/actions/runner/issues/1661.

.circleci/config.yml
wkoot: Yes but `source .venv/bin/activate` would still source that newly created venv

.github/workflows/application-tests.yml
wkoot: Hmm, where does it go wrong? Does it lack a bash interpreter perhaps? If so, `RUN apk add --no-cache bash` should help

ci/uvx-base.sh
wkoot: ouch, no wonder I couldn't understand the docs!

.circleci/config.yml
fniessink: Yes, but `ci/pip-install.sh` runs `uv pip install` which uses the venv without activation. See https://docs.astral.sh/uv/pip/environments/#discovery-of-python-environments.

`ci/unittest.sh` runs `.venv/bin/coverage` (I'll have to check whether I can get `uvx coverage` working) and `ci/quality.sh` run everything via uvx so they don't need an activated virtual env either.

I think we're very close to never having to activate a virtual envs anymore. The only thing I haven't found out is how to run the application (e.g. `python src/quality_time_server.py`) without activating the venv. `uv run src/quality_time_server.py` insists on creating a `uv.lock` file which we don't need as we have compiled requirements files.

fniessink: > I'll have to check whether I can get `uvx coverage` working

Alas, that isn't possible. `mypy` has a `--python-executable` flag for this purpose, but `coverage` does not.

PR #9965, opened 2024-10-08
Fix empty subject column in missing metric details.

### LINKED ISSUE #9854: Subject column empty
**Describe the bug**
Same Quality-time version 5.16
Same environment
Missing metrics → entity tab → one instance subject (name) column empty and other instance it is filled

**Screenshots**
image
image

fniessink:
The subject column in the measurement details of the 'missing metrics' metric would be empty for subjects that not have their default name overridden.

Fixes #9854.

### DIFF
--- a/components/collector/src/source_collectors/quality_time/missing_metrics.py
+++ b/components/collector/src/source_collectors/quality_time/missing_metrics.py
@@ -83,7 +83,7 @@ class QualityTimeMissingMetrics(QualityTimeCollector):
                 key=f"{report_uuid}:{subject_uuid}:{metric_type}",
                 report=report["title"],
                 report_url=report_url,
-                subject=subject.get("name", subject_type_name),
+                subject=subject.get("name") or subject_type_name,
                 subject_url=f"{report_url}#{subject_uuid}",
                 subject_uuid=f"{subject_uuid}",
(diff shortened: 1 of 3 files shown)


components/collector/src/source_collectors/quality_time/missing_metrics.py
```diff
@@ -83,7 +83,7 @@ def __subject_missing_metric_type_entities(
                 key=f"{report_uuid}:{subject_uuid}:{metric_type}",
                 report=report["title"],
                 report_url=report_url,
-                subject=subject.get("name", subject_type_name),
+                subject=subject.get("name") or subject_type_name,
```
wkoot, line 86: Is this not something we can address in the datamodel, or does the issue derive from only storing stringy data in the db?

fniessink: In theory this could fixed by making sure domain objects always have all attributes present (even if with an empty string or empty list as value). Downside is the extra storage needed. Impact could be considerable if all measurement entities have all their attributes (entity status, status rationale, status end date, etc.) stored at all times. Especially when we remove the limit of storing 100 measurement entities. Maybe Mongo deals with the duplication, but it would require some looking into.

PR #9975, opened 2024-10-09
Cut off long text fields.

### LINKED ISSUE #8281: Limit the length of comments when metrics are collapsed.
When a metrics are collapsed, and technical debt comments have long texts, the metrics become unusably wide (row height). This forces the user to scroll a lot.

A metric comment should display no more than 1000 characters when the metric is collapsed (please test this thoroughly, when resizing your viewport to a narrower width the limit should be max 200 chars maybe?).

image

fniessink:
In metric tables, prevent the comment column from getting too wide and cut off comments that are longer than 250 characters. In the measurement entity details, prevent the rationale column from getting too wide and cut off rationales that are longer than 250 characters.

Fixes #8281.

### DIFF
--- a/components/frontend/src/utils.js
+++ b/components/frontend/src/utils.js
@@ -337,6 +337,15 @@ export function pluralize(word, count) {
     return word + (count === 1 ? "" : "s")
 }
 
+export function limitTextLength(text, maxLength = 250) {
+    if (typeof text === "string" && text.length > maxLength) {
+        const ellipsis = "..."
+        const newLength = Math.max(0, maxLength - ellipsis.length)
+        return text.slice(0, newLength) + ellipsis
+    }
+    return text
+}
+
 export function niceNumber(number) {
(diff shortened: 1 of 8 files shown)


components/frontend/src/utils.js
```diff
@@ -337,6 +337,13 @@ export function pluralize(word, count) {
     return word + (count === 1 ? "" : "s")
 }
 
+export function limitTextLength(text, maxLength = 250) {
+    if (typeof text === "string" && maxLength > 3 && text.length > maxLength) {
```
wkoot: So `if maxLength < 4`, then the whole content is displayed anyway?

fniessink: Having a max length smaller than "..." doesn't make sense so I was trying to work around that. I guess I should just throw an exception if the max length is smaller than the length of the ellipsis.

fniessink: Fixed by just returning "..." if the maxLength is equal or smaller than the ellipsis length.

PR #10031, opened 2024-10-14
Upgrade MongoDB from v7.0 to v8.0.

### LINKED ISSUE #9960: Upgrade to MongoDB v8

fniessink:
Closes #9960.


components/database/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mongo:7.0.14
+FROM mongo:8.0.1
```
wkoot, line 1: So with the flag in `components/api_server/src/initialization/database.py` this just keeps working as-is.
How should we now proceed with upping the feature flag to 8? Shouldn't we already migrate the data in this change?
Or do we want to introduce the data format migration in the same change as where we up the feature flag?

wkoot (approved):

components/database/Dockerfile
fniessink: No, you never update the feature compatibility flag together with the Mongo DB version because after updating the feature compatibility to v8 you can't downgrade to v7 anymore (or at least, it's not guaranteed to work). We wait a couple of Quality-time releases to be sure there are no problems with Mongo DB v8 and only then we increase the feature compatibility flag.

PR #10032, opened 2024-10-14
Finish migration to uv.

fniessink:
- Remove pip-tools, pipx, and pip from the dev dependencies
- Remove the pip-tools settings from the pyproject.toml files
- Run uv pip compile with the --update flag as we did with pip compile
- Update indirect dependencies


components/frontend/package-lock.json
```diff
@@ -8133,7 +8133,9 @@
             "license": "MIT"
         },
         "node_modules/cookie": {
-            "version": "0.6.0",
+            "version": "0.7.1",
```
wkoot, line 8136: Are these additional changes, or were they needed for uv?

release/pyproject.toml
```diff
@@ -10,11 +10,7 @@ classifiers = [
 dependencies = [
     "gitpython==3.1.43",
 ]
-optional-dependencies.dev = [
-    "pip==24.2",
-    "pip-tools==7.4.1", # To add hashes to requirements
-    "pipx==1.7.1",
-]
+optional-dependencies.dev = [  ]
```
wkoot: Why not omit?

components/frontend/package-lock.json
fniessink: These were skipped before because I forgot to add `--upgrade` to `uv pip compile`.

release/pyproject.toml
fniessink: I had removed it and `pip compile` complained so I put it back in. But upon looking better (thanks) a better solution is to remove the `--extra dev` flag. Oops.

Issue #9796, opened 2024-09-18
Replace SUIR with MUI


fniessink:
Replace Semantic UI React with Material UI. 

POC: #9550.

Tasks:
- [x] Triangle subject overview title is too small.
- [x] Make border of dashboard cards grey, black on hover
- [x] Make blue darker (menu, footer, headers, buttons)
- [x] The center column and the logo in the footer are misaligned
- [x] Make sure that the headings in the settings panel don't line break
- [x] Don't indent the menu items in the settings panel
- [x] In the desired reaction times panel, make the help texts consistent (means versus colon).
- [x] In the issue tracker panel, make the Jira description wrap
- [x] Add a background color to the input fields to distinguish them more from the background
- [x] Change the anchor the the "Read the Docs" links to "Read the docs about the [item type]"
- [x] The warning icon when small needs a smaller triangle.
- [x] Use the same font in Victory charts as in the rest of the UI.
- [x] More padding between the source parameters (and other fields?)
- [x] Don't show selected items in "autocomplete" fields.
- [x] Change "Tags" to "Metric tags" in the metric configuration tab.
- [x] Use a lighter grey in the pie charts in dark mode.
- [x] Check metric column header alignment
- [x] Fix unit test warnings
- [x] Fix console warnings
- [x] Get rid of `STATUS_COLORS_RGB`.
- [x] Make subject type description wrap in the subject type picker in the subject title.
- [x] Check sorting of drop down menu's.
- [x] Make violation states description wrap in the measurement entity status dropdown.
- [x] Update screenshots.
- [x] Update changelog.

Out of scope:
- #10687
- #10688
- #10686
- #10690
- #10689

Won't do:

- Break long report/subject titles in dashboard cards: won't do, the pie charts will become very small if the report/subject title is long
- Make the delete button color red instead of orange: won't do because red is the color used for errors and orange for warnings; deleting something is not an error.

PR #10051, opened 2024-10-15
Replace SUIR components with SUI components.

fniessink:
Migrates:
- Avatar
- LogoutButton

Partially addresses #9796.


components/frontend/src/header_footer/Menubar.test.js
```diff
@@ -90,7 +90,8 @@ it("logs out", async () => {
     auth.logout = jest.fn().mockResolvedValue({ ok: true })
     const set_user = jest.fn()
     renderMenubar({ set_user: set_user, user: "jadoe" })
-    await fireEvent.click(screen.getByText(/Logout/))
+    await fireEvent.click(screen.getByRole("button", { name: "Logout" }))
+    await fireEvent.click(screen.getByRole("menuitem", { name: "Logout" }))
```
wkoot, line 94: Two clicks?

wkoot (approved):

components/frontend/src/header_footer/Menubar.test.js
fniessink: Yeah, the first one the button with the user avatar to open the menu and then on the menu item "Logout" to actually log out. I guess it's surprising because the user avatar button has "Logout" as accessible name. I'll change that to something like "User options".

PR #10099, opened 2024-10-18
Faulty label in metric summary cards with pie charts.

### LINKED ISSUE #10098: The metric cards with pie chart don't erase the center label when rerendering.
image

fniessink:
The metric summary cards with pie chart wouldn't erase the center label and tooltips when rerendering, causing the label to be visible multiple times.

Fixes #10098.


components/frontend/src/dashboard/StatusPieChart.js
```diff
@@ -12,10 +12,10 @@ nrMetricsLabel.PropTypes = {
     nrMetrics: number,
 }
 
-export function StatusPieChart({ animate, colors, label, tooltip, summary, style, maxY, width, height }) {
+export function StatusPieChart({ animate, colors, events, height, label, maxY, style, summary, tooltip, width }) {
     const nrMetrics = sum(summary)
     const outerRadius = 0.45 * Math.min(height, width)
-    const minInnerRadius = 0.3 * outerRadius
+    const minInnerRadius = 0.4 * outerRadius
```
wkoot: Where do these magic numbers come from?

components/frontend/src/dashboard/MetricSummaryCard.js
```diff
@@ -69,20 +69,36 @@ export function MetricSummaryCard({ header, onClick, selected, summary, maxY })
...
+                },
+            },
+        ],
         height: Math.max(bbHeight, 1), // Prevent "Failed prop type: Invalid prop range supplied to VictoryBar"
-        label: label,
+        label: (
+            <VictoryLabel
+                text={nrMetricsLabel(sum(summary[dates[0]]))}
+                textAnchor="middle"
+                x={bbWidth / 2}
```
wkoot, line 98: Is it possible to use `// 2`, so an int is always returned? Or is it not a problem for VictoryLabel?

docs/src/versioning.md
```diff
@@ -21,12 +21,13 @@ The table below contains the *Quality-time* releases since the last minor of the
 
 | Version    | Date         | Mongo  | FC     | Migrations | Downgrade       | Upgrade         | Manual changes |
 |------------|--------------|--------|--------|------------|-----------------|-----------------|----------------|
-| v5.17.0    | 2024-10-17   | **v8** | v7     |            | v5.14.0-v5.16.2 | n/a             | no             |
-| v5.16.2    | 2024-10-03   | v7     | v7     |            | v5.14.0-v5.16.1 | v5.17.0         | no             |
-| v5.16.1    | 2024-09-26   | v7     | v7     |            | v5.14.0-v5.16.0 | v5.16.2-v5.17.0 | no             |
-| v5.16.0    | 2024-09-19   | v7     | v7     | added      | v5.14.0-v5.15.0 | v5.16.1-v5.17.0 | no             |
-| v5.15.0    | 2024-07-30   | v7     | v7     |            | v5.14.0         | v5.16.0-v5.17.0 | no             |
-| v5.14.0    | 2024-07-05   | v7     | **v7** | added      | not possible    | v5.15.0-v5.17.0 | no             |
+| v5.17.1    | [unreleased] | v8     | v7     |            | v5.14.0-v5.17.0 | n/a             | no             |
```
wkoot, line 24: Due to the mongo upgrade, I don't think we should report Downgrade to anything lower than `v5.17.0`

components/frontend/src/dashboard/StatusPieChart.js
fniessink: Trial and visual inspection I'm afraid. There needs to be enough room in the center for the label (nr of metrics) and that depends on font size.

components/frontend/src/dashboard/MetricSummaryCard.js
fniessink: Javascript doesn't have a `//` operator. It starts a comment. And apparently floats are no problem for the VictoryLabel.

docs/src/versioning.md
fniessink: Disagree. In your proposal, we're basically saying that if Mongo v8 causes problems you shouldn't downgrade Quality-time to a version that has Mongo v7. But the whole idea of the feature compatibility flag is that you can downgrade to the previous version of Mongo as long as the feature compatibility flag stays the same.

PR #10145, opened 2024-10-23
Donut charts would be misaligned after resizing the window.

### LINKED ISSUE #9863: Misalignment of text in donut charts
normal view (browser chrome, latest version of qt)
image

full screen view 
image

fniessink:
Fixes #9863.

### DIFF
--- a/components/frontend/src/dashboard/MetricSummaryCard.js
+++ b/components/frontend/src/dashboard/MetricSummaryCard.js
@@ -62,7 +62,7 @@ export function MetricSummaryCard({ header, onClick, selected, summary, maxY }) 
             flyoutPadding={5}
             flyoutStyle={{ fill: flyoutBgColor }}
             pointerWidth={20}
-            renderInPortal={true}
+            renderInPortal={false}
             style={{ fontFamily: "Arial", fontSize: 12, fill: labelColor }}
         />
     )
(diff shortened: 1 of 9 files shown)


wkoot (approved):

components/frontend/src/dashboard/MetricSummaryCard.js
```diff
@@ -72,27 +72,10 @@ export function MetricSummaryCard({ header, onClick, selected, summary, maxY })
...
-            {
-                childName: "all",
-                target: "data",
-                eventHandlers: {
-                    onClick: () => {
-                        return [
-                            {
-                                target: "labels",
-                                mutation: () => {
-                                    return { active: false } // Clear tooltip before filtering
```
wkoot, line 85: No longer needed?

components/frontend/src/dashboard/StatusPieChart.js
```diff
@@ -21,7 +21,7 @@ export function StatusPieChart({ animate, colors, events, height, label, maxY, s
     const data = STATUSES.map((status) => {
         const y = summary[STATUS_COLORS[status]]
         const yPercentage = Math.round((y / nrMetrics) * 100)
-        return { y: y, label: `${STATUS_NAME[status]}: ${nrMetricsLabel(y)} (${yPercentage}%)` }
+        return { y: y, label: `${STATUS_NAME[status]}\n${nrMetricsLabel(y)} (${yPercentage}%)` }
```
wkoot, line 24: This seems very similar to the change in `StatusBarChart.js`, is the pattern used elsewhere?

components/frontend/src/dashboard/MetricSummaryCard.js
fniessink: Yes, this is made obsolete by not rendering the tooltip in the portal.

components/frontend/src/dashboard/StatusPieChart.js
fniessink: I don't think so.

PR #10177, opened 2024-10-25
 Correctly sort subjects in the "Add subject" dropdown menu.

### LINKED ISSUE #10176: Subjects in the "Add subject" dropdown menu are not sorted correctly.
image

fniessink:
Fixes #10176.

### DIFF
--- a/components/frontend/src/widgets/Button.js
+++ b/components/frontend/src/widgets/Button.js
@@ -102,7 +102,7 @@ ActionButton.propTypes = {
     position: string,
 }
 
-export function AddDropdownButton({ itemSubtypes, itemType, onClick, allItemSubtypes, usedItemSubtypeKeys }) {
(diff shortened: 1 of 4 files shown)


components/frontend/src/widgets/Button.js
```diff
@@ -115,7 +115,9 @@ export function AddDropdownButton({ itemSubtypes, itemType, onClick, allItemSubt
         items = items.filter((item) => !usedItemSubtypeKeys.includes(item.key))
     }
     const options = items.filter((itemSubtype) => itemSubtype.text.toLowerCase().includes(query.toLowerCase()))
-    options.sort((a, b) => a.text.localeCompare(b.text))
+    if (sort !== false) {
```
wkoot, line 119: Why the double negate? Prop type says it's bool so there shouldn't be more options than true or false right?

fniessink: Prop types are optional by default, and evaluating the sort parameter this way makes its default value true.

PR #10179, opened 2024-10-25
When calculating how much time is left until the technical debt end d…

### LINKED ISSUE #10063: Technical debt end date does not seem to be applied correctly
**Describe the bug**
Technical debt end date does not seem to be applied correctly. This indicates by the metric that this has expired, but the metric is still gray (technical debt has not yet expired). 

Documentation says: accept technical debt until this date. It looks like it is a up to and including date?

**To Reproduce**
1. Accept Technical Debt 
2. Configure technical debt end date of today (e.g. 15-10-2024)
3. Metric "says" Technical debt end date (x hours ago) see screenshot
4. Metric is still gray

**Expected behaviour**
Metric is coloured based on the evaluated targets (any colour except gray)

**Screenshots**
image

 - Quality-time v5.16.2

fniessink:
…ate, include the technical debt end date itself in the calculation.

Fixes #10063.

### DIFF
--- a/components/frontend/src/metric/MetricDebtParameters.js
+++ b/components/frontend/src/metric/MetricDebtParameters.js
@@ -68,17 +68,17 @@ function TechnicalDebtEndDate({ metric, metric_uuid, reload }) {
             </p>
         </>
     )
+    let debtEndDateTime = null
+    if (metric.debt_end_date) {
+        debtEndDateTime = new Date(metric.debt_end_date)
+        debtEndDateTime.setHours(23, 59, 59)
+    }
     return (
         <DateInput
             ariaLabelledBy={labelId}
             requiredPermissions={[EDIT_REPORT_PERMISSION]}
             label={
-                <LabelWithDate
-                    date={metric.debt_end_date}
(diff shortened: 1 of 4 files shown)


components/frontend/src/metric/MetricDebtParameters.js
```diff
@@ -68,17 +68,17 @@ function TechnicalDebtEndDate({ metric, metric_uuid, reload }) {
             </p>
         </>
     )
+    let debtEndDateTime = null
+    if (metric.debt_end_date) {
+        debtEndDateTime = new Date(metric.debt_end_date)
```
wkoot, line 73: Why don't we do this in the deserialization step, but only here? Or even add a `metric.debt_end_date_time` (e.g. when deserializing) so other methods can use whichever type is more convenient for them?

fniessink: There is no deserialisation step that converts date and date time strings in the JSON received from the API-server into `Date`s. The only thing we do is call `response.json()` but that doesn't know about dates and date times.

PR #10231, opened 2024-11-04
Don't trim long comments as they can contain HTML, but instead overfl…

### LINKED ISSUE #10183: Comments not visible or partially visible
**Describe the bug**
Comments not visible or partially visible

**To Reproduce 1-2**
1. Add comment with a URL with many characters (250+)
2. Comment section is empty

**To Reproduce 2-2**
1. Add comment with multiple short URL's 
2. Comment section only shows a few URL's

**Expected behaviour**
See comment and URL

 - Browser: Chrome
 - Quality-time version: v5.17.1

fniessink:
…ow the table cell.

Fixes #10183.

### DIFF
--- a/components/frontend/src/widgets/DivWithHTML.js
+++ b/components/frontend/src/widgets/DivWithHTML.js
@@ -1,0 +1,32 @@
+import { createRef, useLayoutEffect, useState } from "react"
+
+import { childrenPropType } from "../sharedPropTypes"
+
+export function DivWithHTML({ children }) {
+    const minHeight = 15
+    const maxHeight = 60
+    const ref = createRef()
+    // Keep track of whether the div is overflown, so the resize handle can be hidden when not needed
+    const [overflown, setOverflown] = useState(false)
+    const [scrollHeight, setScrollHeight] = useState(minHeight)
+    useLayoutEffect(() => {
+        setOverflown(ref.current.clientHeight < ref.current.scrollHeight)
+        setScrollHeight(ref.current.scrollHeight)
+    }, [ref])
+    return (
+        <div
+            ref={ref}
+            style={{
+                wordBreak: "break-word",
+                overflow: "auto",
+                minHeight: minHeight,
+                height: Math.min(maxHeight, scrollHeight),
+                resize: overflown ? "vertical" : "",
+            }}
+            dangerouslySetInnerHTML={{ __html: children }}
+        />
+    )
+}
+DivWithHTML.propTypes = {
+    children: childrenPropType,
+}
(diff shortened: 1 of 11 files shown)


components/frontend/src/widgets/DivWithHTML.js
```diff
@@ -0,0 +1,32 @@
...
+export function DivWithHTML({ children }) {
+    const minHeight = 15
+    const maxHeight = 60
+    const ref = createRef()
+    // Keep track of whether the div is overflown, so the resize handle can be hidden when not needed
+    const [overflown, setOverflown] = useState(false)
+    const [scrollHeight, setScrollHeight] = useState(minHeight)
+    useLayoutEffect(() => {
+        setOverflown(ref.current.clientHeight < ref.current.scrollHeight)
+        setScrollHeight(ref.current.scrollHeight)
```
wkoot, line 14: I don't understand how this works to adjust the `scrollHeight`, but seems like some nice js magic :)

components/frontend/src/widgets/DivWithHTML.js
```diff
@@ -0,0 +1,32 @@
...
+    }, [ref])
+    return (
+        <div
+            ref={ref}
+            style={{
+                wordBreak: "break-word",
+                overflow: "auto",
+                minHeight: minHeight,
+                height: Math.min(maxHeight, scrollHeight),
+                resize: overflown ? "vertical" : "",
```
wkoot, line 24: So this means that the div can only grow vertically, i.e. in height, and will therefore not break table columns that it's in?

components/frontend/src/widgets/DivWithHTML.js
```diff
@@ -0,0 +1,32 @@
...
+            }}
+            dangerouslySetInnerHTML={{ __html: children }}
```
wkoot, line 26: This inner html cannot break the parent bounding box and resizing rules (e.g. with something like display block)?

wkoot (approved):

components/frontend/src/widgets/DivWithHTML.js
fniessink: The `ref.current.scrollHeight` isn't available until after the size of the children prop has been calculated. So we can't pass it directly to the div style. The `useLayoutEffect` hook runs after the children prop has been rendered and the scrollHeight is known.

components/frontend/src/widgets/DivWithHTML.js
fniessink: Yes!

components/frontend/src/widgets/DivWithHTML.js
fniessink: I don't think so, but if users want to shoot themselves in the foot that way they can try :-)

Issue #8944, opened 2024-06-11
Replace Create React App (CRA) with another build tool


fniessink:
Tasks:
-  Run `npm run eject` and see how hard it is to get a working build (using webpack?)
- Check which dependencies can be removed (e.g. tailwindcss?).

Possible alternatives:
- Vite 
  - Docs: https://vitejs.dev 
  - Migration: https://www.freecodecamp.org/news/how-to-migrate-from-create-react-app-to-vite/
- [ParcelJS](https://parceljs.org/)  
- Also see [Google](https://www.google.nl/search?q=create+react+app+migration%C2%A0&safe=active&client=safari&sca_esv=7220d14f8ee464f5&biw=1658&bih=723&tbs=qdr%3Ay&ei=LpUsZ6DEKJKG9u8P2r3wqA0&ved=0ahUKEwigk926gMqJAxUSg_0HHdoeHNUQ4dUDCA8&oq=create+react+app+migration%C2%A0&gs_lp=Egxnd3Mtd2l6LXNlcnAiHGNyZWF0ZSByZWFjdCBhcHAgbWlncmF0aW9uwqBIAFAAWABwAHgBkAEAmAEAoAEAqgEAuAEMyAEAmAIAoAIAmAMAkgcAoAcA&sclient=gws-wiz-serp)

PR #10239, opened 2024-11-05
Use multi-stage build for frontend container

fniessink:
```
REPOSITORY                  TAG      IMAGE ID       CREATED          SIZE
ictu/quality-time_frontend  v5.18.0  77aed96f843b   10 minutes ago   186MB
...
ictu/quality-time_frontend  v5.17.1  b136d1ea1ceb   8 days ago       803MB
```

### DIFF
--- a/components/frontend/Dockerfile
+++ b/components/frontend/Dockerfile
@@ -1,18 +1,22 @@
+FROM node:23.1.0-alpine3.20 AS compile-image
+
+WORKDIR /home/frontend
+COPY public /home/frontend/public
+COPY package*.json /home/frontend
+COPY src /home/frontend/src
+COPY .env /home/frontend
+RUN npm install --ignore-scripts && \
+    npm run --ignore-scripts build
+
 FROM node:23.1.0-alpine3.20
 
 LABEL maintainer="Quality-time team <quality-time@ictu.nl>"
 LABEL description="Quality-time frontend"
 
 WORKDIR /home/frontend
-COPY package*.json /home/frontend
-COPY public /home/frontend/public
-COPY src /home/frontend/src
 COPY healthcheck.js /home/frontend
-COPY .env /home/frontend
-RUN npm install --ignore-scripts -g npm@10.9.0 && \
-    npm install --ignore-scripts && \
-    npm install --ignore-scripts -g serve@14.2.4 && \
-    npm run --ignore-scripts build && \
+COPY --from=compile-image /home/frontend/build /home/frontend/build
+RUN npm install --ignore-scripts -g serve@14.2.4 && \
     adduser frontend --disabled-password
 USER frontend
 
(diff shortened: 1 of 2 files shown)


components/frontend/Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY src /home/frontend/src
 COPY healthcheck.js /home/frontend
 COPY .env /home/frontend
 RUN npm install --ignore-scripts -g npm@10.9.0 && \
-    npm install --ignore-scripts && \
+    npm install --ignore-scripts --omit=dev && \
```
wkoot: That's unexpected! I'm surprised that these are installed by default

wkoot (approved):

components/frontend/Dockerfile
fniessink: Yeah, exactly. I had to move two dependencies from dev to production to make it work though, including react-scripts which isn't maintained anymore (#8944). Would be nice to not have react-scripts end up in the image, so I want to look into a staged build for the frontend container. That's why this PR is still draft.

components/frontend/Dockerfile
```diff
@@ -1,18 +1,22 @@
...
-COPY package*.json /home/frontend
-COPY public /home/frontend/public
-COPY src /home/frontend/src
 COPY healthcheck.js /home/frontend
-COPY .env /home/frontend
-RUN npm install --ignore-scripts -g npm@10.9.0 && \
-    npm install --ignore-scripts && \
-    npm install --ignore-scripts -g serve@14.2.4 && \
-    npm run --ignore-scripts build && \
+COPY --from=compile-image /home/frontend/build /home/frontend/build
```
wkoot, line 18: How are the dev dependencies now omitted from the build?

components/frontend/Dockerfile
```diff
@@ -1,18 +1,22 @@
...
 
 WORKDIR /home/frontend
-COPY package*.json /home/frontend
-COPY public /home/frontend/public
-COPY src /home/frontend/src
 COPY healthcheck.js /home/frontend
-COPY .env /home/frontend
-RUN npm install --ignore-scripts -g npm@10.9.0 && \
-    npm install --ignore-scripts && \
-    npm install --ignore-scripts -g serve@14.2.4 && \
```
wkoot, line 14: These two global installs are no longer needed? Nice!

components/frontend/Dockerfile
fniessink: Because they are not copied from the compile-image to the final image. Only the build folder with the app is copied.

components/frontend/Dockerfile
fniessink: I removed the pinned install of npm because the base image has npm already. Serve is still needed, it's installed in the second stage.

PR #10268, opened 2024-11-12
Migrate frontend footer components to MUI.

fniessink:
Partially implements #9796.

Old footer: 
image

New footer:
image

### DIFF
--- a/components/frontend/src/header_footer/Footer.css
+++ b/components/frontend/src/header_footer/Footer.css
@@ -1,11 +1,11 @@
-@media print {
-    div.footer {
-        display: none;
-    }
+.MuiDivider-root {
+    padding: 20px;
 }
 
-div.footer {
-    background-color: var(--inverted-menu-background-color) !important;
-    margin: 5em 0em 0em !important;
-    padding: 5em 0em 3em !important;
+.MuiDivider-root:before {
+    border-top: thin solid rgba(255, 255, 255, 0.3) !important;
+}
+
+.MuiDivider-root:after {
+    border-top: thin solid rgba(255, 255, 255, 0.3) !important;
 }
--- a/components/frontend/src/header_footer/Footer.js
+++ b/components/frontend/src/header_footer/Footer.js
@@ -1,122 +1,112 @@
 import "./Footer.css"
 
-import { number, object, oneOfType, string } from "prop-types"
-import { Container, Divider, Grid, Header, Icon, Image, List, Segment } from "semantic-ui-react"
+import BugReportIcon from "@mui/icons-material/BugReport"
+import CopyrightIcon from "@mui/icons-material/Copyright"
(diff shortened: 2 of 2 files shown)


components/frontend/src/header_footer/Footer.css
```diff
@@ -1,11 +1,11 @@
...
+.MuiDivider-root {
+    padding: 20px;
 }
 
-div.footer {
-    background-color: var(--inverted-menu-background-color) !important;
-    margin: 5em 0em 0em !important;
-    padding: 5em 0em 3em !important;
+.MuiDivider-root:before {
+    border-top: thin solid rgba(255, 255, 255, 0.3) !important;
```
wkoot, line 6: No troubles with dark mode?

components/frontend/src/header_footer/Footer.css
```diff
@@ -1,11 +1,11 @@
-@media print {
-    div.footer {
-        display: none;
-    }
+.MuiDivider-root {
+    padding: 20px;
```
wkoot, line 2: Looks fine on the sample images, but don't you want a difference between horizontal and vertical padding for responsive views?

components/frontend/src/header_footer/Footer.js
```diff
@@ -149,35 +139,38 @@ function QuoteColumn() {
...
-                        <Grid.Column width={1} />
-                        <AboutAppColumn />
-                        {report ? <AboutReportColumn lastUpdate={lastUpdate} report={report} /> : <QuoteColumn />}
-                        <Grid.Column width={1} />
-                        <SupportColumn />
-                    </Grid.Row>
-                </Grid>
-                <Divider inverted section />
-                <Image centered size="mini" src="./favicon.ico" />
+        <Container maxWidth="100%" sx={{ bgcolor: "black", displayPrint: "none", marginTop: "20px", padding: "60px" }}>
```
wkoot, line 156: hehe, additional padding and margin

wkoot (approved):

components/frontend/src/header_footer/Footer.css
fniessink: No, the footer background is always black.

components/frontend/src/header_footer/Footer.js
fniessink: Yeah, top margin to create some distance between the subject tables and the footer, and padding so the footer content keeps its distance from the window border.

components/frontend/src/header_footer/Footer.css
fniessink: It seems to work fine for smaller widths. Of course, really small widths don't work well because we don't have breakpoints. Moreover, if you make the window narrower, the menu bar gets cut off before the footer doesn't have enough space.

PR #10324, opened 2024-11-19
Filter by Dependency-Track component name.

### LINKED ISSUE #9577: Allow for (regex) filtering on components with Dependency-Track as a source for Security Warnings
Add parameters:
- [x] Components to include (name or regexp)
- [x] Components to exclude (name or regexp)

fniessink:
Allow for filtering by component name when measuring dependencies and security warnings with Dependency-Track as source.

Closes #9577.

### DIFF
--- a/components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
+++ b/components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
@@ -101,6 +101,30 @@ class DependencyTrackSecurityWarningsTest(DependencyTrackTestCase):
         response = await self.collect(get_request_json_side_effect=[self.projects(), self.findings()])
         self.assert_measurement(response, value="1", entities=self.entities())
 
+    async def test_include_by_component_name(self):
+        """Test filtering by component name."""
+        self.set_source_parameter("components_to_include", ["other component"])
+        response = await self.collect(get_request_json_side_effect=[self.projects(), self.findings()])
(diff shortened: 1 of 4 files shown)


components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
```diff
@@ -101,6 +101,30 @@ async def test_filter_by_project_name_and_version(self):
...
+        self.set_source_parameter("components_to_ignore", ["component name"])
+        response = await self.collect(get_request_json_side_effect=[self.projects(), self.findings()])
+        self.assert_measurement(response, value="0", entities=[])
+
+    async def test_exclude_by_component_name_regular_expression(self):
+        """Test filtering by component name regular expression."""
+        self.set_source_parameter("components_to_ignore", ["other.*"])
+        response = await self.collect(get_request_json_side_effect=[self.projects(), self.findings()])
+        self.assert_measurement(response, value="1", entities=self.entities())
+
```
wkoot, line 127: The code allows for a combination of including and excluding, i.e. include `"python.*"` and ignore `".*2.7.18"` should probably select `["python-alpine", "python-2.7.18-devel"]`, but exclude `["alpine-python", "python-2.7.18"]`. Although this might not be true if we do partial string matching instead of full string

fniessink: We do a regular expression match (not a search), so it works as you describe.

PR #10404, opened 2024-11-28
When copying an item, don't add "(copy)".

### LINKED ISSUE #9859: When copying an item, do not add (copy) to the name
When copying a report, subject, source or metric, do not add "(copy)" to the name.

fniessink:
When copying a report, subject, metric, or source, don't add "(copy)" to the name.

Closes #9859.

### DIFF
--- a/components/api_server/src/model/actions.py
+++ b/components/api_server/src/model/actions.py
@@ -7,57 +7,51 @@ from shared.model.subject import Subject
 from shared.model.metric import Metric
 from shared.model.source import Source
 from shared.utils.type import ItemId
-from shared_data_model import DATA_MODEL
 
 from model.report import Report
 from utils.functions import uuid
 from utils.type import Position
 
 
-def copy_item(item, **kwargs):
+def copy_item[Item: Metric | Report | Source | Subject](item: Item, **kwargs) -> Item:
     """Return a copy of the item."""
-    item_copy = item.copy()
+    item_copy = cast(Item, item.copy())
     for key, value in kwargs.items():
         item_copy[key] = value
     return item_copy
 
 
-def copy_source(source, change_name: bool = True):
+def copy_source(source: Source) -> Source:
     """Return a copy of the source."""
-    kwargs = {}
-    if change_name:
-        kwargs["name"] = f"{source.get('name') or DATA_MODEL.sources[source['type']].name} (copy)"
-    return copy_item(source, **kwargs)
+    return copy_item(source)
 
 
-def copy_metric(metric, change_name: bool = True):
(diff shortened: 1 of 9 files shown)


components/api_server/src/model/actions.py
```diff
@@ -7,73 +7,66 @@
...
 from shared.utils.type import ItemId
-from shared_data_model import DATA_MODEL
 
 from model.report import Report
 from utils.functions import uuid
 from utils.type import Position
 
 
-def copy_item(item, **kwargs):
+def copy_item[Item: Metric | Report | Source | Subject](item: Item, **kwargs) -> Item:
```
wkoot, line 16: Interesting, I hadn't seen this dynamic type attribution yet!

components/api_server/src/model/actions.py
```diff
@@ -7,73 +7,66 @@
...
-    return copy_item(source, **kwargs)
+    return copy_item(source)
 
 
-def copy_metric(metric, change_name: bool = True):
+def copy_metric(metric: Metric) -> Metric:
     """Return a copy of the metric and its sources."""
     kwargs: dict[str, Any] = {
-        "sources": {uuid(): copy_source(source, change_name=False) for source in metric["sources"].values()},
+        "sources": {uuid(): copy_source(source) for source in metric["sources"].values()},
```
wkoot, line 32: Why does the `Metric.copy()` not implement this instead?

components/api_server/src/model/actions.py
```diff
@@ -7,73 +7,66 @@
...
+
 def move_item(
     container: Report | Subject | Metric,
     item_to_move: Subject | Metric | Source,
     new_position: Position,
 ) -> tuple[int, int]:
     """Change the item position."""
-    items_dict_type = MutableMapping[ItemId, Metric | Source | Subject]
-    items_dict: items_dict_type
+    items_dict: ItemsDictType
```
wkoot, line 63: Is this magic not something we can implement in a `move_child` implementation of the various classes? It seems like we need to do a lot of housekeeping with these type checks

components/api_server/src/model/actions.py
fniessink: Because the Metric class is (still) a dict subclass, so `Metric.copy()` is `dict.copy()`. I think we should first convert `Metric` (and `Subject`, `Source`, etc.) to a proper class before moving these methods there.

components/api_server/src/model/actions.py
fniessink: I think these changes should be done as part of refactoring `Metric`, `Source`, etc. to proper domain classes. Also see previous comment.

PR #10434, opened 2024-12-04
Add a new metric 'test suites'.

### LINKED ISSUE #10078: Count test suites
# User story

As a tester I want to count test scenario's so that I can report the number of test scenario's without needing to count them manually.

Because Robot Framework itself calls test scenario's test suites in its reports, and since other sources such as Junit and TestNG also use the term suite, the metric is called 'test suites' instead of 'test scenario's'.

## Acceptance criteria

- [x] Add a metric 'test suites'
- [x] Allow Robot Framework XML as source for the test suites metric
- [x] Allow Junit XML as source for the test suites metric
- [x] Allow TestNG as source for the test suites metric

### Using the test cases metric to count test scenario's

The test cases metric is meant to match two sources (it matches logical test cases from Jira against physical test cases in test reports generated by e.g. Robot Framework) and report "untested" test cases. In this case there is just one source; Robot Framework and there should never be untested test cases.

### Using the tests metric

In theory, a parameter could be added for the user to configure whether they want to count tests or test suites. However, sources also support filtering by test result. That logic would need to be applied to test suites if the user wants to count test suites. Also, which measurement entity attributes (suite result, test result, test name) to show would depend on whether test suites or tests are to be counted.

## Out of scope

These sources could potentially also support the test suites metric, but are out of scope for now:
- Jenkins test report
- Robot Framework Jenkins plugin

These sources cannot support the test suites metric:
- Azure DevOps Server: does not expose suite information in its TestRun response object, see: https://learn.microsoft.com/en-us/rest/api/azure/devops/test/runs/list?view=azure-devops-rest-7.1&tabs=HTTP#testrun
- Visual Studio TRX: does not contain test suites in its report file
- Gatling: does not have the concept of test suites
- JMeter CSV: does not have the concept of test suites
- JMeter JSON: does not have the concept of test suites
- Performancetest-runner: does not have the concept of test suites
- SonarQube: does not expose test suites

# Tasks

### Shared code

- [x] Extend the example reports
- [x] Add metric `test suites` to the data model
- [x] Configure Robot Framework to support the test suites metric
- [x] Configure Junit to support the test suites metric
- [x] Configure TesTNG to support the test suites metric
- [x] Configure Manual number to support the test suites metric

### Collector

- [x] Add a Robot Framework test suites collector
  - [x] Also include the documentation in the test suite entities
- [x] Add a Junit test suites collector
- [x] Add a TestNG test suites collector

## Documentation

- [x] Update changelog

fniessink:
Add a new metric 'test suites' that can be used to count test suites (or test scenario's in Robot Framework parlance). Test suites can be filtered by test result. Supporting sources are Robot Framework, JUnit, and TestNG.

Closes #10078.


docs/src/changelog.md
```diff
@@ -14,6 +14,10 @@ If your currently installed *Quality-time* version is not the latest version, pl
 
 ## [Unreleased]
 
+### Added
+
+- Add a new metric 'test suites' that can be used to count test suites (or test scenario's in Robot Framework parlance). Test suites can be filtered by test result. Supporting sources are Robot Framework, JUnit, and TestNG. Closes [#10078](https://github.com/ICTU/quality-time/issues/10078).
```
wkoot:
```suggestion
- Add a new metric 'test suites' that can be used to count test suites (or test scenarios in Robot Framework parlance). Test suites can be filtered by test result. Supporting sources are Robot Framework, JUnit, and TestNG. Closes [#10078](https://github.com/ICTU/quality-time/issues/10078).
```

wkoot (approved):

PR #10462, opened 2024-12-06
When measuring test cases with Visual Studio TRX as source, search al…

### LINKED ISSUE #10460: When matching test cases to tests with Visual Studio TRX as source, search all test category items for test case ids
When matching test cases to tests with Visual Studio TRX as source, search all test category items for test case ids, instead cutting the search short after the first match:

```python
    @staticmethod
    def __entity(test: Element, result: str, namespaces: Namespaces) -> Entity:
        """Transform a test case into a test entity."""
        name = test.attrib["name"]
        for category in test.findall(".//ns:TestCategoryItem", namespaces):
            if match := re.search(TestCases.TEST_CASE_KEY_RE, category.attrib["TestCategory"]):
                name += f" ({match[0]})"
                break # <= remove the break
        key = test.attrib["id"]
        return Entity(key=key, name=name, test_result=result)
```

fniessink:
…l test category items for test case ids, instead cutting the search short after the first match.

Fixes #10460.

### DIFF
--- a/components/collector/src/metric_collectors/test_cases.py
+++ b/components/collector/src/metric_collectors/test_cases.py
@@ -59,8 +59,9 @@ class TestCases(MetricCollector):
         "timeout": "errored",
         "warning": "errored",
     }
-    # Regular expression to identify test case ids in entity attributes:
-    TEST_CASE_KEY_RE = re.compile(r"\w+\-\d+")
+    # Regular expression to identify test case ids in entity attributes. Matches identifiers of the form BAR-123,
+    # while ensuring they are not part of longer identifiers such as FOO-BAR-123 or BAR-123-456:
+    TEST_CASE_KEY_RE = re.compile(r"(?<![-\w])\b[A-Za-z]+\-\d+\b(?![-\w])")
     ENTITY_ATTRIBUTES_TO_SEARCH = ("name", "test_name", "description")
     # The supported source types for test cases and test reports:
     TEST_CASE_SOURCE_TYPES: ClassVar[list[str]] = ["jira"]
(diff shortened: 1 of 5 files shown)


components/collector/src/metric_collectors/test_cases.py
```diff
@@ -59,8 +59,9 @@ class TestCases(MetricCollector):
         "timeout": "errored",
         "warning": "errored",
     }
-    # Regular expression to identify test case ids in entity attributes:
-    TEST_CASE_KEY_RE = re.compile(r"\w+\-\d+")
+    # Regular expression to identify test case ids in entity attributes. Matches identifiers of the form BAR-123,
+    # while ensuring they are not part of longer identifiers such as FOO-BAR-123 or BAR-123-456:
+    TEST_CASE_KEY_RE = re.compile(r"(?<![-\w])\b[A-Za-z]+\-\d+\b(?![-\w])")
```
wkoot, line 64: Shouldn't this be a metric source config parameter?

components/collector/src/source_collectors/visual_studio_trx/tests.py
```diff
@@ -39,9 +39,10 @@ def _include_entity(self, entity: Entity) -> bool:
...
         name = test.attrib["name"]
-        for category in test.findall(".//ns:TestCategoryItem", namespaces):
-            if match := re.search(TestCases.TEST_CASE_KEY_RE, category.attrib["TestCategory"]):
-                name += f" ({match[0]})"
-                break
+        category_items = test.findall(".//ns:TestCategoryItem", namespaces)
+        categories = [item.attrib["TestCategory"] for item in category_items]
+        matches = [match[0] for category in categories if (match := re.search(TestCases.TEST_CASE_KEY_RE, category))]
+        if matches:
+            name += f" ({', '.join(sorted(matches))})"
```
wkoot, line 46: Should we deduplicate as well as sorting?

components/collector/src/metric_collectors/test_cases.py
fniessink: Hmm, maybe. But for now we only support Jira ids, so we don't need different regular expressions per test case source yet.

components/collector/src/source_collectors/visual_studio_trx/tests.py
fniessink: If there are duplicates here it means one test has multiple references to the same Jira id. That would be weird, so I'd rather have a chance to detect it.

PR #10570, opened 2024-12-23
Add link to docs for missing metrics metric.

### LINKED ISSUE #10528: Add link to docs for each metric type in missing metrics
 Ex. I'm always frustrated when missing metrics reports missing metrics but I forgot why this metric is important or what it actually is.

a link in the metric type column to the docs page of each metric

image

fniessink:
Closes #10528.


components/shared_code/src/shared_data_model/meta/base.py
```diff
@@ -39,3 +42,15 @@ def check_description(self) -> Self:
         """Check the description."""
         self.check_punctuation("description", self.description)
         return self
+
+
+class DocumentedModel(DescribedModel):
+    """Extend the described model with a reference documentation URL."""
+
+    reference_documentation_url: str | None = None  # Set auomatically by set_reference_documentation_url() below
```
wkoot:
```suggestion
    reference_documentation_url: str | None = None  # Set automatically by set_reference_documentation_url() below
```

components/shared_code/src/shared_data_model/meta/base.py
```diff
@@ -39,3 +42,15 @@ def check_description(self) -> Self:
...
+
+    @model_validator(mode="after")
+    def set_reference_documentation_url(self) -> Self:
+        """Set the reference documentation URL based on the metric name."""
+        self.reference_documentation_url = f"{REFERENCE_DOCUMENTATION_URL}{slugify(self.name)}"
+        return self
```
wkoot, line 56: Why does this return an instance of `Self` instead of `None`?
The self object is not a deepcopy so the return might be confusing, as the current self is mutated in the process

fniessink: That's how Pydantic's "after" model validators work, see https://docs.pydantic.dev/latest/concepts/validators/#model-after-validator.

PR #10649, opened 2025-01-13
When measuring the number of job runs within a time period with Jenki…

### LINKED ISSUE #10610: Job runs within time period is flaky
**Describe the bug**
The metric: Job runs within time period
using source: Jenkins
sometimes gives a parse error for an unknown reason, while other Jenkins-based metric still have a working connection.
 

**To Reproduce**
1. Create above-mentioned metric with respective source.
2. Wait for the error to appear (inform your wife and family that you could be waiting for days)
3. See error

**Expected behaviour**

**Screenshots**

The parse error 
image

The flakiness visible
image

fniessink:
…ns as source, don't throw an exception when a Jenkins pipeline build has no result yet.

Fixes #10610.

### DIFF
--- a/components/collector/src/source_collectors/jenkins/job_runs_within_time_period.py
+++ b/components/collector/src/source_collectors/jenkins/job_runs_within_time_period.py
@@ -36,4 +36,4 @@ class JenkinsJobRunsWithinTimePeriod(JenkinsJobs):
         result_types = [result_type.lower() for result_type in self._parameter("result_type")]
         lookback_days = int(cast(str, self._parameter("lookback_days")))
         build_datetime = datetime_from_timestamp(int(build["timestamp"]))
-        return days_ago(build_datetime) <= lookback_days and build["result"].lower() in result_types
+        return days_ago(build_datetime) <= lookback_days and build.get("result", "").lower() in result_types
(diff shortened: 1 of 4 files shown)


components/collector/src/source_collectors/jenkins/job_runs_within_time_period.py
```diff
@@ -36,4 +36,4 @@ def _include_build(self, build: Build) -> bool:
         result_types = [result_type.lower() for result_type in self._parameter("result_type")]
         lookback_days = int(cast(str, self._parameter("lookback_days")))
         build_datetime = datetime_from_timestamp(int(build["timestamp"]))
-        return days_ago(build_datetime) <= lookback_days and build["result"].lower() in result_types
+        return days_ago(build_datetime) <= lookback_days and build.get("result", "").lower() in result_types
```
wkoot, line 39: Is this the only place where we get `build["result"]`? I'd expect it to be in multiple jenkins collectors

wkoot (approved):

components/collector/src/source_collectors/jenkins/job_runs_within_time_period.py
fniessink: We already did `build.get("result")` in other locations. This was the only place where it was `build["result"]`.

PR #10651, opened 2025-01-13
Allow for measuring the source up-to-dateness of Harbor JSON reports.

### LINKED ISSUE #10609: Add support for source up-to-dateness for Harbor JSON
Add support for source up-to-dateness for Harbor JSON to measure the number of days since the source was last updated.

Get date from "generated_at":
https://github.com/ICTU/quality-time/blob/master/components/testdata/reports/harbor/harbor.json

fniessink:
Closes #10609.


components/collector/src/source_collectors/harbor_json/base.py
```diff
@@ -0,0 +1,30 @@
+"""Base classes for Harbor JSON collectors."""
+
+from typing import Final, TypedDict
+
+REPORT_MIME_TYPE: Final = "application/vnd.security.vulnerability.report; version=1.1"
+
+
+class Vulnerability(TypedDict):
```
wkoot: Perhaps a good moment to rename this to `HarborVulnerability` or `HarborJSONVulnerability` ?

components/collector/tests/source_collectors/harbor_json/test_source_up_to_dateness.py
```diff
@@ -0,0 +1,17 @@
...
+
+class HarborJSONSourceUpToDatenessTest(HarborJSONCollectorTestCase):
+    """Unit tests for the source up-to-dateness metric."""
+
+    METRIC_TYPE = "source_up_to_dateness"
+
+    async def test_souce_up_to_dateness(self):
+        """Test the source up-to-dateness."""
+        response = await self.collect(get_request_json_return_value=self.VULNERABILITIES_JSON)
+        expected_value = str(days_ago(parse_datetime("2023-08-26T16:32:21.923910328Z")))
```
wkoot, line 16: Shouldn't we declare the `datetime.datetime` instead of re-using `parse_datetime`, or can we assume that `parse_datetime` is sufficiently tested elsewhere?

fniessink: It's basically `dateutil.parser.parse()` with some logic to add the local timezone if the parsed date time doesn't have a timezone (for which there are unit tests), so I'd say the latter is fine :-)

PR #10691, opened 2025-01-19
Fix developer documentation wrt docker-compose.

### LINKED ISSUE #10684: Fix docker compose command
In the documentation, the docker compose command is spelled as `docker-compose`. This should be `docker compose`.

fniessink:
Docker compose has been integrated into Docker as a subcommand for a while, but the developer documentation did not reflect that. Change `docker-compose` to `docker compose` in the documentation.

Fixes #10684.

### DIFF
--- a/docs/src/intro.md
+++ b/docs/src/intro.md
@@ -98,6 +98,7 @@ Implemented features include:
 - Side-by-side comparison of measurements at different points in time.
 - Integration with issue tracker (Jira only at the moment) to manage actions and technical debt.
 - Dark and light UI mode.
+- Runs on Docker and Kubernetes.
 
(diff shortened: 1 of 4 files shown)


docs/src/intro.md
```diff
@@ -105,7 +105,7 @@ For more plans, see the [issue tracker](https://github.com/ICTU/quality-time/iss
 
 # Trying it out
 
-*Quality-time* requires Docker and Docker-compose.
+*Quality-time* requires Docker.
```
wkoot: Since we support kubernetes, docker technically isn't a requirement but still the easiest way to deploy.. Perhaps something like 'can be tried out with docker locally' ?

PR #10696, opened 2025-01-20
Use browser locale to determine the first day of the week in date pic…

### LINKED ISSUE #7250: Date picker does not adapt first day of the week to user's locale
**Describe the bug**
In any browser, any date picker, displays Sunday as the first day of the week 

**To Reproduce**
1. Go to time travelling option in the top and open the date picker

**Expected behaviour**
Depending on the browser settings (user's locale), the week should either start with Monday or Sunday

see also: https://www.timeanddate.com/calendar/days/first-day-of-the-week.html

fniessink:
…kers.

Fixes #7250.

### DIFF
--- a/components/frontend/src/locale.js
+++ b/components/frontend/src/locale.js
@@ -1,0 +1,31 @@
+import "dayjs/locale/en-gb"
+import "dayjs/locale/nl"
+
+import dayjs from "dayjs"
+import localeData from "dayjs/plugin/localeData"
+import updateLocale from "dayjs/plugin/updateLocale"
+import { string } from "prop-types"
+
+export function adapterLocale(language) {
+    // Quick and dirty solution while we wait for a more generic solution to finding the user's locale
+    // Our issue: https://github.com/ICTU/quality-time/issues/10710
+    // Upstream issue: https://github.com/iamkun/dayjs/issues/732
+    let locale = "en-GB"
+    if (language === "nl-NL") {
+        // Use the Dutch locale, but keep the language English
+        locale = "nl" // dayjs doesn't have nl-NL
+        dayjs.extend(updateLocale)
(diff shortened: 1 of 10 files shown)


components/frontend/src/locale.js
```diff
@@ -0,0 +1,30 @@
+import "dayjs/locale/en-gb"
+import "dayjs/locale/nl"
+
+import dayjs from "dayjs"
+import localeData from "dayjs/plugin/localeData"
+import updateLocale from "dayjs/plugin/updateLocale"
+import { string } from "prop-types"
+
+export function adapterLocale(language) {
+    // Quick and dirty solution while we wait for a more generic solution to finding the user's locale
```
wkoot, line 10: Do we have a tracking issue to solve this ourselves?

components/frontend/src/locale.js
```diff
@@ -0,0 +1,30 @@
...
+    // https://github.com/iamkun/dayjs/issues/732
+    let locale = "en-GB"
+    if (language === "nl-NL") {
+        // Use the Dutch locale, but keep the language English
```
wkoot, line 15: Shouldn't we keep the language English for any locale? We're writing everything in English so it would be very weird if random words might be translated due to some locale

fniessink: Yes, what I was thinking exactly. And that is what the code is doing: we always use en-GB, except when the user's locale is nl-NL. Then we use the Dutch locale with the language parts (weekday names and months) reset to English. We can't generically do what we do for Dutch because of the issue mentioned in the code.

components/frontend/src/locale.js
fniessink: Good point. I have created an issue: https://github.com/ICTU/quality-time/issues/10710

PR #10697, opened 2025-01-21
Fix issue tracker error message being too narrow and remove duplicati…

fniessink:
…on between ErrorMessage and WarningMessage components.

No GitHub issue because this issue was not released yet.


components/frontend/src/widgets/WarningMessage.js
```diff
@@ -3,17 +3,18 @@ import { bool, string } from "prop-types"
 
 import { childrenPropType } from "../sharedPropTypes"
 
-export function WarningMessage({ children, title, showIf }) {
+export function WarningMessage({ children, pre, showIf, title }) {
```
wkoot, line 6: I would expect that the change of order in args would result in more changes, are all calls by keyword arg?

components/frontend/src/source/Source.js
```diff
@@ -104,9 +104,25 @@ function Parameters({
...
+            {parse_error && (
+                <Grid size={{ xs: 1, sm: 2, md: 2 }}>
+                    <WarningMessage pre title="Parse error">
+                        {parse_error}
+                    </WarningMessage>
+                </Grid>
+            )}
+            {config_error && (
+                <Grid size={{ xs: 1, sm: 2, md: 2 }}>
+                    <WarningMessage title="Configuration error">{config_error}</WarningMessage>
```
wkoot, line 123: Why not `{config_error}` on a newline like lines 110 and 117? Or also put those on the same line?

components/frontend/src/widgets/WarningMessage.js
fniessink: Yeah, reordering is no problem since React component props are passed by name: `<Component name={value}/>`.

components/frontend/src/source/Source.js
fniessink: Ah, that's my IDE's auto formatter. Apparently it stops putting a component on one line as soon as there is more than one prop passed. Parse errors and connections errors have tracebacks, so they get an extra `pre` prop.

PR #10791, opened 2025-02-06
Make the API-server return HTTP status 404 on non-existing endpoints …

### LINKED ISSUE #9860: API health endpoint doesn't report on health
**Describe the bug**
The health endpoint is not actually an endpoint, it just happens to return http 200 ok like any api url path.

**To Reproduce**
1. Navigate to `/api/health`
2. Navigate to `/api/randompage`
3. Observe that the return data is the same

**Expected behaviour**
JSON return data, describing the instance health such as
```json
{
  "health": "GREEN"
}
```

And, return a 404 for non-existing endpoints.

fniessink:
…instead of 200.

Fixes #9860.


components/api_server/src/healthcheck.py
```diff
@@ -8,7 +8,7 @@
 api_server_port = getenv("API_SERVER_PORT", "5001")
 exit_code = 1
 try:
-    with urlopen(f"http://localhost:{api_server_port}/api/health") as response:  # nosec B310
-        exit_code = 0 if response.status == HTTPStatus.OK else 1
+    with urlopen(f"http://localhost:{api_server_port}/api/internal/health") as response:  # nosec B310
+        exit_code = 0 if response.status == HTTPStatus.OK and response.read() == b'{"healthy": true}' else 1
```
wkoot, line 12: Yikes, this wasn't actually working?

components/api_server/src/routes/health.py
```diff
@@ -0,0 +1,9 @@
+"""Health end-point."""
+
+import bottle
+
+
+@bottle.get("/api/internal/health", authentication_required=False)
+def get_health():
+    """Return the API-server health."""
+    return {"healthy": True}
```
wkoot: Perhaps useful to inline comment that we don't monitor any additional processes, so up=up

components/api_server/tests/test_healthcheck.py
```diff
@@ -1,21 +1,48 @@
...
+        response.status = status
+        response.read.return_value = b'{"healthy": true}' if healthy else b'{"healthy": false}'
         response.__enter__.return_value = response
-        mock_urlopen.return_value = response
-        import healthcheck  # noqa: F401
+        return response
+
+    def tearDown(self) -> None:
+        """Remove the healthcheck module."""
+        del sys.modules["healthcheck"]
```
wkoot, line 28: Is the `del healthcheck` at the end of each test then still needed?

components/api_server/src/healthcheck.py
fniessink: It was working, but only looking at the status code, not the payload contents.

components/api_server/src/routes/health.py
fniessink: Done.

components/api_server/tests/test_healthcheck.py
fniessink: Good catch. It isn't needed.

PR #10827, opened 2025-02-11
Next to ignoring branches, tags, and jobs, also allow for including b…

### LINKED ISSUE #4520: Allow for including branches/tags and jobs when measuring failed CI-jobs with GitLab as source
Allow for including branches/tags and jobs:

- [x] #10825
- [x] #10826

fniessink:
…ranches, tags, and jobs when measuring failed CI-jobs with GitLab as source.

Closes #4520.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -14,6 +14,10 @@ If your currently installed *Quality-time* version is not the latest version, pl
 
 ## [Unreleased]
 
+### Added
+
+- Next to ignoring branches, tags, and jobs, also allow for including branches, tags, and jobs when measuring 'failed CI-jobs', 'unused CI-jobs', 'change failure rate', and 'jobs runs within time period' with GitLab as source. Closes [#4520](https://github.com/ICTU/quality-time/issues/4520).
+
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


docs/src/changelog.md
```diff
@@ -14,6 +14,10 @@ If your currently installed *Quality-time* version is not the latest version, pl
 
 ## [Unreleased]
 
+### Added
+
+- Next to ignoring branches, tags, and jobs, also allow for including branches, tags, and jobs when measuring failed CI-jobs with GitLab as source. Closes [#4520](https://github.com/ICTU/quality-time/issues/4520).
```
wkoot: Does the implementation in `GitLabJobsBase` only apply this to `failed CI-jobs` ? The change in data model suggests more metrics :)

wkoot (approved):

docs/src/changelog.md
fniessink: Yes, you're right. I'll have to see if we need unit tests for those other metrics as well.

fniessink: Extra unit tests not needed, the four metrics share a base class that test the new parameters.

PR #10833, opened 2025-02-12
When measuring CI-pipeline duration with GitLab as source, allow for …

### LINKED ISSUE #10426: Filter specific pipelines for metric CI-pipeline duration
When using the metric CI-pipeline duration with GitLab as source, and configuring scheduled pipeline runs, GitLab may return multiple pipelines and it is not possible to further select a specific pipeline.

A possibility to filter (scheduled) pipelines by description.
Also include the matched pipelines in the measurement entities.

**Additional context**
Unfortunately, the pipelines endpoints do not include the schedule id in the responses. So, to know which pipeline was triggered by which schedule, we need to retrieve the schedules via the [schedules endpoint](https://docs.gitlab.com/ee/api/pipeline_schedules.html) and then get the pipelines scheduled by the schedule via the [ pipelines triggered by a pipeline schedule endpoint](https://docs.gitlab.com/ee/api/pipeline_schedules.html#get-all-pipelines-triggered-by-a-pipeline-schedule). 

The scheduled pipelines response includes the pipeline schedule description, so users can use that to filter schedules. No need to use the schedule id.

fniessink:
…filtering by pipeline description.

Closes #10426.

### DIFF
--- a/components/collector/src/source_collectors/gitlab/base.py
+++ b/components/collector/src/source_collectors/gitlab/base.py
@@ -3,15 +3,14 @@
 from abc import ABC
 from dataclasses import dataclass, fields
 from datetime import datetime, timedelta
-from typing import cast
+from typing import Self, cast
 
 from dateutil.tz import tzutc
 
 from shared.utils.date_time import now
 
 from base_collectors import SourceCollector
-from collector_utilities.date_time import parse_datetime
-from collector_utilities.exceptions import CollectorError
+from collector_utilities.date_time import minutes, parse_datetime
 from collector_utilities.functions import add_query, match_string_or_regular_expression
 from collector_utilities.type import URL, Job
 from model import Entities, Entity, SourceResponses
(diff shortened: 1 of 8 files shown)


components/collector/src/source_collectors/gitlab/base.py
```diff
@@ -139,53 +138,58 @@ def _lookback_datetime(self) -> datetime:
...
     @property
     def datetime(self) -> datetime:
         """Return the datetime of the pipeline."""
         return parse_datetime(self.updated_at or self.created_at)
 
     @property
-    def pipeline_duration(self) -> timedelta:
+    def duration(self) -> timedelta:
         """Return the duration of the pipeline."""
-        return self.end - self.start if self.duration is None else timedelta(seconds=self.duration)
```
wkoot, line 188: Where did the old `self.duration` come from, was it something included in the json originally?

components/collector/src/source_collectors/gitlab/base.py
```diff
@@ -201,26 +205,69 @@ async def _landing_url(self, responses: SourceResponses) -> URL:
...
     async def _pipelines(self, responses: SourceResponses) -> list[Pipeline]:
         """Get the pipelines from the responses."""
         pipelines = []
-        try:
-            for response in responses:
-                json = await response.json()
-                # The GitLab pipelines endpoint returns a list or, if a pipeline id is passed, one pipeline. Harmonize:
-                json_list = json if isinstance(json, list) else [json]
-                pipelines.extend([Pipeline(**pipeline) for pipeline in json_list])
-        except StopAsyncIteration:
```
wkoot, line 213: We no longer need to guard against the `StopAsyncIteration`?

components/collector/src/source_collectors/gitlab/base.py
fniessink: In the previous version, Quality-time would get more detailed information about the pipeline with the `started_at`, `finished_at` and `duration` fields. However, this needs a call to the `pipeline/id` endpoint for each pipeline individually. Also, there is this open bug: https://gitlab.com/gitlab-org/gitlab/-/issues/19594. So I opted to simply used the `created_at` and `updated_at` fields as proxy for the pipeline duration.

components/collector/src/source_collectors/gitlab/base.py
fniessink: Apparently, the need to catch `StopAsyncIteration` was caused by our test code. It doesn't occur (so far) with testing against a live GitLab.

PR #10893, opened 2025-02-21
Don't install fonts in renderer image.

### LINKED ISSUE #10835: Check whether we need to install (all) fonts in the renderer component

fniessink:
Don't install fonts in the renderer image, but let Chromium use its default fonts for PDF exports. This reduces the renderer image size with about 44MB.

Closes #10835.

### DIFF
--- a/components/renderer/Dockerfile
+++ b/components/renderer/Dockerfile
@@ -3,19 +3,7 @@ FROM node:23.8.0-alpine3.21
 LABEL maintainer="Quality-time team <quality-time@ictu.nl>"
 LABEL description="Quality-time PDF render service"
 
-RUN apk add --no-cache \
-    msttcorefonts-installer=3.8.1-r1 \
-    fontconfig=2.15.0-r1 \
-    font-droid=20200215-r3 \
-    font-noto=24.7.1-r0 \
-    freetype=2.13.3-r0 \
-    ttf-dejavu=2.37-r5 \
-    ttf-freefont=20120503-r4 \
-    ttf-liberation=2.1.5-r2 \
-    chromium=132.0.6834.83-r0 && \
-    rm -rf /var/cache/apk/* /tmp/* && \
-    update-ms-fonts && \
-    fc-cache -f
+RUN apk add --no-cache chromium=132.0.6834.83-r0
 
 WORKDIR /home/renderer
 COPY package*.json /home/renderer/
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -26,6 +26,10 @@ If your currently installed *Quality-time* version is not the latest version, pl
 - Keep the footer at the bottom of the page even if the report is very short. Fixes [#10877](https://github.com/ICTU/quality-time/issues/10877).
 - Automatically expand long comments when exporting to PDF. Fixes [#10892](https://github.com/ICTU/quality-time/issues/10892).
 
+### Removed
+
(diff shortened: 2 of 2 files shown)


components/renderer/Dockerfile
```diff
@@ -3,19 +3,8 @@ FROM node:23.7.0-alpine3.21
 LABEL maintainer="Quality-time team <quality-time@ictu.nl>"
 LABEL description="Quality-time PDF render service"
 
-RUN apk add --no-cache \
-    msttcorefonts-installer=3.8.1-r1 \
-    fontconfig=2.15.0-r1 \
-    font-droid=20200215-r3 \
-    font-noto=24.7.1-r0 \
```
wkoot, line 10: Interesting that these font installs were mostly added for the `v3.35` and `v3.36` changes, but their requirement was never explicitly mentioned. Nice that we can get rid of it all :)

components/renderer/Dockerfile
```diff
@@ -3,19 +3,8 @@ FROM node:23.7.0-alpine3.21
...
-    freetype=2.13.3-r0 \
-    ttf-dejavu=2.37-r5 \
-    ttf-freefont=20120503-r4 \
-    ttf-liberation=2.1.5-r2 \
-    chromium=132.0.6834.83-r0 && \
-    rm -rf /var/cache/apk/* /tmp/* && \
-    update-ms-fonts && \
-    fc-cache -f
+RUN apk add --no-cache chromium=132.0.6834.83-r0 && \
+    rm -rf /var/cache/apk/* /tmp/*
```
wkoot: Because we're using `--no-cache`, there shouldn't be a need of the `rm` right? Or does chromium do something sneaky

docs/src/changelog.md
```diff
@@ -25,6 +25,10 @@ If your currently installed *Quality-time* version is not the latest version, pl
 - The software documentation was outdated (among other things the API-server health check endpoint). Fixes [#10858](https://github.com/ICTU/quality-time/issues/10858).
 - Keep the footer at the bottom of the page even if the report is very short. Fixes [#10877](https://github.com/ICTU/quality-time/issues/10877).
 
+### Removed
+
+- Don't install fonts in the renderer image, but let Chromium use its default fonts for PDF exports. This reduces the renderer image size with about 44MB. Closes [#10835](https://github.com/ICTU/quality-time/issues/10835).
```
wkoot:
```suggestion
- Don't install fonts in the renderer image, but let Chromium use its default fonts for PDF exports. Closes [#10835](https://github.com/ICTU/quality-time/issues/10835).
```

components/renderer/Dockerfile
fniessink: Many of the examples on how to use Chromium in Docker installed included installation of fonts. I never bothered to check before if it was really needed.

components/renderer/Dockerfile
fniessink: I checked: the image size is the same with or without the `rm`, so I'll remove the `rm` command.

docs/src/changelog.md
fniessink: I like having the reason in the change log so it's easier to remember why we removed the fonts. I can make it shorter though:
```suggestion
- To reduce the size of the renderer image, don't install fonts and let Chromium use its default fonts for PDF exports. Closes [#10835](https://github.com/ICTU/quality-time/issues/10835).
```

PR #10896, opened 2025-02-22
Automatically expand long comments when exporting to PDF.

### LINKED ISSUE #10892: Long comments are truncated in the PDF-export
- Report comments
- Subject comments
- Metric comments
- Measurement entity comments

fniessink:
Fixes #10892.

### DIFF
--- a/components/frontend/src/widgets/DivWithHTML.css
+++ b/components/frontend/src/widgets/DivWithHTML.css
@@ -1,0 +1,8 @@
+@media print {
+    /* Make sure no scrollbar appears and the HTML content is printed completely */
+    .with-html {
+        height: auto !important;
+        overflow: visible !important;
+        resize: none !important;
+    }
+}
--- a/components/frontend/src/widgets/DivWithHTML.jsx
+++ b/components/frontend/src/widgets/DivWithHTML.jsx
@@ -1,3 +1,5 @@
+import "./DivWithHTML.css"
+
 import { createRef, useLayoutEffect, useState } from "react"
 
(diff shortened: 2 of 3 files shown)


components/frontend/src/widgets/DivWithHTML.jsx
```diff
@@ -1,3 +1,5 @@
+import "./DivWithHTML.css"
```
wkoot, line 1: Huh, odd import that isn't referenced by name exactly. Is that because it's a css import, doesn't a top level css import suffice?

components/frontend/src/widgets/DivWithHTML.css
```diff
@@ -0,0 +1,8 @@
+@media print {
+    /* Make sure no scrollbar appears and the HTML content is printed completely */
+    .with-html {
+        height: auto !important;
+        overflow: visible !important;
```
wkoot, line 5: Does this never create issues i.e. with some flexbox with absolute sizing?

components/frontend/src/widgets/DivWithHTML.jsx
fniessink: This is indeed how importing CSS-files works in React applications.

components/frontend/src/widgets/DivWithHTML.css
fniessink: Never say never, but the `DivWithHTML` component is only used in the comment table cell, so I don't expect such issues.

PR #10909, opened 2025-02-25
Do not assume that Dependency-Track projects and components always ha…

### LINKED ISSUE #10848: latest version in Dependency-Track does not work any more
**Describe the bug**
Selecting the latest version results in a parse error, this used to work in Quality-time 5.23

Image

fniessink:
…ve a version number.

Fixes #10848.

### DIFF
--- a/components/collector/src/source_collectors/dependency_track/base.py
+++ b/components/collector/src/source_collectors/dependency_track/base.py
@@ -24,7 +24,7 @@ class DependencyTrackProject(TypedDict):
     lastBomImport: int
     name: str
     uuid: str
-    version: str
+    version: NotRequired[str]
     metrics: NotRequired[DependencyTrackMetrics]
 
 
@@ -87,7 +87,8 @@ class DependencyTrackBase(SourceCollector):
     def _project_matches(project: DependencyTrackProject, names: list[str], versions: list[str]) -> bool:
         """Return whether the project name matches the project names and versions."""
         project_matches_name = match_string_or_regular_expression(project["name"], names) if names else True
-        project_matches_version = match_string_or_regular_expression(project["version"], versions) if versions else True
+        project_version = project.get("version", "unknown")
+        project_matches_version = match_string_or_regular_expression(project_version, versions) if versions else True
         return project_matches_name and project_matches_version
 
     @staticmethod
--- a/components/collector/tests/source_collectors/dependency_track/test_dependencies.py
+++ b/components/collector/tests/source_collectors/dependency_track/test_dependencies.py
@@ -15,8 +15,8 @@ class DependencyTrackDependenciesTest(DependencyTrackTestCase):
         dependency: DependencyTrackComponent = {
             "name": "component name",
             "project": self.projects()[0],
-            "uuid": "component-uuid",
             "version": "1.0",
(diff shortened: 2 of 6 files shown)


components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -24,7 +24,7 @@ class DependencyTrackProject(TypedDict):
     lastBomImport: int
     name: str
     uuid: str
-    version: str
+    version: NotRequired[str]
```
wkoot, line 27: Does it not work to set `version: str = "unknown"` ?

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
```diff
@@ -92,3 +92,9 @@ async def test_filter_by_project_name_and_version(self):
         self.set_source_parameter("project_versions", ["1.3", "1.4"])
         response = await self.collect(get_request_json_side_effect=[self.projects(), self.dependencies("1.0")])
         self.assert_measurement(response, value="1", entities=self.entities("1.0", "up-to-date"))
+
+    async def test_filter_by_project_version_when_project_has_no_version(self):
+        """Test filtering projects by version."""
+        self.set_source_parameter("project_versions", ["1.2", "1.3"])
+        response = await self.collect(get_request_json_return_value=self.projects(version=""))
+        self.assert_measurement(response, value="0", entities=[])
```
wkoot, line 118: Shouldn't we have at least one value returned so we know it can actually match something?

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
```diff
@@ -20,7 +20,7 @@ class DependencyTrackSourceUpToDatenessVersionTest(DependencyTrackTestCase):
     METRIC_TYPE = "source_up_to_dateness"
     LANDING_URL = "https://dependency_track"
 
-    def projects(self) -> list[DependencyTrackProject]:
+    def projects(self, version: str = "1.4") -> list[DependencyTrackProject]:
```
wkoot, line 23: Do we test with version not being given?

components/collector/src/source_collectors/dependency_track/base.py
fniessink: No, `TypedDict` does not allow for setting default values.

components/collector/tests/source_collectors/dependency_track/test_dependencies.py
fniessink: Added a test that matches the version number.

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
fniessink: Yes:

```python
    async def test_filter_by_project_version_when_project_has_no_version(self):
        """Test filtering projects by version."""
        self.set_source_parameter("project_versions", ["1.2", "1.3"])
        response = await self.collect(get_request_json_return_value=self.projects(version=""))
        self.assert_measurement(response, value="0", entities=[])
```

PR #10913, opened 2025-02-26
Gatling measurement details would not be displayed.

### LINKED ISSUE #10911: Gatling transaction details not visible
**Describe the bug**
Gatling transaction details are not visible in Quality-Time. The transactions are visible, the details not. For example: sample count, error count or percentiles.

**To Reproduce**
1. Open Gatling source tab
2. Details are absent

**Expected behaviour**
Details are present.

**Screenshots**

Image

 - OS: MacOS
 - Browser: Chrome
 - Quality-time: v5.25.0

fniessink:
Fixes #10911.


components/collector/src/source_collectors/jmeter_json/slow_transactions.py
```diff
@@ -18,8 +18,8 @@ def _parse_json(self, json: JSON, filename: str) -> Entities:
             entity = TransactionEntity(
                 key=transaction["transaction"],
                 name=transaction["transaction"],
-                sample_count=transaction["sampleCount"],
-                error_count=transaction["errorCount"],
+                sample_count=str(transaction["sampleCount"]),
```
wkoot, line 21: Seems like we make the entity the same way three times, not something we should extract?

components/collector/tests/source_collectors/gatling/test_slow_transactions.py
```diff
@@ -15,30 +15,30 @@ def setUp(self):
...
-                "mean_response_time": 110.0,
-                "min_response_time": 50.0,
-                "max_response_time": 250.0,
-                "percentile_50_response_time": 100.0,
-                "percentile_75_response_time": 115.0,
-                "percentile_95_response_time": 135.0,
-                "percentile_99_response_time": 195.0,
+                "sample_count": "123",
+                "error_count": "2",
+                "error_percentage": str(round((2.0 / 123) * 100, 1)),
```
wkoot, line 20: Shouldn't the `round()` no longer be necessary since we do it in the collector?

components/collector/src/source_collectors/jmeter_json/slow_transactions.py
fniessink: The code is similar, but not exactly the same. Extracting the commonalities does not look trivial.

components/collector/tests/source_collectors/gatling/test_slow_transactions.py
fniessink: Not sure what you mean: this is a unit test for the Gatling slow transactions collector...

PR #10914, opened 2025-02-26
When adding a new source to a metric, don't show the spinner until th…

### LINKED ISSUE #9994: When adding a new source, the spinner starts spinning before a connection/change is made
**Describe the bug**
When adding a new source, the spinner starts spinning before a connection/change is made

**To Reproduce**
1. create metric Security Warnings
2. Add new source ZAP
3. See spinner start spinning before any change is made

**Expected behaviour**
Wait until toaster says: Connection OK

**Screenshots**
image

fniessink:
…e source has been configured sufficiently to start collecting data.

Fixes #9994.

### DIFF
--- a/components/frontend/src/utils.jsx
+++ b/components/frontend/src/utils.jsx
@@ -4,6 +4,7 @@ import { PERMISSIONS } from "./context/Permissions"
 import { defaultDesiredResponseTimes } from "./defaults"
 import { STATUSES_NOT_REQUIRING_ACTION } from "./metric/status"
 import {
+    dataModelPropType,
     datePropType,
     metricPropType,
     metricsPropType,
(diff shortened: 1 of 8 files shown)


components/frontend/src/utils.jsx
```diff
@@ -104,12 +105,34 @@ export function isMeasurementOutdated(metric) {
...
+    // all mandatory parameters configured
+    const sources = Object.values(metric.sources ?? {})
+    if (sources.length === 0) {
+        return false
+    }
+    return sources.every((source) => {
+        const parameters = dataModel.sources[source.type].parameters
+        return Object.entries(parameters).every(([parameterKey, parameter]) => {
+            if (parameter.mandatory && parameter.metrics.includes(metric.type) && !parameter.default_value) {
+                return source?.parameters?.[parameterKey]
```
wkoot, line 125: Why return the value here? Do we ever return `false`?

fniessink: This is how `every` works, it takes a callback for each item that should return true or false, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

PR #10931, opened 2025-02-28
When measuring average issue lead time with Jira as source, the lead …

### LINKED ISSUE #10929: Issue lead time not collected
**Describe the bug**
When measuring the average issue lead time, the lead time of individual issues is not collected.

**Screenshots**

Image

 - Quality-time version v5.25.0

fniessink:
…times per issue would not be collected.

Fixes #10929.

### DIFF
--- a/components/collector/src/source_collectors/jira/average_issue_lead_time.py
+++ b/components/collector/src/source_collectors/jira/average_issue_lead_time.py
@@ -24,7 +24,7 @@ class JiraAverageIssueLeadTime(JiraIssues):
     def _create_entity(self, issue: dict, url: URL) -> Entity:
         """Extend to also add the lead time to the entity."""
         entity = super()._create_entity(issue, url)
-        entity["lead_time"] = self.__lead_time(issue["fields"])
+        entity["lead_time"] = str(self.__lead_time(issue["fields"]))
         return entity
 
     @classmethod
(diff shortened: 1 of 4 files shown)


components/collector/src/source_collectors/jira/average_issue_lead_time.py
```diff
@@ -24,7 +24,7 @@ def _include_issue(self, issue: dict) -> bool:
     def _create_entity(self, issue: dict, url: URL) -> Entity:
         """Extend to also add the lead time to the entity."""
         entity = super()._create_entity(issue, url)
-        entity["lead_time"] = self.__lead_time(issue["fields"])
+        entity["lead_time"] = str(self.__lead_time(issue["fields"]))
```
wkoot, line 27: Did we get warnings about this typing issue in logs? If not, does it make sense to add such?

wkoot (approved):

components/collector/src/source_collectors/jira/average_issue_lead_time.py
fniessink: The problem is that entity values that are not of type `Value` (which is `str | None`) are allowed, but not stored in the database (see below). When we introduced that change we didn't check thoroughly enough whether all attributes were already of the type `Value`, so missed places where entity attributes were of type `float` or `int`. 

```python
@dataclass
class SourceMeasurement:
    """Class to hold measurement values, entities, and error messages from collecting the measurement from a source."""

    # ...

    def entities_to_store(self) -> JSONList:
        """Return the entities which should be stored, omitting ephemeral attributes."""
        return [
            # only include the attributes of type Value, so native objects like datetime can be used as ephemeral attrs
            {key: val for key, val in entity.items() if isinstance(val, Value)}
            for entity in self.get_entities()
        ]
```

PR #10943, opened 2025-03-03
Replace react-timeago with dayjs functionality.

fniessink:


components/frontend/src/issue/IssueStatus.jsx
```diff
@@ -59,7 +62,7 @@ function cardDetails(issueStatus, settings) {
         details.push(
             <ListItem key="created">
                 <Typography noWrap variant="inherit">
-                    Created <TimeAgo date={issueStatus.created} />
+                    Created {dayjs(issueStatus.created).fromNow()}
```
wkoot, line 65: Shouldn't we create the `fromNow` string once, outside all the if statements, e.g. on line 55?

fniessink: That's not possible, because `fromNow` is applied to different dates (`issueStatus.created`, `issueStatus.updated`, etc.).

PR #10991, opened 2025-03-17
Use component and module names for logging in Python components.

fniessink:
Ruff 0.10.0 frowns upon log calls that log to the root logger. Use component and module names to construct a non-root logger name for each log call.

### DIFF
--- a/components/api_server/src/model/issue_tracker.py
+++ b/components/api_server/src/model/issue_tracker.py
@@ -2,13 +2,16 @@
 
 from dataclasses import asdict, dataclass, field
 from typing import cast
-import logging
 
 from shared.utils.functions import first
 
 import requests
 
+from utils.log import get_logger
 from utils.type import URL
+
+
+logger = get_logger()
 
 
 @dataclass
@@ -128,7 +131,7 @@ class IssueTracker:
             response_json = self.__post_json(api_url, json)
         except Exception as reason:
             error = str(reason) if str(reason) else reason.__class__.__name__
-            logging.warning("Creating a new issue at %s failed: %s", api_url, error)
+            logger.warning("Creating a new issue at %s failed: %s", api_url, error)
             return "", error
         return response_json["key"], ""  # pragma: no feature-test-cover
 
@@ -148,7 +151,8 @@ class IssueTracker:
         try:
             json = cast(JiraIssueSuggestionJSON, self.__get_json(api_url))
         except Exception as reason:
-            logging.warning("Retrieving issue id suggestions from %s failed: %s", api_url, reason)
+            message = "Retrieving issue id suggestions from %s failed: %s"
+            logger.warning(message, self.suggestions_api % (self.url, "<query redacted>"), reason)
             return []
         return self._parse_suggestions(json)  # pragma: no feature-test-cover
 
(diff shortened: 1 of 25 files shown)


components/api_server/src/model/issue_tracker.py
```diff
@@ -148,7 +151,8 @@ def get_suggestions(self, query: str) -> list[IssueSuggestion]:
         try:
             json = cast(JiraIssueSuggestionJSON, self.__get_json(api_url))
         except Exception as reason:
-            logging.warning("Retrieving issue id suggestions from %s failed: %s", api_url, reason)
+            message = "Retrieving issue id suggestions from %s failed: %s"
+            logger.warning(message, self.suggestions_api % (self.url, "<query redacted>"), reason)
```
wkoot, line 155: Did the query include possible secrets? Also, the inline string format this way will probably evaluate on runtime instead of when written to logger; then we might as well use f-string for readability

components/notifier/src/destinations/ms_teams.py
```diff
@@ -40,9 +39,10 @@ def create_connector_card(destination: str, notification: Notification) -> pymst
...
     """Send notification to Microsoft Teams using a Webhook."""
-    logging.info("Sending notification to configured teams webhook")
+    logger = get_logger()
+    logger.info("Sending notification to configured teams webhook")
     card = create_connector_card(destination, notification)
     try:
         card.send()
     except Exception:
-        logging.exception("Could not deliver notification")
+        logger.exception("Could not deliver notification")
```
wkoot, line 48: Wow, that's exotic. Is there an added benefit to using `exception` over `error`?

components/shared_code/tests/shared/utils/test_log.py
```diff
@@ -0,0 +1,23 @@
...
+    def test_get_logger(self):
+        """Test getting the logger."""
+        self.assertEqual("quality_time.shared_code.tests.shared.utils.test_log", get_logger("shared_code").name)
+
+    def test_get_logger_indirectly(self):
+        """Test getting the logger via an extra function call."""
+
+        def get_logger_indirectly() -> logging.Logger:
+            """Return the logger."""
+            return get_logger("shared_code", call_stack_depth=1)
```
wkoot: Shouldn't this use a different `component_name`, so the test outcome is actually different from default behaviour?

components/api_server/src/model/issue_tracker.py
fniessink: SonarQube started complaining about possible logging injection because the query is user-provided data. I decided that we probably don't need to exact query in the logging anyway, so removed it from the logged message. 

Using a f-string is certainly possible, but since `suggestions_api` is part of a bunch of URLs and endpoints defined as class attributes I don't want to change just one to an f-string.

components/notifier/src/destinations/ms_teams.py
fniessink: That you haven't encountered it before doesn't make it exotic :-)

See https://docs.python.org/3/library/logging.html#logging.Logger.exception

components/shared_code/tests/shared/utils/test_log.py
fniessink: Good point.

PR #10994, opened 2025-03-17
Show an error message when measurement entities cannot be loaded.

### LINKED ISSUE #10478: Suppressed warning says it has no entities when it should
**Describe the bug**
When the browser has lost its connection, the suppressed warning metric says that it has no entities when it should. 

**To Reproduce**
1. Go to metric Suppressed warnings (SonarQube source)
2. Make sure the metric is collapsed and it has a measured value above 0
3. Make sure the browser lost its connection (reproducible by killing the API-server)
4. Open the entities tab and see that it say it never had entities, where the user should have an error (i.e. I know there is something to display, and I know for this metric it should be possible, but I cannot fetch the results).

Image

**Expected behaviour**
Different metrics, like the suppressed warnings metric, should know they are supposed to have entities. When the measured value is above 0 the entities tab should display an error message (⚠️) because it knows that entities should be present, but it cannot show then for some reason (like a connection loss).

fniessink:
Fixes #10478.

### DIFF
--- a/components/frontend/src/metric/MetricDetails.test.jsx
+++ b/components/frontend/src/metric/MetricDetails.test.jsx
@@ -12,12 +12,14 @@ import * as source_api from "../api/source"
 import { DataModel } from "../context/DataModel"
 import { EDIT_ENTITY_PERMISSION, EDIT_REPORT_PERMISSION, Permissions } from "../context/Permissions"
 import { expectNoAccessibilityViolations } from "../testUtils"
+import * as toast from "../widgets/toast"
 import { MetricDetails } from "./MetricDetails"
 
 vi.mock("../api/fetch_server_api.js")
 vi.mock("../api/changelog.js")
 vi.mock("../api/measurement.js")
 vi.mock("../api/source.js")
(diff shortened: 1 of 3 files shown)


components/frontend/src/metric/MetricDetails.test.jsx
```diff
@@ -12,12 +12,14 @@ import * as source_api from "../api/source"
 import { DataModel } from "../context/DataModel"
 import { EDIT_ENTITY_PERMISSION, EDIT_REPORT_PERMISSION, Permissions } from "../context/Permissions"
 import { expectNoAccessibilityViolations } from "../testUtils"
+import * as toast from "../widgets/toast"
```
wkoot, line 15: yikes `import *`, well at least it's test code

fniessink: This is the pattern we use for mocking modules. But it may well be an outdated pattern. I don't see it in the Jest docs: https://jestjs.io/docs/mock-functions. Changing it is out of scope for this issue, but I'll open a separate issue.

PR #11006, opened 2025-03-18
Increase limit of 100 measurement entities.

### LINKED ISSUE #6278: Increase limit of 100 measurement entities
# User story

Currently, Quality-time keeps track of at most 100 measurement entities per metric. Which 100 is arbitrary. This makes it impossible to do complete management of technical debt via Quality-time.

Removing the limit completely causes a few problems however when testing with 7000 measurement entities:
- I'm getting `ProtocolError: Protocol error (Page.printToPDF): Printing failed` with larger number of measurement entities (tried 7000). Possible related Puppeteer issue: https://github.com/puppeteer/puppeteer/issues/8070.
- Rendering and scrolling is pretty slow with 7000 measurement entities.
- SonarQube collectors assume no pagination is needed because we don't store more than 100 issues anyway. It gets the first 500 violations from SonarQube. Related issue: #11004. 

So instead of removing the limit we increase it to 250. If all goes well, we can later increase it to 500, 1000, etc.

## Assumptions

- Measurement entities can be downloaded fast enough without the need for pagination: this seems to be no problem with 7000 measurement entities.
- Measurement entities can be shown fast enough without the need for pagination: this is not true for 7000 measurement entities.
- Extra database growth will be limited since not many metrics have a large number of measurement entities.

## Out of scope

- Implement pagination and search for measurement entities, this is described in #10686.

# Tasks

### Collector

- [x] Increase the limit to 250 entities

## Documentation

- [x] Update changelog
- [x] Check docs for mention of the limit

fniessink:
Increase the limit of 100 measurement entities per metric source.

Closes #6278.

### DIFF
--- a/components/collector/src/source_collectors/azure_devops/issues.py
+++ b/components/collector/src/source_collectors/azure_devops/issues.py
@@ -7,7 +7,7 @@ import aiohttp
 from base_collectors import SourceCollector
 from collector_utilities.functions import iterable_to_batches
 from collector_utilities.type import URL, Value
-from model import Entities, Entity, SourceMeasurement, SourceResponses
+from model import Entities, Entity, SourceResponses
 
 
 class AzureDevopsIssues(SourceCollector):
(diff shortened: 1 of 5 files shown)


components/collector/src/source_collectors/azure_devops/issues.py
```diff
@@ -46,8 +46,7 @@ def _item_select_fields(self) -> list[str]:
     async def _get_work_item_responses(self, auth: aiohttp.BasicAuth) -> SourceResponses:
         """Separately get each work item from the API."""
         api_url = await self._api_wit_url(endpoint="workitemsbatch")
-        batch_size = min(self.MAX_IDS_PER_WORK_ITEMS_API_CALL, SourceMeasurement.MAX_ENTITIES)
```
wkoot, line 49: Do we not want to retain this dynamic resolve?

fniessink: No, because I think this was a (harmless) bug: why base the batch size with which to get the work items from Azure DevOps Server on the maximum number of measurement entities if we still get all the work items anyway?

PR #11016, opened 2025-03-20
Make header of measurement entity table sticky.

fniessink:


components/frontend/src/subject/SubjectTitle.jsx
```diff
@@ -71,38 +74,40 @@ export function SubjectTitle({
...
+                    tabs={[
+                        { label: "Configuration", icon: <SettingsIcon /> },
+                        { label: "Changelog", icon: <HistoryIcon /> },
+                    ]}
+                    uuid={subject_uuid}
+                >
+                    <SubjectParameters
+                        subject={subject}
+                        subject_uuid={subject_uuid}
+                        subject_name={subjectName}
```
wkoot, line 96: Mixed snake case and camel case?

components/frontend/src/subject/Subject.jsx
```diff
@@ -1,5 +1,3 @@
-import "./Subject.css"
```
wkoot, line 1: This file seems to have already been deleted before, was this not throwing an error?

wkoot (approved):

components/frontend/src/subject/Subject.jsx
fniessink: The file was renamed.

components/frontend/src/subject/SubjectTitle.jsx
fniessink: Yeah, the frontend has a lot of that. Maybe it's best to create a separate PR for that.

PR #11108, opened 2025-04-10
Show up-to-dateness of individual DT-projects

### LINKED ISSUE #10545: Last BOM import results from Dependency-Track unclear
**Describe the bug**
When using multiple projects in Dependency-Track, with:

Metric type: Source up-to-dateness
Source type: Dependency Track
Project event type(s): last BOM import

Then **all** projects are displayed in the Dependency-Track tab, regardless whether they meet the metric target or not.
The metric may indicate (with the correct metric target) which project has the oldest BOM import, based on the number of days.

The above causes confusion, because other metrics, most of the time, show the number of violations in the metric and only the findings in the detail tab.

**Possible adjustment(s)**

- Make metrics configurable for:
  - Metric scale: Count and Metric unit: days
  - Metric scale: Count and Metric unit: number of violations
- Keep the implementation as-is and in the detail tab provide the violations with a color with the deviations from the metric target
- Keep implementation as-is and use a toggle to show only violations in the detail tab, or all projects
- ..

fniessink:
When measuring source-up-to-dateness with Dependency-Track as source, also show the up-to-dateness of individual projects in the measurement details.

Closes #10545.

### DIFF
--- a/components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
+++ b/components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
@@ -2,6 +2,9 @@
 
 from collections.abc import Sequence
 from datetime import datetime
+from typing import Literal
+
+from shared_data_model import DATA_MODEL
 
 from base_collectors import TimePassedCollector
 from collector_utilities.date_time import datetime_from_timestamp
@@ -47,6 +50,7 @@ class DependencyTrackSourceUpToDateness(DependencyTrackBase, TimePassedCollector
                     project_landing_url=f"{landing_url}/projects/{uuid}",
                     version=project.get("version", ""),
                     is_latest="true" if self._is_latest(project) else "false",
+                    up_to_date=self._up_to_dateness(project),
                 )
                 entities.append(entity)
         return entities
@@ -87,3 +91,23 @@ class DependencyTrackSourceUpToDateness(DependencyTrackBase, TimePassedCollector
     def _is_latest(project: DependencyTrackProject) -> bool:
         """Return whether the project is the latest version."""
         return project.get("isLatest", False)
+
+    def _up_to_dateness(self, project: DependencyTrackProject) -> str:
+        """Return the up-to-dateness of the project."""
+        date_times = self._project_event_datetimes(project)
+        if not date_times:
+            return "unknown"
+        # Use the minimum of the project's last BOM-import and last BOM-analysis dates as the project's up-to-dateness:
+        min_date_time = self.days(self.minimum(date_times))
+        if self._target_met(min_date_time, "target"):
+            return "yes"
+        if self._target_met(min_date_time, "near_target"):
+            return "nearly"
+        return "no"
+
+    def _target_met(self, value: int, target_type: Literal["target", "near_target"]) -> bool:
+        """Return whether the value meets the target value."""
+        target = int(self._metric.get(target_type) or getattr(DATA_MODEL.metrics[self._metric["type"]], target_type))
+        # In practice, the direction of the metric should always be "<" (smaller is better), but the user can change
+        # the direction, so better be prepared:
+        return value <= target if self.metric_direction == "<" else value >= target
--- a/components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
+++ b/components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
@@ -60,6 +60,7 @@ class DependencyTrackSourceUpToDatenessVersionTest(DependencyTrackTestCase):
                 "is_latest": "true",
                 "project": "Project 1",
                 "project_landing_url": "/projects/p1",
+                "up_to_date": "yes",
                 "version": "1.1",
             },
             {
@@ -69,6 +70,7 @@ class DependencyTrackSourceUpToDatenessVersionTest(DependencyTrackTestCase):
                 "last_bom_import": self.last_week.isoformat(),
                 "project": "Project 2",
                 "project_landing_url": "/projects/p2",
+                "up_to_date": "nearly",
                 "version": "1.2",
             },
         ]
@@ -100,6 +102,7 @@ class DependencyTrackSourceUpToDatenessVersionTest(DependencyTrackTestCase):
                 "is_latest": "true",
                 "project": "Project 3",
                 "project_landing_url": "/projects/p3",
+                "up_to_date": "nearly",
                 "version": "1.3",
             }
         )
@@ -125,6 +128,7 @@ class DependencyTrackSourceUpToDatenessVersionTest(DependencyTrackTestCase):
                 "is_latest": "false",
                 "project": "Project 3",
                 "project_landing_url": "/projects/p3",
+                "up_to_date": "unknown",
                 "version": "1.3",
             }
         )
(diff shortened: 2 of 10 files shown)


components/shared_code/src/shared_data_model/sources/dependency_track.py
```diff
@@ -138,6 +138,15 @@
                 EntityAttribute(name="Latest", key="is_latest", type=EntityAttributeType.BOOLEAN),
                 EntityAttribute(name="Last BOM import", type=EntityAttributeType.DATETIME),
                 EntityAttribute(name="Last BOM analysis", type=EntityAttributeType.DATETIME),
+                EntityAttribute(
+                    name="Up-to-date",
+                    color={
+                        "almost": Color.WARNING,
```
wkoot: Don't we use the term "near" instead of "almost", or is that only for parameter and not value?

components/shared_code/src/shared_data_model/meta/entity.py
```diff
@@ -58,7 +58,7 @@ class EntityAttribute(NamedModel):
     def set_key(self) -> Self:
         """Set the key to the lower case version of the name if there's no key."""
         if self.key is None:
-            self.key = self.name.lower().replace(" ", "_")
+            self.key = self.name.lower().replace(" ", "_").replace("-", "_")
```
wkoot, line 61: Are there no attributes that need to be migrated for this, or are they mapped automatically between versions (i.e. when time travelling or comparing with older dates which are on older datamodel format)?

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
```diff
@@ -87,3 +91,21 @@ def _timestamp_to_datetime(timestamp: int | None) -> datetime | None:
     def _is_latest(project: DependencyTrackProject) -> bool:
         """Return whether the project is the latest version."""
         return project.get("isLatest", False)
+
+    def _up_to_dateness(self, project: DependencyTrackProject) -> str:
+        """Return the up-to-dateness of the project."""
+        date_times = self._project_event_datetimes(project)
+        if not date_times:
+            return "unknown"
+        min_date_time = self.days(self.minimum(date_times))
```
wkoot, line 101: Why do we compare on the minimum? I'd expect the latest event on a project to be polled, or are these project events somehow overwritten with each SBOM import and/or analysis?

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
```diff
@@ -87,3 +91,21 @@ def _timestamp_to_datetime(timestamp: int | None) -> datetime | None:
...
+        if self._target_met(min_date_time, "target"):
+            return "yes"
+        if self._target_met(min_date_time, "near_target"):
+            return "almost"
+        return "no"
+
+    def _target_met(self, value: int, target_type: Literal["target", "near_target"]) -> bool:
+        """Return whether the value meets the target value."""
```
wkoot, line 109: Since this method is only used by up to dateness, the direction should always be `<` right? Now it looks like this is actually quite a generic method that belongs in a collector baseclass

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
```diff
@@ -87,3 +91,21 @@ def _timestamp_to_datetime(timestamp: int | None) -> datetime | None:
...
+        target = int(self._metric.get(target_type) or getattr(DATA_MODEL.metrics[self._metric["type"]], target_type))
+        direction = str(self._metric.get("direction") or DATA_MODEL.metrics[self._metric["type"]].direction.value)
+        return value <= target if direction == "<" else value >= target
```
wkoot: If for some reason the up to dateness could have a direction `>`, shouldn't this also invert the min (or max) selector on line 100?

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
```diff
@@ -181,3 +188,12 @@ async def test_source_up_to_dateness_with_pagination(self):
         response = await self.collect(get_request_json_return_value=self.projects())
         self.assert_measurement(response, value="7", landing_url=self.LANDING_URL, entities=self.entities())
         DependencyTrackBase.PAGE_SIZE = default_size
+
+    async def test_source_up_to_dateness_with_reversed_direction(self):
+        """Test that the source up-to-dateness can be measured if the metric direction has been reversed."""
+        self.metric["direction"] = ">"
```
wkoot, line 194: Huh but what does `up-to-dateness` in a descending manner mean? That it's **not** supposed to be more up to date than a certain timeframe 🤔

components/shared_code/src/shared_data_model/meta/entity.py
fniessink: No, I checked. This is the first time we use hyphens in attribute names.

components/shared_code/src/shared_data_model/sources/dependency_track.py
fniessink: Yeah, I thought about that, but given that the attribute name is "Up-to-date", "near" doesn't make sense as an answer to the question "Is the the project's BOM up-to-date?". Maybe "nearly"?

By the way, I've also considered making the attribute name "Status", but that conflicts with the suppression status that all measurement entities have ("fixed", "false positive", etc.).

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
fniessink: Because DT has two event types for each project: the last BOM-import and the last BOM-analysis. For each event type there is one date. So `date_times` is at most two dates. If the user wants to take both event types into account for the up-to-dateness, it makes sense to look at the minimum, because that is the age of the last analysis of the last imported BOM.

components/shared_code/src/shared_data_model/sources/dependency_track.py
wkoot: What about using something like a label "age" with a values such as "new", "recent" and "old"? I think the trouble is mostly due to the term "up-to-dateness" which is uneasy, although it is an official noun.. But I'm also inclined to agree that all other options are worse than what you already chose

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
wkoot: Right, makes sense when known. Perhaps worth adding an inline comment or docstring to explain

components/shared_code/src/shared_data_model/sources/dependency_track.py
fniessink: Ha, I considered "Age" as well, but "new", "recent", and "old" don't really cover what is happening, namely that the age is compared to the metric target. If the target is 1 day, 2 days would already be listed as "old" which sounds weird. I'll change "almost " to "nearly" and leave it at that.

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
fniessink: You're right, but unfortunately the user can change the direction of the source up-to-dateness metric. We have no way (yet) to make changing the direction impossible for certain metric types.

fniessink: I'll add a comment.

components/collector/src/source_collectors/dependency_track/source_up_to_dateness.py
fniessink: Good point. Until now, source collectors didn't have access to the metric parameters, but now they do, so may just as well use that in the `minimum()` method.

components/collector/tests/source_collectors/dependency_track/test_source_up_to_dateness.py
fniessink: As mentioned above, there is currently no way to prevent users from changing the direction.

PR #11120, opened 2025-04-11
Improve mock handling in the frontend unit tests.

### LINKED ISSUE #11000: Follow best practices for mocking with Jest
See https://jestjs.io/docs/mock-functions.

fniessink:
Mock individual methods instead of complete modules where possible.

Closes #11000.

### DIFF
--- a/components/frontend/src/App.test.jsx
+++ b/components/frontend/src/App.test.jsx
@@ -3,15 +3,10 @@ import dayjs from "dayjs"
 import history from "history/browser"
 import { vi } from "vitest"
 
-import * as auth from "./api/auth"
 import * as fetchServerApi from "./api/fetch_server_api"
 import App from "./App"
 import { expectNoAccessibilityViolations } from "./testUtils"
 import * as toast from "./widgets/toast"
-
-vi.mock("./api/fetch_server_api.js")
-vi.mock("./api/auth.js")
-vi.mock("./widgets/toast.jsx")
 
 function setUserInLocalStorage(sessionExpirationDatetime, email) {
     localStorage.setItem("user", "admin")
@@ -28,18 +23,12 @@ beforeAll(() => {
 
 beforeEach(async () => {
     history.push("")
-    fetchServerApi.apiWithReportDate = (await vi.importActual("./api/fetch_server_api.js")).apiWithReportDate
-    fetchServerApi.fetchServerApi = vi.fn().mockReturnValue({
+    vi.spyOn(fetchServerApi, "fetchServerApi").mockReturnValue({
         then: vi.fn().mockReturnValue({ catch: vi.fn().mockReturnValue({ finally: vi.fn() }) }),
-    })
-    auth.login = vi.fn().mockResolvedValue({
-        ok: false,
     })
 })
 
-afterEach(() => {
-    vi.restoreAllMocks()
-})
+afterEach(() => vi.restoreAllMocks())
 
 it("shows spinner", async () => {
     const { container } = render(<App />)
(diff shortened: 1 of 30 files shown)


components/frontend/src/App.test.jsx
```diff
@@ -28,18 +23,12 @@ beforeAll(() => {
 
 beforeEach(async () => {
     history.push("")
-    fetchServerApi.apiWithReportDate = (await vi.importActual("./api/fetch_server_api.js")).apiWithReportDate
-    fetchServerApi.fetchServerApi = vi.fn().mockReturnValue({
+    vi.spyOn(fetchServerApi, "fetchServerApi").mockReturnValue({
```
wkoot, line 26: This `spyOn` is an improvement that does all the other things automagically?

components/frontend/src/header_footer/buttons/DownloadAsPdfButton.test.jsx
```diff
@@ -6,15 +6,14 @@ import * as fetchServerApi from "../../api/fetch_server_api"
...
-
 beforeEach(() => {
-    vi.resetAllMocks()
     history.push("")
-    fetchServerApi.fetchServerApi.mockImplementation(() => Promise.resolve({ ok: true }))
+    vi.spyOn(fetchServerApi, "fetchServerApi").mockImplementation(() => Promise.resolve({ ok: true }))
+    vi.spyOn(toast, "showMessage")
 })
 
+afterEach(() => vi.resetAllMocks())
```
wkoot: I don't understand when `afterEach(() => vi.resetAllMocks())` is used; it seems unneeded when a `spyOn` mock promise is placed in a `beforeAll`, but only occasionally used when `beforeEach` contains a `spyOn` mock promise?

components/frontend/src/widgets/toast.test.jsx
```diff
@@ -6,10 +6,9 @@ import { showConnectionMessage, showMessage } from "./toast"
 
 vi.mock("react-toastify")
 
-beforeEach(() => {
-    vi.resetAllMocks()
-    history.push("")
-})
+beforeEach(() => history.push(""))
+
+afterEach(() => vi.resetAllMocks())
```
wkoot, line 11: What is reset about the mocks here, is it their call content? Funny that doesn't automatically happen after each test case

components/frontend/src/changelog/ChangeLog.test.jsx
```diff
@@ -72,13 +70,13 @@ it("loads more changes", async () => {
...
+    fetchServerApi.fetchServerApi.mockImplementation(() => Promise.reject(new Error("Couldn't retrieve changelog")))
+    const showMessage = vi.spyOn(toast, "showMessage")
     await act(async () => fireEvent.click(screen.getByText(/Load more changes/)))
     await waitFor(async () => {
-        expect(toast.showMessage).toHaveBeenCalledTimes(1)
+        expect(showMessage).toHaveBeenCalledTimes(1)
         await expectNoAccessibilityViolations(container)
     })
-    expectNrEventsToBe(0)
+    expectNrEventsToBe(1)
```
wkoot, line 79: Why does this change now mean there was an event?

components/frontend/src/App.test.jsx
fniessink: `spyOn` mocks an individual method, where before we would mock the whole module. So in the case above, we don't need to revert the mocking of `apiWithReportDate` anymore.

components/frontend/src/header_footer/buttons/DownloadAsPdfButton.test.jsx
fniessink: I've added it here to be consistent I think. I'll check whether these `resetAllMock` are really necessary.

fniessink: You're right. No `resetAllMocks` is necessary when using `spyOn`. Only when using `vi.mock(module)` it is needed.

components/frontend/src/widgets/toast.test.jsx
fniessink: This one has been removed, see previous comment.

components/frontend/src/changelog/ChangeLog.test.jsx
fniessink: Because the default `mockImplementation` was changed. Before, it returned no changes, now it returns 1.

PR #11140, opened 2025-04-16
Fix test suites count.

### LINKED ISSUE #11138: Wrong suite count
**Describe the bug**
When measuring test suites with JUnit as source and a Junit.xml file that contains duplicate suite names, the suite count will be incorrect.

**Expected behaviour**
Use the package name to disambiguate suites with the same name.

 - Quality-time version 5.27.0

**Additional context**
Related issue: #9883.

fniessink:
When measuring test suites with JUnit XML as source, the count of test suites would be incorrect if the test suite names are not unique. Fixed by using the test suite id to disambiguate suites, if available.

Fixes #11138.

### DIFF
--- a/components/collector/tests/source_collectors/junit/test_test_suites.py
+++ b/components/collector/tests/source_collectors/junit/test_test_suites.py
@@ -23,14 +23,16 @@ class JUnitTestSuitesTest(JUnitCollectorTestCase):
         self,
         name: str,
         result: str,
+        *,
         passed: int = 0,
         errored: int = 0,
         failed: int = 0,
         skipped: int = 0,
+        key: str = "",
     ) -> dict[str, str]:
         """Create a JUnit measurement entity."""
         return {
-            "key": name,
+            "key": key or name,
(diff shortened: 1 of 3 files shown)


components/collector/tests/source_collectors/junit/test_test_suites.py
```diff
@@ -60,7 +62,7 @@ async def test_empty_test_suites(self):
     async def test_one_top_level_test_suite(self):
         """Test that a JUnit XML file with one top level test suite works."""
         xml = """
-        <testsuite name="ts1" timestamp="2009-12-19T17:58:59" failures="0" errors="1" skipped="0" tests="1">
```
wkoot, line 63: Is "skipped" no longer a valid attribute?

fniessink: It is valid, but optional, so I removed it in one or two tests.

wkoot (approved):

PR #11189, opened 2025-04-25
Don't magically create an HTML landing URL for XML sources.

### LINKED ISSUE #11136: Automatically created html URL
**Describe the bug**
Quality-time generates a .html URL that was not specified when the user configured an xml source

**To Reproduce**
1. create new metric test cases
2. add source JUnit
3. Fill in valid XML target
4. See how Quality-time automatically creates a .html URL that was never specified

**Expected behaviour**
For browser-readable sources like xml just link to the xml, else do not make the source text a clickable hyperlink, just static text.

**Screenshots**

Image

fniessink:
Fixes #11136.

### DIFF
--- a/components/collector/src/base_collectors/source_collector.py
+++ b/components/collector/src/base_collectors/source_collector.py
@@ -217,13 +217,9 @@ class SourceCollector:
     async def _landing_url(self, responses: SourceResponses) -> URL:
         """Return a user-friendly landing url.
 
-        Return the user supplied landing url parameter if there is one, otherwise translate the url parameter into
-        a default landing url.
+        Return the user supplied landing url parameter if there is one, otherwise return the API URL.
         """
-        if landing_url := self.__parameters.landing_url():
-            return landing_url
-        url = str(self.__parameters.api_url())
-        return URL(url.removesuffix("xml") + "html" if url.endswith(".xml") else url)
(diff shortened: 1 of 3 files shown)


components/collector/src/base_collectors/source_collector.py
```diff
@@ -217,13 +217,9 @@ async def __safely_parse_landing_url(self, responses: SourceResponses) -> URL:
...
         """Return a user-friendly landing url.
 
-        Return the user supplied landing url parameter if there is one, otherwise translate the url parameter into
-        a default landing url.
+        Return the user supplied landing url parameter if there is one, otherwise return the API URL.
         """
-        if landing_url := self.__parameters.landing_url():
-            return landing_url
-        url = str(self.__parameters.api_url())
-        return URL(url.removesuffix("xml") + "html" if url.endswith(".xml") else url)
```
wkoot, line 226: Hah, looks like this was introduced to parse testdata only! https://github.com/ICTU/quality-time/pull/164/files

wkoot (approved):

components/collector/src/base_collectors/source_collector.py
fniessink: I think it was to "help" the user not having to create a landing URL.

PR #11227, opened 2025-05-07
Faster opening of home page and reports

### LINKED ISSUE #11226: Quality-time slow to load many reports
**Describe the bug**
When there are many reports, opening the home page or a specific report is slow. This is caused by Quality-time having to load all reports from the database, even when opening just one, because it needs the contents of all reports in case the user wants to copy or move a subject, metric, or source.

Currently, Quality-time loads the reports one by one from the database. Fix by using a aggregation pipeline.

fniessink:
When opening the home page or a specific report don't load reports one by one, but rather all at once, for improved performance.

Fixes #11226.

### DIFF
--- a/components/api_server/src/database/reports.py
+++ b/components/api_server/src/database/reports.py
@@ -79,17 +79,22 @@ def latest_reports(database: Database) -> list[Report]:
     return cast(list[Report], get_reports(database, Report))
 
 
+type AggregationPipeline = list[dict[str, dict[str, int | str | dict[str, str]]]]
+
+
 def latest_reports_before_timestamp(database: Database, data_model: dict, max_iso_timestamp: str) -> list[Report]:
     """Return the latest, undeleted, reports in the reports collection before the max timestamp."""
-    report_filter = {"timestamp": {"$lte": max_iso_timestamp}} if max_iso_timestamp else {}
-    report_uuids = database.reports.distinct("report_uuid", report_filter)
-    report_dicts = []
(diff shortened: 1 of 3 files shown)


components/api_server/src/database/reports.py
```diff
@@ -79,17 +79,22 @@ def latest_reports(database: Database) -> list[Report]:
...
-    for report_uuid in report_uuids:
-        report_filter["report_uuid"] = report_uuid
-        report_dict = database.reports.find_one(report_filter, sort=TIMESTAMP_DESCENDING)
-        if report_dict and "deleted" not in report_dict:
-            report_dicts.append(report_dict)
-    return [Report(data_model, report_dict) for report_dict in report_dicts]
+    pipeline: AggregationPipeline = [
+        {"$sort": {"timestamp": pymongo.DESCENDING}},
+        # Group reports by report uuid and add the first report, which is the latest due to the sorting, to the group:
+        {"$group": {"_id": "$report_uuid", "latestReport": {"$first": "$$ROOT"}}},
```
wkoot, line 90: Magic, but neat! Are there other places where we can aggregate queries?

wkoot (approved):

components/api_server/src/database/reports.py
fniessink: We have used it in the past for measurements apparently (I just found some dead code in the tests, will remove it as part of this change), but we're not using it anywhere else at the moment.

PR #11246, opened 2025-05-12
Filter missing metrics by source type

### LINKED ISSUE #11109: Allow for Missing Metrics to filter on source
I'm always frustrated when the missing metrics metric reports on metric that we will never use. (Because we don't run a performance test).

It would be very useful that the missing metrics metric could be enhanced with the source.
When setting up a new Quality-time board, the user know his/her stack.
I know that I have front-end and I know we use SonarQube.
If you could configure the missing metrics metric to report only metrics that can be utilised based on that stack (that you know you have), it would report only sensible metrics that are missing.

(based on a discussion with @martijndevrieze )

fniessink:
When measuring missing metrics, allow for filtering metrics by source type.

Closes #11109.


components/collector/tests/source_collectors/quality_time/test_missing_metrics.py
```diff
@@ -165,3 +167,18 @@ async def test_subject_without_overridden_subject_name(self):
             total=str(self.nr_supported_metric_types()),
             value=str(len(entities)),
         )
+
+    async def test_source_types_to_include(self):
+        """Test that the number of non-ignored missing metrics is returned when filtered by source type."""
+        self.set_source_parameter("source_types_to_include", ["Jira"])
```
wkoot, line 173: Should we not also have test cases which deal with a combination of include and ignore, or does that functionally not make sense?

components/collector/src/source_collectors/quality_time/missing_metrics.py
```diff
@@ -25,16 +25,20 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
     async def _parse_source_responses(self, responses: SourceResponses) -> SourceMeasurement:
         """Get the metric entities from the responses."""
         datamodel_response, reports_response = responses
-        data_model = await datamodel_response.json()
+        self.data_model = await datamodel_response.json()
```
wkoot, line 28: I'm surprised that the linter doesn't demand typing defined for this class instance var

components/collector/tests/source_collectors/quality_time/test_missing_metrics.py
fniessink: The combination would be the intersection of both filters because they are applied sequentially. But it can't harm to add a test.

components/collector/src/source_collectors/quality_time/missing_metrics.py
fniessink: I assume it's because `json()` returns `Any` (https://docs.aiohttp.org/en/stable/_modules/aiohttp/client_reqrep.html#ClientResponse.json).

PR #11365, opened 2025-05-26
Bulk editing of tags.

### LINKED ISSUE #11245: Bulk removal of metric tags
As a Quality Time user I would like to be able to bulk remove a particular (custom) tag from my metrics rather than having to go through all metrics one by one to remove the tag. 

Ideally it would also be possible to bulk change a custom tag, see #11280.

### LINKED ISSUE #11280: Bulk rename of metric tags
Tags cannot be renamed.

Bulk rename of metric tags.

**Describe alternatives you've considered**
Manually removing the old tag and adding the new one.

**Additional context**
Related to: #11245

fniessink:
Closes #11245 and #11280.


components/shared_code/src/shared/model/metric.py
```diff
@@ -172,3 +172,18 @@ def source_parameter_hash(self) -> str:  # pragma: no feature-test-cover
         sources = [source for _, source in sorted(self.sources_dict.items())]  # Make hash independent of source order
         source_parameters = [source.get("parameters") for source in sources]
         return md5_hash(str(source_parameters))
+
+    def delete_tag(self, tag: str) -> bool:
+        """Delete a tag from the metric if the metric has it. Return whether the metric was affected."""
+        if tag in self.get("tags", []):
+            self["tags"].remove(tag)
```
wkoot, line 179: I don't understand how this is processed by the datastore, although there are working unittests. Isn't this a simple list in a dictionary? How does the `[].remove` call propagate to the db, or is that handled elsewhere; if so, shouldn't the delete/rename then also be implemented there?

components/shared_code/tests/shared/model/test_metric.py
```diff
@@ -182,3 +182,39 @@ def test_source_parameter_hash_does_not_depend_on_source_order(self):
...
+    def test_rename_existing_tag(self):
+        """Test renaming an existing tag."""
+        metric = self.create_metric(tags=["foo"])
+        self.assertTrue(metric.rename_tag("foo", "bar"))
+        self.assertEqual(["bar"], metric["tags"])
+
+    def test_rename_existing_tag_among_multiple_tags(self):
+        """Test renaming an existing tag among multiple tags."""
+        metric = self.create_metric(tags=["foo", "bar"])
+        self.assertTrue(metric.rename_tag("bar", "baz"))
```
wkoot, line 219: What happens when tag `"foo"` is renamed to `"bar"`; do we get the undesired state `tags = ["bar", "bar"]`?

components/shared_code/tests/shared/model/test_report.py
```diff
@@ -120,6 +120,31 @@ def test_find_without_source_and_metric_uuid(self):
         """Test that passing neither a source or a metric uuid throws an exception."""
         self.assertRaises(RuntimeError, self.report.instance_and_parents_for_uuid)
 
+    def test_delete_tag(self):
+        """Test that a tag can be deleted from all metrics in the report."""
+        self.assertEqual([REPORT_ID, SUBJECT_ID, METRIC_ID], self.report.delete_tag("tag"))
+        for metric in self.report.metrics:
+            self.assertNotIn("tag", metric.get("tags", []))
```
wkoot: Shouldn't this be `metric["tags"]`, so we don't accidentally miss a test failure?

components/api_server/src/routes/report.py
```diff
@@ -236,3 +236,24 @@ def get_report_issue_tracker_options(database: Database, report: Report):  # noq
...
     return issue_tracker.get_options().as_dict() | {"ok": True}
+
+
+@bottle.delete("/api/internal/report/<report_uuid>/tag/<tag>", permissions_required=[EDIT_REPORT_PERMISSION])
+@with_report(pass_report_uuid=False)
+def delete_tag(database: Database, report: Report, tag: str):
+    """Delete a tag from all metrics in a report."""
+    if changed_uuids := report.delete_tag(tag):
+        delta_description = f"{{user}} deleted tag '{tag}' from all metrics in report '{report.name}'."
+        return insert_new_report(database, delta_description, changed_uuids, report)
```
wkoot, line 247: If there are metrics which did not have the tag, they should not be marked as changed. Doesn't that mean these metrics will silently be deleted from the report because they're not included in the newly inserted version of the report?

tests/feature_tests/src/features/report.feature
```diff
@@ -71,6 +71,38 @@ Feature: report
...
     Then the report metric status summary is returned
 
+  Scenario: remove a tag from all metrics in a report
+    Given an existing report with title "Scenario: remove tag"
+    And an existing subject
+    And an existing metric with tags "safety, maintainability"
+    And an existing subject
+    And an existing metric with tags "security, maintainability"
+    When the client removes the tag "security" from the report
+    Then the metric tags is "maintainability"
```
wkoot, line 79: What about the other metric which has tags `"safety, maintainability"`?

tests/feature_tests/src/features/report.feature
```diff
@@ -71,6 +71,38 @@ Feature: report
...
+
+  Scenario: remove a non-existing tag from all metrics in a report
+    Given an existing report with title "Scenario: remove non-existing tag"
+    And an existing subject
+    And an existing metric with tags "maintainability"
+    When the client removes the tag "security" from the report
+    Then the metric tags is "maintainability"
```
wkoot, line 86: Should we also check for a toast message with the expected error?

components/frontend/src/report/Tags.jsx
```diff
@@ -0,0 +1,106 @@
...
+function nrMetricsWithTag(report, tag) {
+    let count = 0
+    Object.values(report.subjects).forEach((subject) => {
+        Object.values(subject.metrics).forEach((metric) => {
+            if (getMetricTags(metric).includes(tag)) {
+                count++
+            }
+        })
+    })
+    return count
```
wkoot, line 24: Don't we also need a `Tags.test.jsx` to test this?

components/frontend/src/report/Tags.jsx
```diff
@@ -0,0 +1,106 @@
...
+TagRow.propTypes = {
+    reload: func.isRequired,
+    report: reportPropType,
+    tag: string.isRequired,
+}
+
+export function Tags({ reload, report }) {
+    const tags = getReportTags(report)
+    if (tags.length === 0)
+        return <InfoMessage title="No tags">None of the metrics in this report has tags.</InfoMessage>
```
wkoot:
```suggestion
        return <InfoMessage title="No tags">None of the metrics in this report have tags.</InfoMessage>
```

components/frontend/src/report/ReportTitle.test.jsx
```diff
@@ -238,9 +242,52 @@ describe("issue tracker tab", () => {
...
 
+    it("shows the tags", async () => {
+        const { container } = renderReportTitle()
+        expect(screen.getAllByText(/Tags/).length).toBe(1)
+        await expectNoAccessibilityViolations(container)
+    })
+
+    it("shows an info message if there are no tags", async () => {
+        const { container } = renderReportTitle({ tags: [] })
+        expect(screen.getAllByText(/None of the metrics in this report has tags/).length).toBe(1)
```
wkoot:
```suggestion
        expect(screen.getAllByText(/None of the metrics in this report have tags/).length).toBe(1)
```

components/shared_code/src/shared/model/metric.py
fniessink: The API-server retrieves the report from the database, changes it in memory, and then inserts the changed report to the database. See `components/api_server/src/routes/report.py`, for example: 

```python
@bottle.delete("/api/internal/report/<report_uuid>/tag/<tag>", permissions_required=[EDIT_REPORT_PERMISSION])
@with_report(pass_report_uuid=False)
def delete_tag(database: Database, report: Report, tag: str):
    """Delete a tag from all metrics in a report."""
    if changed_uuids := report.delete_tag(tag):
        delta_description = f"{{user}} deleted tag '{tag}' from all metrics in report '{report.name}'."
        return insert_new_report(database, delta_description, changed_uuids, report)
    return {"ok": False, "error": f"Cannot find tag '{tag}' in report '{report.name}'."}
```

components/shared_code/tests/shared/model/test_metric.py
fniessink: Oh, good one! I'll need to address that.

components/shared_code/tests/shared/model/test_report.py
fniessink: Yes, you're right.

components/api_server/src/routes/report.py
fniessink: This feature does not delete metrics, only tags.

tests/feature_tests/src/features/report.feature
fniessink: That one is there for coverage :-). I'll see if I can add an assertion for that one as well.

tests/feature_tests/src/features/report.feature
fniessink: There is no toast message because the frontend should prevent errors like deleting a non-existing tag.

components/frontend/src/report/Tags.jsx
fniessink: The tests are currently in `ReportTitle.test.js`. I'll move them to `Tags.test.jsx`.

PR #11369, opened 2025-05-28
3131 allow to change the metric order by means of drag and drop

### LINKED ISSUE #3131: Allow to change the metric order by means of drag-and-drop
the order of the rows should be easily changed by the users

this can be done with a dot image at the end of a row.

see the 5th image on this webpage

https://www.uxbooth.com/articles/designing-user-friendly-data-tables/

image

Drag handle should not be visible in read mode.

RicardoDeden:


fniessink:
The drop indicator suggests that it's possible to move a metric from one subject to another (see image), but that doesn't work. Could you prevent showing a drop indicator when the drop target subject is different from the drop source subject?

image

fniessink:
On MacOS, Safari (18.5), there's some offset problem: if I drag a metric and only move the mouse vertically, the dragged metric is too far to the right. It's as if the drag handle is left of the tag instead of at the right end of the metric. I don't see this when using Firefox on MacOS.

image

fniessink:
It's possible to drag a metric all the way to the left or the right, but the reordering is only vertical. Is it possible to restrict dragging to vertical movement only?

image

fniessink:
@RicardoDeden Looks great! I do have some minor test findings, see above. Could you address these first before I look at the code?

RicardoDeden:
@fniessink thanks, I'll take a look at your findings!

RicardoDeden:
@fniessink I've fixed the drag indicator issue when dragging over other subjects than the origin. Regarding the Safari issue, I have been unable to reproduce this with the latest changes.
Finally, regarding the drag image locking, it is to my knowledge not possible to lock the dragging direction in the current configuration. Perhaps a different implementation of drag-and-drop behaviour would allow for this, but I'm not aware of one yet, and it would also mean an entirely different implementation. Also, in my opinion, this would feel too restrictive in relation to the action one is performing, and the current behaviour does not influence the actual functionality.

After the latest updates from the main branch, I have some failing tests, which I'll now have to fix. Unfortunately, I won't be able to pick up the work until September 22nd.

components/api_server/src/model/actions.py
```diff
@@ -88,3 +80,48 @@ def move_item(
...
+    items_dict: ItemsDictType,
+    item_to_move: Subject | Metric | Source,
+    new_index: int,
+) -> dict[str, dict]:
+    """Return a reordered dict with item_to_move at new_index."""
+    # Dicts are guaranteed to be (insertion) ordered starting in Python 3.7, but there's no API to change the order so
+    # we construct a new dict in the right order and insert that in the report.
+    reordered_items: dict[str, dict] = {}
+    del items_dict[item_to_move.uuid]
+    for _idx, (item_id, item) in enumerate(items_dict.items()):
```
fniessink: `_idx` is unused so why the `enumerate`?

components/frontend/src/subject/SubjectTableBody.test.jsx
```diff
@@ -0,0 +1,202 @@
...
+
+    const dropEvent = createEvent.drop(row0)
+    dropEvent.preventDefault = vi.fn()
+    fireEvent(row0, dropEvent)
+
+    // Verify no API call was made
+    expect(setMetricAttribute).not.toHaveBeenCalled()
+})
+
+it("Shows a console log if API call fails", async () => {
```
fniessink:
```suggestion
it("shows a console log if API call fails", async () => {
```

components/frontend/src/subject/SubjectTableRow.jsx
```diff
@@ -226,14 +250,35 @@ export function SubjectTableRow({
...
         <TableRowWithDetails
+            data-testid={`metric-row-${index}`}
+            ref={rowRef}
+            onDragStart={(e) => onDragStart(index, rowRef, e)}
+            onDragEnter={() => onDragEnter(index)}
+            onDrop={onDrop}
+            onDragOver={(e) => e.preventDefault()}
+            style={{
+                transition: "transform 150ms ease",
+                transform: isDropTarget ? "translateY(10px)" : "none",
```
fniessink: This creates extra white space beneath the blue separator line that appears when dragging. If I remove this transform, the blue line appears beween the metrics with additional white space. I think we can remove line 280?

fniessink (approved):
Looks good to me. Big thanks for this feature!

Just a few small comments. I've also rebased and fixed the unit tests.

Let me know if you want me to fix the last few issues or if you want to do it yourself.

RicardoDeden:
@fniessink thanks for the review, will pick up the comments today.

components/api_server/src/model/actions.py
RicardoDeden: Thanks, probably a leftover from something I tried earlier. Will refactor.

components/frontend/src/subject/SubjectTableRow.jsx
RicardoDeden: I agree, the added whitespace is redundant. I'll remove this.

RicardoDeden:
@fniessink I've updated the pull request and applied your suggestions.

fniessink:
@RicardoDeden Merged! Thanks a lot for your contribution!

PR #11551, opened 2025-06-23
Configuring Grafana k6 slow transactions

### LINKED ISSUE #11462: Add parameters to slow transactions metric for k6
I'm always frustrated when I cannot select transactions when using the metric 'slow transaction' and having as source: grafana labs k6 summary.json.

When using JMeter as a source for the same metric one gets the desired parameters as can be read here https://quality-time.readthedocs.io/en/latest/reference.html#slow-transactions-from-jmeter-csv
This allows for selecting and measuring the desired transactions from the source artefact. This, however, is not possible when using k6 as a source.

fniessink:
When measuring 'slow transactions' using Grafana k6 as source, allow for:
- filtering transactions by name,
- configuring the response time type to evaluate,
- setting the target response time and transaction-specific response times.

Closes #11462.

### DIFF
--- a/components/collector/src/source_collectors/grafana_k6/slow_transactions.py
+++ b/components/collector/src/source_collectors/grafana_k6/slow_transactions.py
@@ -2,14 +2,44 @@
 
 from typing import cast
 
-from base_collectors import JSONFileSourceCollector
+from shared_data_model import DATA_MODEL
+
+from base_collectors import JSONFileSourceCollector, SlowTransactionsCollector, TransactionEntity
 from collector_utilities.type import JSON
-from model import Entities, Entity
+from model import Entities
 
 from .base import Metric, SummaryJSON
 
 
-class GrafanaK6SlowTransactions(JSONFileSourceCollector):
+def is_default_response_time(response_time_to_evaluate: str) -> bool:
+    """Return whether the response time to evaluate is the default type."""
+    default_value = DATA_MODEL.sources["grafana_k6"].parameters["response_time_to_evaluate"].default_value
+    return response_time_to_evaluate == default_value
+
+
+class GrafanaK6TransactionEntity(TransactionEntity):
+    """Entity for a Grafana k6 transaction."""
+
+    def is_slow(
+        self,
+        response_time_to_evaluate: str,
+        target_response_time: float,
+        transaction_specific_target_response_times: list[str],
+    ) -> bool:
+        """Return whether the transaction is slow."""
+        # If the user did not change the response time to evaluate, the thresholds from the summary.json have already
+        # been used (in GrafanaK6SlowTransactions._include_metric() below) to determine that this transaction is slow,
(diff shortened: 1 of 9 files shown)


docs/src/changelog.md
```diff
@@ -12,6 +12,12 @@ If your currently installed *Quality-time* version is not the latest version, pl
 
 <!-- The line "## <square-bracket>Unreleased</square-bracket>" is replaced by the release/release.py script with the new release version and release date. -->
 
+## [Unreleased]
+
+### Added
+
+- When measuring 'slow transactions' using Grafana k6 as source, allow for filtering transactions by name, configuring the response time to evaluate, setting the target response time, and for adding transaction-specific response times. Closes [#11462](https://github.com/ICTU/quality-time/issues/11462).
```
wkoot: This might be a bit lengthy

components/collector/src/source_collectors/grafana_k6/slow_transactions.py
```diff
@@ -30,12 +60,14 @@ def _parse_json(self, json: JSON, filename: str) -> Entities:
                 max_response_time=str(values["max"]),
                 thresholds=self._format_thresholds(metric),
             )
-            entities.append(entity)
+            if self._is_to_be_included_and_is_slow(entity):
```
wkoot, line 63: Perhaps a good improvement to move the `_is_to_be_included_and_is_slow` from parse methods to `_include_entity` methods?

components/collector/src/source_collectors/grafana_k6/slow_transactions.py
```diff
@@ -2,14 +2,44 @@
...
+        target_response_time: float,
+        transaction_specific_target_response_times: list[str],
+    ) -> bool:
+        """Return whether the transaction is slow."""
+        # If the user did not change the response time to evaluate, the thresholds from the summary.json have already
+        # been used (in GrafanaK6SlowTransactions._include_metric() below) to determine that this transaction is slow,
+        # so return True in that case. Otherwise, evaluate the response time against the target response time.
+        return (
+            True
+            if is_default_response_time(response_time_to_evaluate)
```
wkoot, line 35: This check is also done in `GrafanaK6SlowTransactions._include_metric`, are they both needed for different code paths?

components/collector/src/source_collectors/grafana_k6/slow_transactions.py
fniessink: Yeah, good idea. Bit much for this PR. I'll tackle it in a separate one.

components/collector/src/source_collectors/grafana_k6/slow_transactions.py
fniessink: Yes. The complicated thing with Grafana k6 is that transactions in the the summary.json file can have an evaluation (threshold) done by Grafana k6 itself or they can have no evaluation, in which case the user can set a target response time and let Quality-time do the evaluation.

PR #11560, opened 2025-06-25
Don't show expand/collapse buttons in PDF-exports.

### LINKED ISSUE #11555: Expand/collapse buttons included in PDF
**Describe the bug**
The PDF-exports show the expand/collapse buttons, but they are not functional in the PDF.

**To Reproduce**
Export any report to PDF.

**Expected behaviour**
Hide the expand/collapse buttons from the PDF. Make sure the text following the buttons is moved to the left slightly.

- Quality-time version 5.33.0

fniessink:
Fixes #11555.

### DIFF
--- a/components/frontend/src/widgets/TableRowWithDetails.jsx
+++ b/components/frontend/src/widgets/TableRowWithDetails.jsx
@@ -1,11 +1,20 @@
-import { TableCell, TableRow } from "@mui/material"
+import { Stack, TableCell, TableRow } from "@mui/material"
 import { bool, func, string } from "prop-types"
+import React from "react"
 
 import { childrenPropType } from "../sharedPropTypes"
 import { ExpandButton } from "./buttons/ExpandButton"
 
 export function TableRowWithDetails(props) {
     const { color, children, details, expanded, onExpand, ...otherProps } = props
+    const [firstTableCell, ...otherTableCells] = React.Children.toArray(children)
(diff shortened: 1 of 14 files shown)


components/frontend/src/widgets/TableRowWithDetails.jsx
```diff
@@ -1,11 +1,20 @@
...
 import { bool, func, string } from "prop-types"
+import React from "react"
 
 import { childrenPropType } from "../sharedPropTypes"
 import { ExpandButton } from "./buttons/ExpandButton"
 
 export function TableRowWithDetails(props) {
     const { color, children, details, expanded, onExpand, ...otherProps } = props
+    const [firstTableCell, ...otherTableCells] = React.Children.toArray(children)
+    const firstTableCellWithExpandButton = React.cloneElement(
```
wkoot, line 11: Why clone? That sounds a bit unexpected

fniessink: We want to insert the expand/collapse button into the contents of the first table cell. So if the `TableRowWithDetails` children props look like this:

```jsx
<TableCell>first column</TableCell>
<TableCell>second column</TableCell>
```

Then the first cell gets cloned, and an expand/collapse button is inserted in the clone so we end up with (simplified):

```jsx
<TableCell><ExpandButton/>first column</TableCell>
<TableCell>second column</TableCell>
```

PR #11605, opened 2025-07-01
Filtering missing metrics by subject

### LINKED ISSUE #11106: Allow for including subjects in missing metrics metric
I'm always frustrated when configuring the missing metrics metrics and in the source tab I need to exclude subjects.

Allow for users to include subjects instead of excluding what you don't want to see.
Additionally, provide a drop-down menu for selecting the subject that you want to measure.

fniessink:
When measuring missing metrics, allow for configuring 'subjects to include' to only count missing metrics in the configured subjects.

Closes #11106.

### DIFF
--- a/components/collector/src/source_collectors/quality_time/missing_metrics.py
+++ b/components/collector/src/source_collectors/quality_time/missing_metrics.py
@@ -37,15 +37,21 @@ class QualityTimeMissingMetrics(QualityTimeCollector):
         """Return whether to include the entity in the measurement."""
         source_types_to_include = set(self._parameter("source_types_to_include"))
         metric_source_types = set(self.data_model["metrics"][entity["metric_type"]]["sources"])
-        if source_types_to_include and not (source_types_to_include & metric_source_types):
+        if not (source_types_to_include & metric_source_types):
             return False
-        subjects_to_ignore = self._parameter("subjects_to_ignore")
-        if not subjects_to_ignore:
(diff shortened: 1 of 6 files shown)


components/collector/src/source_collectors/quality_time/missing_metrics.py
```diff
@@ -37,15 +37,21 @@ def _include_entity(self, entity: Entity) -> bool:
...
-            subjects_to_ignore,
-        ) and not match_string_or_regular_expression(entity["subject_uuid"], subjects_to_ignore)
+
+        if subjects_to_include := self._parameter("subjects_to_include"):
+            subject_name_matches = match_string_or_regular_expression(entity["subject"], subjects_to_include)
+            subject_uuid_matches = match_string_or_regular_expression(entity["subject_uuid"], subjects_to_include)
+            if not (subject_name_matches or subject_uuid_matches):
+                return False
+
+        if subjects_to_ignore := self._parameter("subjects_to_ignore"):
```
wkoot, line 49: Is it possible for the partial name of any subject to be included within another subject? I.e. that it would be included and then also ignored due to partial regex match?

fniessink: Yes, that is possible. For example, if you have "Subject 1" and Subject 12" and then include `Subject \d+` and ignore `.*2`.

PR #11606, opened 2025-07-01
Ignore missing metrics that need unused source types

### LINKED ISSUE #11604: Allow for ignoring missing metric types based on supporting source type
The missing metric reports metric types as metrics for source types I don't have in my project. For example, it reports the "performancetest duration" metric as missing, with "manual number" and "performancetestrunner" as supporting sources.

Allow for ignoring source types. If all of the supporting source types of a metric type are ignored, don't report the metric type as missing.

fniessink:
When measuring missing metrics, allow for configuring 'source types to ignore' so that metrics that need source types that are not available are not reported.

Closes #11604.


components/shared_code/src/shared_data_model/parameters.py
```diff
@@ -61,14 +61,40 @@ class SingleChoiceParameter(Parameter):
...
+
+    @model_validator(mode="after")
+    def set_default_value(self) -> Self:
+        """Make the default value equal to all values, if it was not set explicitly."""
+        if not self.default_value and self.values:
+            self.default_value = self.values
+        return self
+
+
+class MultipleChoiceParameterWithoutDefaults(Parameter):
```
wkoot: It seems more consistent to name this `MultipleChoiceWithoutDefaultsParameter`

components/shared_code/src/shared_data_model/parameters.py
```diff
@@ -61,14 +61,40 @@ class SingleChoiceParameter(Parameter):
     type: ParameterType = ParameterType.SINGLE_CHOICE
 
 
-class MultipleChoiceParameter(Parameter):
-    """Multiple choice parameter."""
+class MultipleChoiceParameterWithDefaults(Parameter):
```
wkoot: It seems more consistent to name this `MultipleChoiceWithDefaultsParameter`

components/collector/src/model/parameters.py
```diff
@@ -43,15 +43,10 @@ def quote_if_needed(parameter_value: str) -> str:
             return urllib.parse.quote(parameter_value, safe="") if quote else parameter_value
 
         parameter_info = DATA_MODEL.sources[self.__source["type"]].parameters[parameter_key]
-        if parameter_info.type == "multiple_choice":
-            # If the user didn't pick any values, select the default value if any, otherwise select all values:
-            default_value = parameter_info.default_value
-            value = self.__parameters.get(parameter_key) or default_value or parameter_info.values or []
+        value = self.__parameters.get(parameter_key) or parameter_info.default_value
```
wkoot, line 46: Hmm, this is slightly unexpected; I guess the case `multiple_choice_without_defaults` would automagically be part of `or parameter_info.default_value` ?

fniessink: Not sure what you find unexpected? In any case, this code is now simpler because the datamodel itself makes sure that the default value of `MultipleChoiceParameterWithDefaults` is filled with all values if no default values are provided at instantiation.

PR #11618, opened 2025-07-03
Separate date settings under the report date

### LINKED ISSUE #11616: Move date settings to a separate date menu
Replace the report date picker with a date menu (panel) that contains the report date picker and the settings for number of dates, time between dates, and date order.

Nest the date picker in the panel using a `StaticDatePicker`. See https://mui.com/x/react-date-pickers/date-picker/#available-components.

The menu bar keeps showing the report date.

Changes in the date menu panel are applied immediately, like we do for changes made in the settings panel.

Prepares for #11567.

fniessink:
Replace the report date picker with a panel that contains the report date picker and the settings for number of dates, time between dates, and date order.

Closes #11616.


components/frontend/src/header_footer/Menubar.jsx
```diff
@@ -29,19 +30,7 @@ export function Menubar({
...
-            if (event.key === "Escape") {
-                setSettingsPanelVisible(false)
-            }
-        }
-        window.addEventListener("keydown", closePanels)
-        return () => {
-            window.removeEventListener("keydown", closePanels)
-        }
-    }, [])
-
```
fniessink, line 44: Turns out this is superfluous. The Drawer component listens to Escape itself.

components/frontend/src/header_footer/Menubar.jsx
```diff
@@ -29,19 +30,7 @@ export function Menubar({
...
+    const [visibleSettingsPanel, setVisibleSettingsPanel] = useState("")
```
fniessink, line 36: Make the state a string instead of a boolean so we can keep track of which panel is actually being shown.

components/frontend/src/header_footer/Menubar.jsx
```diff
@@ -55,12 +44,14 @@ export function Menubar({
                     <Stack direction="row" spacing={2} sx={{ flexGrow: 1 }}>
                         <HomeButton
                             atReportsOverview={atReportsOverview}
-                            openReportsOverview={openReportsOverview}
-                            setSettingsPanelVisible={setSettingsPanelVisible}
+                            openReportsOverview={() => {
+                                setVisibleSettingsPanel("")
+                                openReportsOverview()
```
wkoot, line 52: Github is highlighting this line differently than the one before, is it expecting a semicolon to end the previous line or so?

components/frontend/src/header_footer/Menubar.jsx
```diff
@@ -72,19 +63,29 @@ export function Menubar({
...
+                            <UserButton email={email} user={user} setUser={setUser} />
                         )}
                     </Stack>
                 </Toolbar>
             </AppBar>
-            <Drawer anchor="top" open={settingsPanelVisible} onClose={() => setSettingsPanelVisible(false)}>
+            <Drawer anchor="top" open={visibleSettingsPanel} onClose={() => setVisibleSettingsPanel("")}>
                 <Toolbar /* Add an empty toolbar to the drawer so the panel is not partly hidden by the appbar. */ />
-                {panel}
+                {visibleSettingsPanel === "settings" ? (
```
wkoot, line 87: Why is this only the case for the Settings panel and not for the Period panel?

components/frontend/src/header_footer/buttons/ReportPeriodButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+import { Stack } from "@mui/material"
+import { bool, func } from "prop-types"
+
+import { datePropType } from "../../sharedPropTypes"
+import { AppBarButton } from "./AppBarbutton"
+
+export function ReportPeriodButton({ reportDate, setSettingsPanelVisible, settingsPanelVisible }) {
+    return (
+        <AppBarButton
+            tooltip="Change report date, number of dates shown, time between dates, and date order"
```
wkoot, line 13: Do we also write an exhaustive list for the settings panel tooltip?

components/frontend/src/header_footer/buttons/ReportPeriodButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+            startIcon={
+                <Stack direction="row" spacing={0}>
+                    {settingsPanelVisible ? <ArrowDropDownIcon /> : <ArrowRightIcon />}
+                    <EventIcon />
+                </Stack>
+            }
+            onClick={() => setSettingsPanelVisible(!settingsPanelVisible)}
```
wkoot: How does the click know that it's opening the period panel and not the settings panel; shouldn't it also pass `visibleSettingsPanel` ?

components/frontend/src/header_footer/buttons/ReportPeriodButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+        >
+            {reportDate ? reportDate.toLocaleDateString() : "today"}
+        </AppBarButton>
+    )
+}
+ReportPeriodButton.propTypes = {
+    reportDate: datePropType,
+    setSettingsPanelVisible: func,
+    settingsPanelVisible: bool,
```
wkoot: Wasn't this now the string `visibleSettingsPanel`?

components/frontend/src/header_footer/Menubar.jsx
fniessink: I think it's because `openReportsOverview` is a prop and `setVisibleSettingsPanel` is a local variable.

components/frontend/src/header_footer/Menubar.jsx
fniessink: `visibleSettingsPanel == "settings"` if the `Drawer` is open with the settings panel in it, `visibleSettingsPanel == "reportperiod"` if the `Drawer` is open with the report period panel in it, and visibleSettingsPanel == ""` if the `Drawer` is closed. I'll add this to a comment.

components/frontend/src/header_footer/buttons/ReportPeriodButton.jsx
fniessink: No, good point :-)

components/frontend/src/header_footer/buttons/ReportPeriodButton.jsx
fniessink: The buttons don't need to know there are multiple panels, they just turn their own panel on and off (information hiding FTW). The click handlers in the menu bar translate the boolean into the correct panel, e.g.: `(visible) => setVisibleSettingsPanel(visible ? "settings" : "")`

components/frontend/src/header_footer/buttons/ReportPeriodButton.jsx
fniessink: No, see above. The buttons only need to know whether their own panel is visible or not (hence a boolean). The menubar keeps track of which panel is visible, hence a string there.

PR #11671, opened 2025-07-11
Use uv.lock files instead of requirements.txt

fniessink:
- Replace` requirements.txt` files with `uv.lock` files.
- Remove `ci/pip-compile.sh` scripts.
- Replace `ci/pip-install.sh` scripts with `ci/install-dependencies.sh` (and add one for the frontend component).
- Update the Dependabot config to use `uv` instead of `pip` as package ecosystem.
- Bump `uv` to `0.7.20`.
- Make the `shared_code` component a package that can be installed. This removes the need to copy the `shared_code` source code into the Python backend docker images. This also means the logos are now a "resource". Where before we simply served the code from the filesystem, we now use the `importlib.resources` module.
- Install the dependencies in the Python backend docker images using an ephemeral mount of `uv`. This removes the need for a two-stage build of the images and results in slightly smaller images (~7 MB less).


docs/ci/install-dependencies.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+npm install --ignore-scripts
+uv sync --active --locked --extra dev
```
wkoot: Why do we only use `--active` here and nowhere else?

.readthedocs.yaml
```diff
@@ -12,6 +12,13 @@ build:
     python: "3.13"
   apt_packages:
     - graphviz
+  jobs:
+    pre_create_environment:
+      - asdf plugin add uv
```
wkoot: Is it possible to pin versions here? It looks like the build is currently failing on master https://github.com/asdf-community/asdf-uv?tab=readme-ov-file

ci/quality-base.sh
```diff
@@ -37,9 +37,7 @@ run_bandit() {
 }
 
 run_pip_audit() {
-    for requirements_file in $(python $(script_dir)/requirements_files.py); do
-        run_uvx pip-audit --disable-pip --progress-spinner=off --requirement $requirements_file --ignore-vuln GHSA-48p4-8xcf-vxj5 --ignore-vuln GHSA-pq67-6m6q-mj2v
-    done
+    run_uvx pip-audit --requirement <(uv export --format requirements-txt --no-emit-package shared-code) --disable-pip
```
wkoot, line 40: Is it not possible to get the requirements from `pyproject.toml` directly? The `pip-audit` docs say that _only pyproject.toml and pylock.*.toml are supported_

components/shared_code/pyproject.toml
```diff
@@ -36,6 +36,9 @@ where = [
     "src",
 ]
 
+[tool.setuptools.package-data]
+"*" = [ "*.png" ]
```
wkoot, line 40: Interesting, so the resulting `MANIFEST.md` only lists png files? Do we not need the python files since we're using the package dependency?

components/api_server/uv.lock
```diff
@@ -0,0 +1,1161 @@
+version = 1
```
wkoot, line 1: Wow, these lockfiles are huge.. I'll just assume that they're okay :( Is there a good way to know that they don't contain something unexpected, i.e. poisoned pipeline execution / supply chain attack?

components/api_server/Dockerfile
```diff
@@ -1,28 +1,22 @@
...
+ENV UV_COMPILE_BYTECODE=1 UV_NO_CACHE=1 UV_PYTHON_DOWNLOADS=0 VIRTUAL_ENV=/home/server/venv
+
 WORKDIR /home/server
+
+RUN --mount=from=ghcr.io/astral-sh/uv,source=/uv,target=/bin/uv \
+    --mount=type=bind,source=shared_code,target=/home/shared_code,rw \
+    --mount=type=bind,source=api_server/uv.lock,target=/home/server/uv.lock \
+    --mount=type=bind,source=api_server/pyproject.toml,target=/home/server/pyproject.toml \
+    uv sync --active --locked --no-dev --no-install-project
+
```
wkoot, line 15: Does passing `--active` here not override `VIRTUAL_ENV=/home/server/venv`, or does it explicitly honour it?

components/shared_code/src/shared_data_model/logos/__init__.py
```diff
@@ -1,5 +1,14 @@
...
+"""Logos for sources."""
 
-import pathlib
+import importlib.resources
 
-LOGOS_ROOT = pathlib.Path(__file__).parent.absolute()
+
+def get_logo(name: str) -> bytes:
+    """Return the logo."""
+    return importlib.resources.read_binary("shared_data_model.logos", f"{name}.png")
```
wkoot, line 8: Why not `("shared_data_model", "logos", f"{name}.png")`; shouldn't it be a `pathlib.Path`?

components/shared_code/src/shared_data_model/logos/__init__.py
```diff
@@ -1,5 +1,14 @@
...
+
+
+def get_logo_names() -> list[str]:
```
wkoot, line 11: Perhaps this should be memoized in some const like `LOGO_NAMES`?

components/shared_code/src/shared_data_model/meta/data_model.py
```diff
@@ -55,15 +56,15 @@ def check_source_urls(self) -> None:
...
         for source_type in self.sources:
-            logo_path = logos_path / f"{source_type}.png"
-            if not logo_path.exists():
-                msg = f"No logo exists for {source_type}"
-                raise ValueError(msg)
-        for logo_path in logos_path.glob("*.png"):
-            if logo_path.stem not in self.sources:
-                msg = f"No source exists for {logo_path}"
+            try:
+                get_logo(source_type)
```
wkoot: Reading the binary seems like a rather expensive operation; if we only want to check that the file exists, the entry in `get_logo_names()` should suffice?

components/api_server/src/routes/logo.py
```diff
@@ -1,11 +1,16 @@
 """Source logos."""
 
 import bottle
+from bottle import abort, response
```
wkoot: When also importing `get` then the line above can be dropped

components/api_server/src/routes/logo.py
```diff
@@ -1,11 +1,16 @@
...
 
 
 @bottle.get("/api/internal/logo/<source_type>", authentication_required=False)
 def get_logo(source_type: str):
     """Return the logo for the source type."""
-    return bottle.static_file(f"{source_type}.png", root=LOGOS_ROOT, mimetype="image/png")
+    response.set_header("Content-Type", "image/png")
+    if source_type not in DATA_MODEL.sources:  # pragma: no feature-test-cover
+        return abort(404, f"Logo not found: '{source_type}' is not a valid source type")
+    return get_logo_png(source_type)
```
wkoot, line 15: Does this also return the `mimetype="image/png"`?

components/api_server/src/routes/logo.py
```diff
@@ -1,11 +1,16 @@
...
+from shared_data_model.logos import get_logo as get_logo_png
 
 
 @bottle.get("/api/internal/logo/<source_type>", authentication_required=False)
 def get_logo(source_type: str):
     """Return the logo for the source type."""
-    return bottle.static_file(f"{source_type}.png", root=LOGOS_ROOT, mimetype="image/png")
+    response.set_header("Content-Type", "image/png")
+    if source_type not in DATA_MODEL.sources:  # pragma: no feature-test-cover
+        return abort(404, f"Logo not found: '{source_type}' is not a valid source type")
```
wkoot, line 13: Does this not return with a `Content-Type=image/png` header?

components/api_server/tests/routes/test_logo.py
```diff
@@ -1,16 +1,21 @@
...
-    def test_logo(self, mock_static_file):
+    @patch("importlib.resources.read_binary", Mock(return_value="logo"))
+    def test_logo(self):
         """Test that a logo can be retrieved."""
-        mock_static_file.return_value = "logo"
         self.assertEqual("logo", get_logo("sonarqube"))
+
+    def test_missing_logo(self):
+        """Test that retrieving a missing logo results in a 404."""
+        self.assertRaises(bottle.HTTPError, get_logo, "missing")
```
wkoot, line 22: Don't we want to listen for `Logo not found: '{source_type}' is not a valid source type` here as well?

components/api_server/tests/routes/test_logo.py
```diff
@@ -1,16 +1,21 @@
...
 
 class LogoTest(unittest.TestCase):
     """Unit tests for the logo route."""
 
-    @patch("bottle.static_file")
-    def test_logo(self, mock_static_file):
+    @patch("importlib.resources.read_binary", Mock(return_value="logo"))
+    def test_logo(self):
         """Test that a logo can be retrieved."""
-        mock_static_file.return_value = "logo"
```
wkoot, line 15: Can we also assert Content-Type and/or mimetype here?

docs/ci/install-dependencies.sh
fniessink: ~Forgot to remove it, I guess. It's superfluous, I'll remove it.~

Nope. It's needed to use the prepared virtual env on ReadTheDocs. Re-added with a comment.

.readthedocs.yaml
fniessink: asdf doesn't support this itself: https://github.com/asdf-vm/asdf/issues/166. There's a plugin manager, but that itself is also a plugin...: https://github.com/asdf-community/asdf-plugin-manager. I'll install `uv` with `pip` so it can be pinned.

ci/quality-base.sh
fniessink: Yes, but that would only check the direct dependencies. This also check the indirect deps.

components/shared_code/pyproject.toml
fniessink: This is only about the package data files. Sources are specified under the `tool.setuptools.packages.find` header just above.

components/api_server/uv.lock
fniessink: I typically collect the daily Dependabot updates in a branch and review the changes before merging. But this is a change too big to review completely yes.

components/api_server/Dockerfile
fniessink: The latter, it means to use it.

components/shared_code/src/shared_data_model/logos/__init__.py
fniessink: No, it can be a module name, see https://docs.python.org/3/library/importlib.resources.html#importlib.resources.Anchor.

components/shared_code/src/shared_data_model/logos/__init__.py
fniessink: It's only used once per run, when the data model is instantiated.

components/shared_code/src/shared_data_model/meta/data_model.py
fniessink: Yes, changed.

components/api_server/src/routes/logo.py
fniessink: Done.

components/api_server/src/routes/logo.py
fniessink: Yes. Bottle adds the returned value to the response.

components/api_server/src/routes/logo.py
fniessink: Yes. Fixed.

components/api_server/tests/routes/test_logo.py
fniessink: No, because that error is checked when validating the data model and discovered when running unit test. The most likely scenario is adding a new source type but forgetting to add a logo for it. So that should never happen in production.

components/api_server/tests/routes/test_logo.py
fniessink: Yes, added.

PR #11694, opened 2025-07-16
Fix dataformat difference between Jira cloud and server version

### LINKED ISSUE #11672: Jira V3 API - custom field sprint has a dictionary instead of a string
**Describe the bug**
Custom field with the name "Sprint" cannot be parsed in the new Jira REST API response.

When the custom field named "Sprint" is parsed, an error is raised with the following stacktrace: 
```
Parse error

  File "/home/collector/base_collectors/source_collector.py", line 152, in __safely_parse_source_responses
    return await self._parse_source_responses(responses)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/collector/source_collectors/jira/issues.py", line 67, in _parse_source_responses
    entities = Entities(self._create_entity(issue, url) for issue in issues if self._include_issue(issue))
  File "/home/collector/model/entity.py", line 34, in __init__
    self.extend(entities)
    ~~~~~~~~~~~^^^^^^^^^^
  File "/home/collector/model/entity.py", line 52, in extend
    for entity in entities:
                  ^^^^^^^^
  File "/home/collector/source_collectors/jira/issues.py", line 67, in <genexpr>
    entities = Entities(self._create_entity(issue, url) for issue in issues if self._include_issue(issue))
                        ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/home/collector/source_collectors/jira/issues.py", line 103, in _create_entity
    entity_attributes["sprint"] = self.__get_sprint_names(fields.get(sprint_field_id) or [])
                                  ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/collector/source_collectors/jira/issues.py", line 118, in __get_sprint_names
    matches = [cls.SPRINT_NAME_RE.search(sprint_text) for sprint_text in sprint_texts]
               ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'dict'
```

This is because of the following change in the Jira REST API:
>From 1 September 2020, Jira REST API responses that include the Sprint custom field type will no longer return string representations of sprints in the body. This data will still be available in its proper JSON representation wherever a string was previously expected.

src: https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-tostring-representation-of-sprints-in-get-issue-response/#deprecation-notice---tostring-representation-of-sprints-in-get-issue-response

This means that the value of `sprint_texts` could be a str or a dict de pending on the Jira REST API version:
- In an older instance of Jira
  ```
  sprint_texts = [
    "com.atlassian.greenhopper.service.sprint.Sprint@1bf75fd[id=1,rapidViewId=1,state=CLOSED,name=Sprint 1
  ,startDate=2016-06-06T21:30:53.537+10:00,endDate=2016-06-20T21:30:00.000+10:00,
  completeDate=2016-06-06T21:30:57.523+10:00,sequence=1]",
      "com.atlassian.greenhopper.service.sprint.Sprint@1689feb[id=2,rapidViewId=1,state=FUTURE,name=Sprint 2
  ,startDate=<null>,endDate=<null>,completeDate=<null>,sequence=2]"
  ]
  ```
 - in an newer instance of Jira
```
sprint_texts = [
    {
        "id": 1,
        "name": "Sprint 1",
        "state": "closed",
        "boardId": 1,
        "goal": "Sprint Goal",
        "startDate": "2016-06-06T21:30:53.537+10:00",
        "endDate": "2016-06-20T21:30:00.000+10:00",
        "completeDate": "2016-06-06T21:30:57.523+10:00"
    },
    {
        "id": 2,
        "name": "Sprint 2",
        "state": "future",
        "boardId": 1
    }
]
```

**Expected behaviour**
It would be nice that the new format of the custom field named `sprint` is supported

wkoot:
Closes #11672

### DIFF
--- a/components/collector/tests/source_collectors/jira/test_user_story_points.py
+++ b/components/collector/tests/source_collectors/jira/test_user_story_points.py
@@ -1,5 +1,7 @@
 """Unit tests for the Jira user story points collector."""
 
+from model import MetricMeasurement
+
 from .base import JiraTestCase
 
 
@@ -7,6 +9,11 @@ class JiraUserStoryPointsTest(JiraTestCase):
     """Unit tests for the Jira story points collector."""
 
     METRIC_TYPE = "user_story_points"
+
+    async def _get_fields_response(self, user_stories_json: dict) -> MetricMeasurement | None | tuple:
+        """Produce sprint custom field response."""
+        fields_json = [{"name": "Field", "id": "field"}, {"name": "Sprint", "id": "custom_field_001"}]
+        return await self.get_response(user_stories_json, fields_json)
 
     async def test_nr_story_points(self):
         """Test that the number of story points is returned."""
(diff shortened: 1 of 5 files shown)


components/collector/tests/source_collectors/jira/test_user_story_points.py
```diff
@@ -36,3 +42,63 @@ async def test_nr_story_points_with_sprint_field(self):
...
+        self.assert_measurement(
+            response,
+            value="42",
+            entities=[
+                self.entity(key="1", points="10.0", sprint=""),
+                self.entity(key="2", points="32.0", sprint=""),
+            ],
+        )
+
+    async def test_nr_story_points_with_mixed_sprints(self):
```
fniessink, line 82: This test seems superflouous? Why would one Jira instance return both formats?

fniessink (approved):
I'm confused by what Atlassian did here: they changed the API for sprints, but didn't change the API version number, and they also didn't make sure Jira Server and Cloud behave the same even if the API version is the same? Grmbl!

components/collector/tests/source_collectors/jira/test_user_story_points.py
wkoot: Yes I agree, it's been very tedious to make the tests but I got a lot of complaints about coverage and typing/formatting

wkoot:
> I'm confused by what Atlassian did here: they changed the API for sprints, but didn't change the API version number, and they also didn't make sure Jira Server and Cloud behave the same even if the API version is the same? Grmbl!

What I gather from the docs is that they already implemented this change in the v2 api, but only on Jira cloud! It has therefore been brought over to the v3 as well, but apparently not backported to Jira server :(
Even better; the Jira cloud ticket advises users to contact Atlassian Support regarding Jira server in https://ecosystem.atlassian.net/browse/ACJIRA-1082?focusedCommentId=393932

PR #11696, opened 2025-07-17
Python 3.14

fniessink:


wkoot (approved):

ci/quality-base.sh
```diff
@@ -33,7 +33,10 @@ run_pyproject_fmt() {
 }
 
 run_bandit() {
-    run_uvx bandit --configfile pyproject.toml --quiet --recursive $(python_files_and_folders)
+    # Bandit doesn't support Python 3.14 yet.
```
wkoot, line 36: Do we have a ticket on our own backlog to bump Bandit when upstream is fixed?

components/shared_code/tests/shared/utils/test_log.py
```diff
@@ -1,10 +1,13 @@
 """Unit tests for the logging utilities."""
 
-import logging
 import unittest
+from typing import TYPE_CHECKING
 
 from shared.utils.log import get_logger
 
+if TYPE_CHECKING:
+    import logging
```
wkoot: Might as well `from logging import Logger`

components/shared_code/tests/shared/initialization/test_database.py
```diff
@@ -1,13 +1,17 @@
...
+from typing import TYPE_CHECKING
 from unittest.mock import Mock, patch
 
 import mongomock
-import pymongo
 
 from shared.initialization.database import get_database, mongo_client
 
+if TYPE_CHECKING:
+    import pymongo
```
wkoot: Here `from pymongo import MongoClient` should work

components/shared_code/src/shared/model/subject.py
```diff
@@ -4,17 +4,17 @@
...
 if TYPE_CHECKING:
+    from .measurement import Measurement
     from .report import Report
 
 
 class Subject(dict):  # noqa: PLW1641
     """Class representing a subject."""
 
-    def __init__(self, data_model: dict, subject_data: dict, subject_uuid: SubjectId, report: "Report") -> None:
+    def __init__(self, data_model: dict, subject_data: dict, subject_uuid: SubjectId, report: Report) -> None:
```
wkoot, line 17: Hmm this is a good find. Surprising that this didn't cause errors before

components/collector/src/source_collectors/visual_studio_trx/tests.py
```diff
@@ -1,15 +1,18 @@
...
+from typing import TYPE_CHECKING, cast
 
 from base_collectors import XMLFileSourceCollector
 from collector_utilities.functions import parse_source_response_xml_with_namespace
-from collector_utilities.type import ElementMap, Namespaces
 from metric_collectors.test_cases import TestCases
 from model import Entities, Entity, SourceMeasurement, SourceResponses
 
+if TYPE_CHECKING:
+    from xml.etree.ElementTree import Element  # nosec # Element is not available from defusedxml, but only used as type
```
wkoot, line 12: Are the comment still relevant now that it's explicitly inside type checking block anyway?

components/collector/src/source_collectors/cxsast/base.py
```diff
@@ -1,15 +1,17 @@
...
+from typing import TYPE_CHECKING, cast
 
 from shared.utils.functions import first
 
 from base_collectors import SourceCollector
 from collector_utilities.type import URL, Response
-from model import SourceResponses
+
+if TYPE_CHECKING:
+    import aiohttp
```
wkoot: Does `from aiohttp import ClientResponse` cover it?

components/api_server/src/initialization/migrations.py
```diff
@@ -1,15 +1,18 @@
...
+from typing import TYPE_CHECKING
 
 from shared.model.metric import Metric
 
 from utils.log import get_logger
 
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    from pymongo.database import Database
```
wkoot, line 13: Shouldn't this be auto checked on alphabetic sort?

components/api_server/src/routes/plugins/injection_plugin.py
```diff
@@ -1,9 +1,12 @@
 """Route value injection plugin."""
 
 import inspect
-from collections.abc import Callable
+from annotationlib import Format
+from typing import TYPE_CHECKING
 
-from bottle import Bottle, Route
+if TYPE_CHECKING:
+    from bottle import Bottle, Route
```
wkoot: Are the external packages not checked on separately sorted?

ci/quality-base.sh
fniessink: We do now :-) https://github.com/ICTU/quality-time/issues/12125

components/shared_code/src/shared/model/subject.py
fniessink: I'm not sure why Report was in quotes, but I think you can put all types in quotes if you want. So it wasn't a bug. I guess it's not needed in Python 3.14 anymore, so ruff removed the quotes.

components/collector/src/source_collectors/visual_studio_trx/tests.py
fniessink: The comment is to explain the `nosec` and keep the `nosec` is to keep bandit happy.

components/api_server/src/initialization/migrations.py
fniessink: I've tried to give it same ordering as the normal imports: stdlib, deps, shared modules, component modules, local modules. Unfortunately, `ruff format` does not format the imports under the `if TYPE_CHECKING:`..

components/api_server/src/routes/plugins/injection_plugin.py
fniessink: See previous comment.

PR #11931, opened 2025-09-10
When copying a metric also copy entity status

### LINKED ISSUE #7403: When copying a metric, technical debt acceptance per entity is not copied.
**Describe the bug**
When copying a metric that has entities and has one or more entities with a technical debt status and/or comment, this information is not copied along. This is unexpected behaviour, because all else, like source configuration and such, will be copied when copying a metric.

**To Reproduce**
1. Go to a metric that has entities and has one or more entities with a technical debt status and/or comment
2. After logging in, duplicate that metric using the copy button on the bottom of that same subject
3. Wait for the metric to gather the results
4. See that the number at the measurement field is higher in the original metric compared to the copied metric

**Expected behaviour**
When copying, everything should be copied, all config.

fniessink:
Fixes #7403.

### DIFF
--- a/components/collector/tests/database/test_measurements.py
+++ b/components/collector/tests/database/test_measurements.py
@@ -6,7 +6,7 @@ import mongomock
 
 from database.measurements import create_measurement
 
-from tests.fixtures import METRIC_ID, REPORT_ID, SOURCE_ID, SUBJECT_ID, create_report
+from tests.fixtures import METRIC_ID, METRIC_ID2, REPORT_ID, SOURCE_ID, SOURCE_ID2, SUBJECT_ID, create_report
 
 
 class TestMeasurements(unittest.TestCase):
@@ -17,7 +17,13 @@ class TestMeasurements(unittest.TestCase):
         self.client: mongomock.MongoClient = mongomock.MongoClient()
         self.database = self.client["quality_time_db"]
 
-    def measurement_data(self, first_seen: str = "2023-07-18", source_parameter_hash: str = "hash") -> dict:
+    def measurement_data(
+        self,
+        first_seen: str = "2023-07-18",
+        source_parameter_hash: str = "hash",
+        metric_uuid: str = METRIC_ID,
+        source_uuid: str = SOURCE_ID,
+    ) -> dict:
         """Create the measurement data."""
         return {
             "start": "2023-07-19T16:50:47+00:000",
@@ -26,17 +32,17 @@ class TestMeasurements(unittest.TestCase):
             "sources": [
                 {
                     "type": "sonarqube",
-                    "source_uuid": SOURCE_ID,
+                    "source_uuid": source_uuid,
                     "name": "Source",
                     "parameters": {"url": "https://url", "password": "password"},
                     "parse_error": None,
                     "connection_error": None,
                     "value": "10",
                     "total": "100",
-                    "entities": [{"key": "key", "first_seen": first_seen}],
+                    "entities": [{"key": "1", "first_seen": first_seen}],
                 },
             ],
-            "metric_uuid": METRIC_ID,
+            "metric_uuid": metric_uuid,
             "report_uuid": REPORT_ID,
             "source_parameter_hash": source_parameter_hash,
         }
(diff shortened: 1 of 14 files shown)


components/frontend/src/source/SourceEntities.jsx
```diff
@@ -235,7 +235,7 @@ export function SourceEntities({ loading, measurements, metric, metricUuid, relo
     }
     const lastMeasurement = measurements[measurements.length - 1]
     const source = lastMeasurement.sources.find((source) => source.source_uuid === sourceUuid)
-    if (!Array.isArray(source.entities) || source.entities.length === 0) {
+    if (!Array.isArray(source?.entities) || source.entities.length === 0) {
```
wkoot, line 238: Additional bugfix?

components/shared_code/tests/shared/model/test_measurement.py
```diff
@@ -282,6 +302,31 @@ def test_copy_first_seen_timestamps(self):
...
+                {
+                    "source_uuid": SOURCE_ID2,
+                    "type": "azure_devops",
+                    "entities": [{"key": "key1", "first_seen": "2023-07-17"}],
+                },
+            ],
+        )
+        measurement_2.copy_entity_first_seen_timestamps(measurement_1)
+        self.assertEqual("2023-07-15", measurement_2["sources"][0]["entities"][0]["first_seen"])
+
```
wkoot, line 329: Missing changelog message

components/shared_code/tests/shared/model/test_measurement.py
```diff
@@ -251,6 +251,26 @@ def test_entity_user_data(self):
...
+                },
+            ],
+        )
+        measurement_2 = Measurement(
+            self.metric(sources={SOURCE_ID2: cast(Source, {"copied_from": SOURCE_ID})}),
+            sources=[{"source_uuid": SOURCE_ID2, "type": "azure_devops"}],
+        )
+
+        measurement_2.copy_entity_user_data(measurement_1)
+        self.assertIn("entity_user_data", measurement_2.sources()[0])
```
wkoot:
```suggestion
        self.assertIn("entity_user_data", measurement_2.sources()[0])
        self.assertIsNotNone(measurement_2.sources()[0]["entity_user_data"])
```

wkoot: hmm maybe the walrus operator works here too, I'm not sure if that's possible :D

components/shared_code/tests/shared/database/test_measurements.py
```diff
@@ -35,12 +35,15 @@ def test_latest_measurement(self):
 
     def test_no_latest_successful_measurement(self):
         """Test no successful measurements found."""
-        self.assertIsNone(latest_successful_measurement(self.database, self.metric))
+        self.assertIsNone(latest_measurement(self.database, self.metric, skip_measurements_with_error=True))
```
wkoot, line 38: Shouldn't there also be a test that a measurement with error can be found? Or is it pointless to test because it's all mocked anyway

components/collector/tests/database/test_measurements.py
```diff
@@ -106,3 +112,60 @@ def test_create_measurement_after_measurement_was_requested(self):
...
+            "copied_from": METRIC_ID,
+            "sources": {SOURCE_ID2: {"type": "pip", "copied_from": SOURCE_ID}},
+        }
+        self.database["reports"].insert_one(report)
+        create_measurement(self.database, self.measurement_data(metric_uuid=METRIC_ID2, source_uuid=SOURCE_ID2))
+        inserted_measurement = self.database.measurements.find_one(filter={"metric_uuid": METRIC_ID2})
+        self.assertNotIn("entity_user_data", inserted_measurement["sources"][0])
+
+    def test_create_measurement_for_copied_metric_without_origin(self):
+        """Test that the entity user data are not copied if the original metric was deleted."""
```
wkoot, line 160: Why does this happen, is it a strong reference? Or was the metric deleted in between clicks?

components/frontend/src/source/SourceEntities.jsx
fniessink: Yeah, not sure it happens in production though. I think I ran into this because a feature tests created a source without entities array.

components/shared_code/tests/shared/model/test_measurement.py
fniessink: Added.

components/shared_code/tests/shared/model/test_measurement.py
fniessink: I've changed the assert to: 

```python
self.assertIn("key", measurement_2.sources()[0]["entity_user_data"])
```

components/shared_code/tests/shared/database/test_measurements.py
fniessink: There is a test for that. The previous test doesn't pass `skip_measurements_with_error` so it defaults to `False` there.

components/collector/tests/database/test_measurements.py
fniessink: The latter. In theory the original metric could be deleted before the copy operation is finished. I wouldn't have bothered with this test normally, but the code that copies the metric needs to get the original metric. Getting the original metric may return None, so there is indeed a code path where the original metric does not exist. Hence the unit test.

PR #12020, opened 2025-09-24
Make Dependency-Track landing URL mandatory

### LINKED ISSUE #11896: When no landing page URL is set for a Dependency-Track source the entities mistakenly point to Quality-time itself
**Describe the bug**
When no landing page URL is set for a Dependency-Track source, the entities mistakenly point to Quality-time itself

**To Reproduce**
1. Go to a Sec Warnings metric having DT as a source AND having entities and NOT having the landing page set
2. Click on entities tab
3. Click on a component or project link

**Expected behaviour**
The landing page URL should be a mandatory field.

fniessink:
Make the landing URL of Dependency-Track sources mandatory, since without it Quality-time sends users to the Dependency-Track API.

Fixes #11896.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -17,6 +17,7 @@ If your currently installed *Quality-time* version is not the penultimate versio
 ### Fixed
 
 - Small improvements to the documentation about importing reports. Fixes [#12013](https://github.com/ICTU/quality-time/issues/12013).
+- Make the landing URL of Dependency-Track sources mandatory, since without it Quality-time sends users to the Dependency-Track API. Fixes [#11896](https://github.com/ICTU/quality-time/issues/11896).
 
 ## v5.42.0 - 2025-09-18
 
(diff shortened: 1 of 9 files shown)


docs/src/changelog.md
```diff
@@ -17,6 +17,7 @@ If your currently installed *Quality-time* version is not the penultimate versio
 ### Fixed
 
 - Small improvements to the documentation about importing reports. Fixes [#12013](https://github.com/ICTU/quality-time/issues/12013).
+- Make the landing URL of Dependency-Track sources mandatory, since without it Quality-time sends users to the Dependency-Track API. Fixes [#11896](https://github.com/ICTU/quality-time/issues/11896).
```
wkoot, line 20: Isn't this a Changed?

wkoot (approved):

docs/src/changelog.md
fniessink: I don't think so: the landing URL should have been mandatory from the start since the API URL was never a reasonable URL to send the users to.

PR #12035, opened 2025-09-26
Replace parsing Gatling internal files with HTML

### LINKED ISSUE #11955: Add support for the Gatling .html report
Recent versions of Gatling don't output the internal components anymore, like: stats.json, global_stats.json, assertions.xml and assertions.json. They were eliminated and won’t be restored. Quality-Time used to parse these outputs for the metrics.

Add support for the Gatling .html report so there is no need anymore for: stats.json and simulation.log

**Additional context**
https://docs.gatling.io/tutorials/faq/#dropped-internals

fniessink:
Quality time parsed files that are no longer produced by Gatling. Parse the HTML instead.

Closes #11955.


components/collector/tests/source_collectors/gatling/test_slow_transactions.py
```diff
@@ -13,79 +13,79 @@ def setUp(self):
...
         """Test that the number of slow transactions is 0 if there are no transactions in the JSON."""
-        response = await self.collect(get_request_json_return_value={})
+        response = await self.collect(get_request_text="<div/>")
         self.assert_measurement(response, value="0")
 
     async def test_all_transactions(self):
-        """Test retrieving all transaction."""
-        response = await self.collect(get_request_json_return_value=self.GATLING_JSON)
-        self.assert_measurement(response, value="2", entities=self.expected_entities)
+        """Test retrieving all slow transaction."""
```
wkoot:
```suggestion
        """Test retrieving all slow transactions."""
```

components/collector/tests/source_collectors/gatling/test_slow_transactions.py
```diff
@@ -13,79 +13,79 @@ def setUp(self):
...
+        self.set_source_parameter("transactions_to_include", [self.TRANSACTION1])
+        response = await self.collect(get_request_text=self.GATLING_STATS_HTML)
         self.assert_measurement(response, value="1", entities=self.expected_entities[:1])
 
     async def test_evaluate_different_response_time(self):
         """Test that a different response time type can be evaluated against the target response time."""
         self.set_source_parameter("response_time_to_evaluate", "min_response_time")
-        self.set_source_parameter("target_response_time", "45")
-        response = await self.collect(get_request_json_return_value=self.GATLING_JSON)
+        self.set_source_parameter("target_response_time", "7")
```
wkoot, line 74: Shouldn't we also assert that setting it to 6 yields both entities?

components/collector/tests/source_collectors/gatling/test_slow_transactions.py
```diff
@@ -13,79 +13,79 @@ def setUp(self):
...
         self.assert_measurement(response, value="1", entities=self.expected_entities[:1])
 
     async def test_transaction_specific_response_time_target(self):
         """Test that a transaction specific target response time can be set."""
-        self.set_source_parameter("transaction_specific_target_response_times", ["[Bb]ar:150"])
-        response = await self.collect(get_request_json_return_value=self.GATLING_JSON)
-        self.assert_measurement(response, value="1", entities=self.expected_entities[:1])
+        self.set_source_parameter("transaction_specific_target_response_times", ["[Tt]ransaction 1:400"])
+        response = await self.collect(get_request_text=self.GATLING_STATS_HTML)
+        self.assert_measurement(response, value="0")
```
wkoot, line 85: Don't we want to assert that the measurement then contains the overridden time?

components/collector/tests/source_collectors/gatling/test_slow_transactions.py
```diff
@@ -13,79 +13,79 @@ def setUp(self):
...
+        """Test that an API URL that ends with index.html is not changed."""
         self.set_source_parameter("url", "https://gatling/index.html")
-        response = await self.collect(get_request_json_return_value={})
-        self.assert_measurement(response, api_url="https://gatling/js/stats.json")
+        response = await self.collect(get_request_text="")
+        self.assert_measurement(response, api_url="https://gatling/index.html")
 
     async def test_api_url_not_ending_with_index_html(self):
-        """Test that an API URL that ends with index.html is transformed to an URL that end with js/stats.json."""
+        """Test that an API URL that does not end with index.html get index.html added."""
```
wkoot:
```suggestion
        """Test that an API URL that does not end with index.html, has index.html added."""
```

components/collector/src/collector_utilities/date_time.py
```diff
@@ -28,6 +31,17 @@ def parse_datetime(text: str) -> datetime:
     return date_time.replace(tzinfo=tzlocal()) if date_time.tzinfo is None else date_time
 
 
+def parse_duration(text: str) -> int:
+    """Parse the duration in minutes from the text. Expects '<nr>h <nr>m <nr>s' with all groups optional."""
+    regexp = r"(?:(?P<hours>\d+)h\s*)?(?:(?P<minutes>\d+)m\s*)?(?:(?P<seconds>\d+)s\s*)?"
```
wkoot: This feels like something you might waste a while on when trying to debug. The `\s*` can be brought out of the groups and I don't think the non-capturing groups are needed, since only named groups are looked up. Something like this should provide the same functionality:
```
rs = r"((?P<seconds>\d+)s)?"
rm = r"((?P<minutes>\d+)m)?"
rh = r"((?P<hours>\d+)h)?"
regexp = r"\s*".join((rh, rm, rs))
```

components/collector/tests/collector_utilities/test_date_time.py
```diff
@@ -119,3 +120,47 @@ def test_hours(self):
     def test_days(self):
         """Test multiple days."""
         self.assertEqual(2 * 24 * 60, minutes(timedelta(days=2)))
+
+
+class ParseDurationTest(unittest.TestCase):
+    """Unit tests for the parse_duration function."""
```
wkoot, line 126: Perhaps some more specific combinations can be covered

components/collector/src/source_collectors/gatling/base.py
```diff
@@ -1,44 +1,35 @@
...
+        return api_url if api_url.endswith(self.INDEX_HTML) else URL(api_url + self.INDEX_HTML)
+
+    async def simulation_information(self, response: Response, label: str) -> str:
+        """Return the simulation information labeled with the label."""
+        soup = await self._soup(response)
+        label_span = soup.find("span", {"class": "simulation-information-label"}, string=f"{label}: ")
+        if label_span and (parent := label_span.parent):
+            return str(parent.find_all("span")[1].get_text())  # The second span contains the information
+        return ""
+
```
wkoot, line 26: These two methods make very strong assumptions about the way the data in HTML is structured. I think it might be worth documenting it in code somewhere, but I'm not sure as to the best location. Perhaps a mention in the docstring that the data lookups are strongly based on assumptions of HTML structure?

components/collector/tests/source_collectors/gatling/test_slow_transactions.py
fniessink: The measurement does not contain the regular expression, if that's what you mean. But I have added a second sub test to check that a matching transaction is included in the measurement:

```python
        self.set_source_parameter("transaction_specific_target_response_times", ["[Tt]ransaction 1:400", ".*2:5"])
        response = await self.collect(get_request_text=self.GATLING_STATS_HTML)
        self.assert_measurement(response, value="1", entities=self.expected_entities[1:])
```

components/collector/src/source_collectors/gatling/base.py
fniessink: I'll add a comment, but it's sort of a given when parsing HTML that we depend on the HTML structure.

PR #12044, opened 2025-09-30
Fix enabled issue tracker fields

### LINKED ISSUE #12039: Make issue tracker fields inactive when issue tracker is none
**Describe the bug**
It is possible to fill in the issue tracker fields without picking an issue tracker type. The user can think they have configured the issue tracker while Quality-time ignores the filled in values because no issue type has been picked. 

**Expected behaviour**
Make the issue tracker fields inactive until an issue tracker type has been picked.

fniessink:
Disable issue tracker fields until the user picks an issue tracker type.

Fixes #12039.

### DIFF
--- a/components/frontend/src/report/IssueTracker.jsx
+++ b/components/frontend/src/report/IssueTracker.jsx
@@ -83,6 +83,9 @@ export function IssueTracker({ report, reload }) {
     const projectKey = report.issue_tracker?.parameters?.project_key
     const issueType = report.issue_tracker?.parameters?.issue_type
     const epicLink = report.issue_tracker?.parameters?.epic_link
+    // The onChange handler below would erronously save the string "None" as report.issue_tracker.type. That has been
+    // fixed in v5.44.0, but we have to take into account that the database may still contain "None":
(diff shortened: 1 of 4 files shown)


components/frontend/src/report/IssueTracker.jsx
```diff
@@ -83,14 +83,17 @@ export function IssueTracker({ report, reload }) {
     const projectKey = report.issue_tracker?.parameters?.project_key
     const issueType = report.issue_tracker?.parameters?.issue_type
     const epicLink = report.issue_tracker?.parameters?.epic_link
+    const issueTrackerFieldDisabled = disabled || !report.issue_tracker?.type || report.issue_tracker?.type === "None"
```
wkoot, line 88: Why `type: "None"` instead of `null` ?

wkoot (approved):

components/frontend/src/report/IssueTracker.jsx
fniessink: The onChange handler below would save the string "None" to the database. I've changed that, but we still have to take into account that the database may still contain "None" for the `report.issue_tracker.type` instead of a falsey value.

PR #12078, opened 2025-10-06
Fix Dependency-Track projects not found

### LINKED ISSUE #11898: When entering a non-existent project name for a Dependency-Track source, no error is shown
Enter a non-existent project name for a Dependency-Track source and see how Quality-time reports 0 instead of 'non existent project name'.

fniessink:
When measuring security warnings or dependencies with Dependency-Track as source, throw an error if no projects match instead of reporting zero warnings or dependencies.

Fixes #11898.

### DIFF
--- a/components/collector/src/source_collectors/dependency_track/base.py
+++ b/components/collector/src/source_collectors/dependency_track/base.py
@@ -4,6 +4,7 @@ from collections.abc import AsyncIterator
 from typing import Literal, cast
 
 from base_collectors import SourceCollector
+from collector_utilities.exceptions import CollectorError
 from collector_utilities.functions import add_query, match_string_or_regular_expression
 from collector_utilities.type import URL, Response
 from model import Entity, SourceResponses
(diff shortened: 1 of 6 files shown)


components/collector/src/source_collectors/dependency_track/security_warnings.py
```diff
@@ -44,8 +45,11 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
         """Extend to get the findings."""
         api_url = await self._api_url()
         self._projects = await self._get_projects_by_uuid()
-        project_finding_urls = [URL(f"{api_url}/finding/project/{uuid}") for uuid in self._projects]
-        return await super()._get_source_responses(*project_finding_urls)
+        if self._projects:
+            project_finding_urls = [URL(f"{api_url}/finding/project/{uuid}") for uuid in self._projects]
+            return await super()._get_source_responses(*project_finding_urls)
+        error_message = "No projects found"
```
wkoot: This is duplicated code, can it not fit in shared parent `DependencyTrackLatestVersionStatusBase` or even higher?

fniessink: Yeah, you're right. Moved to `DependencyTrackBase`.

wkoot (approved):

PR #12113, opened 2025-10-13
Fix retrieving data from Jira Cloud

### LINKED ISSUE #12093: Quality Time uses a Jira API endpoint that was removed
**Describe the bug**

The metric type "Issues" uses the `/rest/api/2/search` and `/rest/api/3/search` with an optional `jql` query to measure the number of issues that match that query. Unfortunately, the `/search` endpoint [has recently been removed](https://developer.atlassian.com/changelog/#CHANGE-2046) from Jira. (The deprecation warning mentions May 1, but this continued to work until recently.)

This means that the "Issues" metric now gets a `410: GONE` response from the API, with the following response message:

```json
{"errorMessages":["The requested API has been removed. Please migrate to the /rest/api/3/search/jql API. A full migration guideline is available at https://developer.atlassian.com/changelog/#CHANGE-2046"],"errors":{}}
```

I think this is the line in the source code that builds the URL: https://github.com/ICTU/quality-time/blob/180ad185e0af48be2cb964246a95b414ce38ce1a/components/collector/src/source_collectors/jira/issues.py#L36

**To Reproduce**
1. Go to or create an 'Issues' metric in Quality
2. Open the metric
3. Go to "Sources"
4. Configure it to talk to a recently updated version of Jira
5. Observe an error message similar to the one below appearing after the metric is updated:

Image

**Screenshots**

Image

 - OS: MacOs
 - Browser: Mozilla Firefox 143.0.4 (aarch64)
 - Quality-time version: v5.36.1

Our instance of Quality Time is running on Red Hat OpenShift (Logius Standaard Platform) using ICTU container images. It communicates with a Jira Cloud instance, which means we cannot downgrade the Jira version that we use for compatibility with Quality Time.

fniessink:
When the Jira source is a Jira Cloud instance, use the Jira Cloud endpoints.

Fixes #12093.

### DIFF
--- a/components/collector/src/source_collectors/jira/base.py
+++ b/components/collector/src/source_collectors/jira/base.py
@@ -2,6 +2,8 @@
 
 from typing import TYPE_CHECKING
 
+import aiohttp
+
 from base_collectors import SourceCollector
 from collector_utilities.exceptions import CollectorError
 from collector_utilities.type import URL
@@ -32,6 +34,17 @@ class JiraBase(SourceCollector):
     def _rest_api_version(self) -> str:
         """Return the Jira REST API version set by the user."""
         return str(self._parameter("api_version"))
+
+    @property
+    async def is_cloud(self) -> bool:
+        """Return whether the URL points to Jira Cloud or not."""
+        server_info_api = URL(f"{await super()._api_url()}/rest/api/{self._rest_api_version}/serverInfo")
+        try:
+            responses = await super()._get_source_responses(server_info_api)
+            json = await responses[0].json()
+            return bool(json.get("deploymentType") == "Cloud")
+        except (aiohttp.ClientError, AttributeError):
+            return False
 
 
 class JiraBoardBase(JiraBase):
--- a/components/collector/src/source_collectors/jira/issues.py
+++ b/components/collector/src/source_collectors/jira/issues.py
@@ -22,18 +22,27 @@ class JiraIssues(JiraBase):
 
     async def _api_url(self) -> URL:
         """Extend to get the fields from Jira and create a field name to field id mapping."""
-        url = await super()._api_url()
-        fields_url = URL(f"{url}/rest/api/{self._rest_api_version}/field")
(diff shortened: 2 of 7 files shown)


components/collector/src/source_collectors/jira/base.py
```diff
@@ -33,6 +36,18 @@ def _rest_api_version(self) -> str:
         """Return the Jira REST API version set by the user."""
         return str(self._parameter("api_version"))
 
+    @cached_property
+    async def is_cloud(self) -> bool:
```
wkoot, line 39: Why lookup instead of having the user specify this in the API Version option?

components/collector/src/source_collectors/jira/base.py
```diff
@@ -33,6 +36,18 @@ def _rest_api_version(self) -> str:
         """Return the Jira REST API version set by the user."""
         return str(self._parameter("api_version"))
 
+    @cached_property
```
wkoot: This caching is only during the lifetime of the class instance, while the property actually refers to the source (url)

fniessink: Good point. I'll add the URL as parameter and make it a method so the result is cached for each URL separately.

components/collector/src/source_collectors/jira/base.py
fniessink: User friendliness: one parameter less for the user to set.

fniessink: Oh wait, you mean combine it with the API version?

wkoot: Indeed :-) Something stringy like "v3-cloud"

fniessink: Even better would be to get the API-version automatically as well. I'll add an issue for that.

components/collector/src/source_collectors/jira/base.py
fniessink: On second thought, caching isn't needed at all as the collector isn't kept around anyway.

PR #12206, opened 2025-11-03
Add used/unused metrics to source overview

### LINKED ISSUE #12122: Show used metrics and unused metric types in source overview
# User story

As a user, for each source, I want to see which metrics use that source and which metric types I can add using that source, so I can easily review my report for completeness.

The goal of this story is not to replace the missing-metric metric.

## Acceptance criteria

- [x] For each source in a report, show the metrics using that source.
- [ ] For each source in a report, show the metric types that haven't been used with that source
- [ ] Show the number of metric types that haven't been used with that source.

# Tasks

### Front-end

- [ ] Add a column to the source overview with the number of metric types that haven't been used with each source.
- [x] Add tab under each source in the source overview with:
  - [x] A table with the existing metrics with columns for metric name and subject name. 
  - [x] Subject names link to subject
  - [x] Metric names link to metric
- [ ] Add a tab under each source in the source overview with the metric types that can be added in a table.

## Documentation

- [ ] Update changelog

fniessink:
Show used metrics and unused metric types in the source overview of a report.

Closes #12122.

- [x] Indent source details panels under the source name
- [x] Link metric types to docs
- [x] Add rationale column
- [x] Alternating background color for rows (light grey)

### DIFF
--- a/components/frontend/src/report/report_utils.jsx
+++ b/components/frontend/src/report/report_utils.jsx
@@ -10,8 +10,18 @@ import {
     reportPropType,
     reportsPropType,
     settingsPropType,
+    sourcePropType,
+    sourceTypePropType,
 } from "../sharedPropTypes"
-import { addCounts, getMetricScale, getMetricTags, getSourceName, visibleMetrics } from "../utils"
+import {
+    addCounts,
+    getMetricName,
+    getMetricScale,
+    getMetricTags,
+    getSourceName,
+    getSubjectName,
+    visibleMetrics,
+} from "../utils"
 
 export function measurementOnDate(date, measurements, metricUuid) {
     const isoDateString = date.toISOString().split("T")[0]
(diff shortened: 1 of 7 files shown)


components/frontend/src/report/report_utils.jsx
```diff
@@ -133,3 +129,111 @@ reportSources.propTypes = {
...
+    dataModel: dataModelPropType,
+    report: reportPropType,
+    source: sourceTypePropType,
+}
+
+function findSourceId(dataModel, report, targetSourceUuid) {
+    // Return the stringified version of the source with targetSourceUuid
+    for (const { source, sourceUuid } of iterateSources(report)) {
+        if (sourceUuid === targetSourceUuid) {
+            return createSourceId(dataModel, source)
```
wkoot: Seems odd that a `find` function calls a `create` function. But I guess it's ok because it's just string creation and not object instantiation?

components/frontend/src/report/report_utils.jsx
```diff
@@ -10,8 +10,18 @@ import {
     reportPropType,
     reportsPropType,
     settingsPropType,
+    sourcePropType,
```
wkoot, line 13: Is there no `report_utils.test.jsx` ?

fniessink: No, we have 100% test coverage via the components that use the report utils. I like it that way: makes it possible to refactor those utilities without having to change unit tests.

components/frontend/src/report/report_utils.jsx
fniessink: It is odd indeed. I'll see if I can refactor this.

PR #12280, opened 2025-11-18
Don't allow floats with the manual number source

### LINKED ISSUE #12261: Manual Number gets stuck in endless loop
**Describe the bug**
when using manual number as a source, it ends up in an endless loop when a comma value is entered.

**To Reproduce**
1. create metric compliance
2. add source manual number
3. enter as value 9,25 OR 9.25

Image

**ADDITIONAL**
⚠️ please also update the documentation (https://quality-time.readthedocs.io/en/latest/reference.html#manual-number). It says nothing about the allowed values (positive natural numbers).

fniessink:
Fixes #12261.

### DIFF
--- a/components/frontend/src/fields/TextField.jsx
+++ b/components/frontend/src/fields/TextField.jsx
@@ -25,7 +25,7 @@ export function TextField({
     const [textValue, setTextValue] = useState(value)
 
     function submitIfChanged() {
-        if (textValue !== value) {
+        if (textValue !== value && isValid()) {
             onChange(textValue)
         }
     }
@@ -37,6 +37,16 @@ export function TextField({
         if (event.key === "Enter") {
             submitIfChanged()
         }
+    }
+
+    function isValid() {
+        if (required && !textValue) {
+            return false
+        }
+        if (type === "number") {
+            return /^\d+$/.test(textValue)
+        }
+        return true
     }
 
     const startInputAdornment = startAdornment ? (
@@ -54,7 +64,7 @@ export function TextField({
         <MUITextField
             defaultValue={textValue ?? ""}
             disabled={disabled || (select && children.length === 0)}
-            error={error || (required && !textValue)}
+            error={error || !isValid()}
             fullWidth
             helperText={helperText}
             id={id}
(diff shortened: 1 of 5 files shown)


components/frontend/src/widgets/buttons/ReportUploadButton.jsx
```diff
@@ -34,7 +34,6 @@ export function ReportUploadButton({ reload }) {
                     data-testid="report-import-input"
                     type="file"
                     onChange={(event) => {
-                        console.log(event.target.files)
```
wkoot, line 37: hehe

docs/src/changelog.md
```diff
@@ -17,6 +17,7 @@ If your currently installed *Quality-time* version is not the penultimate versio
 ### Fixed
 
 - Clarify in the product vision that checking the correct configuration of sources is out of scope for Quality-time. Fixes [#3821](https://github.com/ICTU/quality-time/issues/3821).
+- Don't allow floats with the manual number source. Fixes [#12261](https://github.com/ICTU/quality-time/issues/12261).
```
wkoot, line 21: I expected a backend change as well, do we not have type checking for this because the user input is all stringyfied?

components/frontend/src/fields/TextField.jsx
```diff
@@ -39,6 +39,16 @@ export function TextField({
         }
     }
 
+    function isValid() {
+        if (required && !textValue) {
+            return false
+        }
+        if (type === "number") {
+            return /^\d+$/.test(textValue)
```
wkoot, line 47: Hmm this feels a bit flimsy, can we not actually send an int to the backend and use the 'being able to store native types' feature?

docs/src/changelog.md
fniessink: No, we don't check the values for e.g. source parameters either in the backend.

components/frontend/src/fields/TextField.jsx
fniessink: Not sure what you mean by "the 'being able to store native types' feature"?

wkoot: Hmm, no wait I misremembered `SourceMeasurement.entities_to_store` - this is explicitly not sent to DB but only retained within the run (and metric collectors)

PR #12282, opened 2025-11-18
Fix suppressed violations SonarQube filters

### LINKED ISSUE #12263: Suppressed violation filter does not yield expected result
**Describe the bug**
When using the suppressed violations metric, having SonarQube as a source, the filter 'Impacted Software Qualities' should enable the user to separate the results in different categories.

**To Reproduce**
1. Create two metric for suppressed violations
2. Give both metric the source SonarQube 
3. For one metric, set filter 'Impacted Software Qualities' to maintainability
4. For the other metric, set filter 'Impacted Software Qualities' to security
5. See both metrics displaying the same entities

**Expected behaviour**
One metric should show security results, one metric should show maintainability.

**Screenshots**
(upon request, if necessary)

fniessink:
When measuring suppressed violations with SonarQube as source, take the impacted software qualities into account if configured by the user.

Fixes #12263.

### DIFF
--- a/components/collector/src/source_collectors/sonarqube/suppressed_violations.py
+++ b/components/collector/src/source_collectors/sonarqube/suppressed_violations.py
@@ -39,6 +39,8 @@ class SonarQubeSuppressedViolations(SonarQubeViolations):
         all_issues_api_url = URL(f"{url}/api/issues/search?projects={component}&branch={branch}")
         resolved_issues_api_url = URL(
             f"{all_issues_api_url}&resolved=yes&resolutions=WONTFIX,FALSE-POSITIVE&additionalFields=comments"
+            f"{self._query_parameter('clean_code_attribute_categories', uppercase=True)}"
+            f"{self._query_parameter('impacted_software_qualities', uppercase=True)}"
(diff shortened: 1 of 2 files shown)


components/collector/src/source_collectors/sonarqube/suppressed_violations.py
```diff
@@ -39,6 +39,8 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
         all_issues_api_url = URL(f"{url}/api/issues/search?projects={component}&branch={branch}")
         resolved_issues_api_url = URL(
             f"{all_issues_api_url}&resolved=yes&resolutions=WONTFIX,FALSE-POSITIVE&additionalFields=comments"
+            f"{self._query_parameter('clean_code_attribute_categories', uppercase=True)}"
```
wkoot, line 42: These all work the same on sonar cloud/server and api v2/3?

wkoot (approved):

components/collector/src/source_collectors/sonarqube/suppressed_violations.py
fniessink: Yes. Note there's no API v3 and issue searching is not part of the API v2 (yet).

PR #12283, opened 2025-11-18
Fix report source overview

### LINKED ISSUE #12281: Report Source Configuration has empty URL's
**Describe the bug**
Report Source Configuration has empty URL's in Quality-time 5.46.

**Screenshots**

Image

fniessink:
Source configurations in the report source overview did not display parameter values.

Fixes #12281.

### DIFF
--- a/components/frontend/src/report/ReportSources.test.jsx
+++ b/components/frontend/src/report/ReportSources.test.jsx
@@ -180,7 +180,7 @@ it("changes the value of a parameter of a source without parameter layout", asyn
         },
         ["source_uuid:0"],
     )
-    await userEvent.type(screen.getAllByLabelText(/URL/)[0], "https://source.org/new{Enter}")
+    await userEvent.type(screen.getAllByLabelText(/URL/)[0], "/new{Enter}")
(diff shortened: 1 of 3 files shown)


components/frontend/src/report/ReportSources.test.jsx
```diff
@@ -180,7 +180,7 @@ it("changes the value of a parameter of a source without parameter layout", asyn
         },
         ["source_uuid:0"],
     )
-    await userEvent.type(screen.getAllByLabelText(/URL/)[0], "https://source.org/new{Enter}")
+    await userEvent.type(screen.getAllByLabelText(/URL/)[0], "/new{Enter}")
```
wkoot, line 183: I don't understand how this test now covers the `.parameters` addition, do we actually test both the case where they are and are not present?

fniessink: This bug escaped because I previously changed this test when I shouldn't have. So this is a revert of that change so that the test again catches the bug.

wkoot: Perhaps leave an inline comment so it's not changed again in the future :D

PR #12288, opened 2025-11-19
Fix Cobertura XML timestamp

### LINKED ISSUE #12247: Throw exception instead of using timestamp 0 for cobertura xml
**Describe the bug**
In `components/collector/src/source_collectors/cobertura/source_up_to_dateness.py` we default to value 0 for timestamp, which results in using 01-01-1970.

**To Reproduce**
1. Add cobertura xml (e.g. produced by mypy) as source for source up-to-dateness metric
2. Wait for collector to run
3. See source up-to-dateness of >20k days

**Expected behaviour**
The collector should throw an exception informing the user that the chosen source does not provide the required information

fniessink:
When using Cobertura XML files as source, be prepared for timestamps without milliseconds. Also, report an error if the timestamp is missing.

Fixes #12247.

### DIFF
--- a/components/collector/src/source_collectors/cobertura/source_up_to_dateness.py
+++ b/components/collector/src/source_collectors/cobertura/source_up_to_dateness.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 
 from base_collectors import TimePassedCollector, XMLFileSourceCollector
 from collector_utilities.date_time import datetime_from_timestamp
+from collector_utilities.exceptions import NotFoundError
 from collector_utilities.functions import parse_source_response_xml
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


components/collector/src/source_collectors/cobertura/source_up_to_dateness.py
```diff
@@ -15,7 +16,12 @@
...
 
+    CUTOFF_YEAR_FOR_TIMESTAMP = 1980
+
     async def _parse_source_response_date_time(self, response: Response) -> datetime:
         """Override to parse the timestamp from the response."""
         tree = await parse_source_response_xml(response)
-        return datetime_from_timestamp(int(tree.get("timestamp", 0)))
+        if timestamp := tree.get("timestamp"):
+            dt = datetime_from_timestamp(float(timestamp))
+            return dt if dt.year > self.CUTOFF_YEAR_FOR_TIMESTAMP else datetime_from_timestamp(float(timestamp) * 1000)
```
wkoot, line 26: It seems that timestamps are typically denoted in `seconds (10-digit), milliseconds (13-digit) and microseconds (16-digit)`.. I think it's better to implement this broader use case in the `collector_utilities.date_time.datetime_from_timestamp`, *even though* the Python documentation specifically says it is **in seconds**. The fact that we've already been defaulting to millis shows that it's not very uniform

components/collector/tests/source_collectors/cobertura/test_source_up_to_dateness.py
```diff
@@ -19,13 +18,27 @@ def setUp(self):
...
+
     async def test_source_up_to_dateness(self):
         """Test that the source age in days is returned."""
-        response = await self.collect(get_request_text=self.COBERTURA_XML)
+        response = await self.collect(get_request_text=self.cobertura_xml())
+        self.assert_measurement(response, value=self.expected_age)
+
+    async def test_source_up_to_dateness_with_timestamp_in_seconds(self):
+        """Test that the source age in days is returned."""
+        response = await self.collect(get_request_text=self.cobertura_xml(timestamp="1553821197"))
```
wkoot: Perhaps add/subtract 24*60*60 and see if it's expected age +/- 1

components/collector/src/source_collectors/cobertura/source_up_to_dateness.py
fniessink: I thought about it and decided against this because that would affect a lot of sources at once. Basically all sources that the support the source up-to-dateness metric. Seemed too much of a risk.

PR #12292, opened 2025-11-20
Remove migration code.

### LINKED ISSUE #12289: Remove migrations
Remove the migrations from `components/api_server/src/initialization/migrations.py`.

Update the versioning policy in `docs/src/versioning.md`:
- Add "removed all" to latest version, column "Migrations"
- Replace "latest" with the version in which the migrations are removed in the column "Upgrade".

fniessink:
Closes #12289.


components/api_server/src/initialization/database.py
```diff
@@ -27,7 +26,6 @@ def init_database(database: Database) -> None:  # pragma: no feature-test-cover
     import_datamodel(database)
     initialize_secrets(database)
     initialize_reports_overview(database)
-    perform_migrations(database)
```
wkoot, line 30: Shouldn't we keep this with some NO-OP action, so it's easier to re-use next time?

wkoot (approved):

components/api_server/src/initialization/database.py
fniessink: That's what I did first but then the unit tests become a bit silly. I reckoned it's just a easy to copy back the code when we get the next migration.

PR #12329, opened 2025-11-28
Change addition when changing metric direction

### LINKED ISSUE #12142: Switch min/max addition strategy when switching metric direction
**Describe the bug**
When adding a metric, the default addition is saved as metric attribute in the database. When the user changes the direction of the metric, the addition is not adapted, for example from "min" to "max".

**To Reproduce**
1. Add a metric with multiple sources with different values
2. Note the value reported (e.g. the minimum of the values)
3. Change the direction of the metric
4. Note that the value reported is not changed (e.g. to the maximum of the values)

- Quality-time: v5.44.2, but this bug has been present for a long time probably.

**Additional context**
Remove the addition attribute of metrics from the database via a migration.

fniessink:
When adding a metric, the default addition is saved as metric attribute in the database. When the user changes the direction of the metric, the addition is not adapted, for example from "min" to "max".

Fixes #12142.


components/api_server/tests/routes/test_metric.py
```diff
@@ -4,7 +4,7 @@
 
 import requests
 
-from shared_data_model.meta.metric import Unit
+from shared_data_model.meta.unit import Unit
```
wkoot, line 7: Oh, yikes! That was a sneaky bug D:

components/shared_code/src/shared/model/metric.py
```diff
@@ -91,7 +91,11 @@ def issue_ids(self) -> list[str]:
 
     def addition(self) -> Callable:
         """Return the addition operator of the metric: sum, min, or max."""
-        addition = self.get("addition") or self.__data_model["metrics"][self.type()]["addition"]
+        addition = self.__data_model["metrics"][self.type()].get("addition", "sum")
+        if addition in ("min", "max") and (actual_direction := self.get("direction")):
```
wkoot, line 103: I get it, but it took a few re-reads. Perhaps something to explain in the docstring?

components/shared_code/tests/shared/model/test_measurement.py
```diff
@@ -651,22 +712,12 @@ def test_percentage_is_100(self):
...
-        metric = self.metric(direction="<", addition="min")
+    def test_multiple_percentages_with_zero_denominator(self):
+        """Test that the value is the calculated percentage when the scale is percentage."""
+        metric = self.metric(scale="percentage", direction="<")
         sources = [
             self.source(metric, value="10", total="70"),
             self.source(metric, value="0", total="0"),
         ]
         measurement = self.measurement(metric, sources=sources)
-        self.assertEqual("0", measurement["percentage"]["value"])
```
wkoot, line 672: Hmm, so the business logic actually changed here; to filter out zero values in this specific case?

components/api_server/tests/routes/test_metric.py
fniessink: Not really a bug, just an import from `shared_data_model.meta.metric` that itself imported `Unit` from `shared_data_model.meta.unit`.

components/shared_code/src/shared/model/metric.py
fniessink: I'll add some explanation, yes.

components/shared_code/tests/shared/model/test_measurement.py
fniessink: This only looks that way due to the diff: the new line 723 is actually a replacement of the old line 662, not 672.

PR #12357, opened 2025-12-03
Add Jira API version to issue tracker config

### LINKED ISSUE #12107: Add API version parameter to issue tracker
When adding Jira as source the source can pick the API version, but not when adding Jira as issue tracker. 

Add the API version parameter to the issue tracker configuration and use it when collecting issue statuses.

Jira Server (Platform) only supports version 2 of the REST API: https://developer.atlassian.com/server/jira/platform/rest/v11002/intro/#structure

This means that if the API version is v3 we know we're dealing with Jira Cloud and we don't need to access the `serverInfo` endpoint to check for the deployment type.

fniessink:
When configuring Jira as issue tracker, also allow for specifying the Jira API version to use.

Fixes #12107.


tests/feature_tests/src/features/metric_issue.feature
```diff
@@ -65,6 +65,15 @@ Feature: metric issues
     And the client opens a new issue
     Then the new issue response error is 'Failed to resolve'
 
+  Scenario: creating a new issue with an completely configured issue tracker with API-version v3
```
wkoot:
```suggestion
  Scenario: creating a new issue in a fully configured issue tracker with API-version v3
```

wkoot: Perhaps even explicitly name it Jira?

tests/feature_tests/src/features/metric_issue.feature
```diff
@@ -65,6 +65,15 @@ Feature: metric issues
...
+    When the client changes the report tracker_type to "jira"
+    And the client changes the report tracker_api_version to "v3"
+    And the client changes the report tracker_url to "https://jira"
+    And the client changes the report tracker_project_key to "KEY"
+    And the client changes the report tracker_issue_type to "Task"
+    And the client opens a new issue
+    Then the new issue response error is 'Failed to resolve'
```
wkoot, line 75: How does this scenario test anything more than the previous one when it's failing on the url? Or is this merely to verify that there problems selecting v3 in the api_version toggle?

components/api_server/src/routes/metric.py
```diff
@@ -213,7 +213,7 @@ def create_issue_text(metric: Metric, measured_value: Value) -> tuple[str, str]:
     """Create an issue description for the metric."""
     metric_url = dict(bottle.request.json)["metric_url"]
     source_names = ", ".join([source.name or DATA_MODEL.sources[str(source.type)].name for source in metric.sources])
-    source_urls = [url for url in [get_source_url(source) for source in metric.sources] if url]
+    source_urls = [f"[{url}|{url}]" for url in [get_source_url(source) for source in metric.sources] if url]
```
wkoot, line 218: Why not `<url>`?

wkoot, line 218:
```suggestion
    source_urls = [f"<{url}>" for url in [get_source_url(source) for source in metric.sources] if url]
```

components/api_server/src/model/issue_tracker.py
```diff
@@ -113,29 +145,39 @@ def create_issue(self, summary: str, description: str = "") -> tuple[str, str]:
...
                 return "", f"Issue tracker has no {name} configured."
-        api_url = self.issue_creation_api % self.url
-        json = {
-            "fields": {
-                "project": {"key": project_key},
-                "issuetype": {"name": issue_type},
-                "summary": summary,
-                "description": description,
-            },
+        api_url = self.issue_creation_api
```
wkoot: Why still create this local var, instead of using `self.issue_creation_api` in the two cases below?

components/api_server/src/model/issue_tracker.py
```diff
@@ -147,18 +189,18 @@ def get_options(self) -> Options:  # pragma: no feature-test-cover
...
             message = "Retrieving issue id suggestions from %s failed: %s"
-            logger.warning(message, self.suggestions_api % (self.url, "<query redacted>"), reason)
+            logger.warning(message, self.suggestions_api("<query redacted>"), reason)
             return []
         return self._parse_suggestions(json)  # pragma: no feature-test-cover
 
     def browse_url(self, issue_key: str) -> URL:
         """Return a URL to a human readable version of the issue."""
-        return URL(self.issue_browse_url % (self.url, issue_key))  # pragma: no feature-test-cover
+        return URL(f"{self.url}/browse/{issue_key}")  # pragma: no feature-test-cover
```
wkoot, line 201: Why does this still use `self.url`, while the other cases were removed?

components/api_server/src/model/issue_tracker.py
```diff
@@ -249,3 +293,38 @@ def __post_json(self, api_url: str, json) -> dict[str, str]:  # pragma: no featu
...
+    def epics_api(self) -> str:  # pragma: no feature-test-cover
+        """Return the epics endpoint."""
+        return (
+            f'{self.search_api}?jql=type=epic and ("Epic Status" != Done or "Epic Status" is empty) and '
+            f"project={self.issue_parameters.project_key}&fields=summary&maxResults=100"
+        )
+
+    def suggestions_api(self, query: str) -> str:
+        """Return the suggestions endpoint."""
+        return f"{self.search_api}?jql=summary~'{query}~10' order by updated desc&fields=summary&maxResults=20"
```
wkoot, line 319: What does this `~10` do exactly? And why maxResults 20; does that automatically cut off the older ones based on the order by clause?

tests/feature_tests/src/features/metric_issue.feature
fniessink: The only difference is indeed the API-version. But that triggers different code paths (e.g. for the ADF format).

components/api_server/src/routes/metric.py
fniessink: Because `[anchor|link]` is the format used by v2 of the API. I'll add a comment.

fniessink: I've added a comment with a link to https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=links.

components/api_server/src/model/issue_tracker.py
fniessink: Because this is not an API endpoint, but it creates a landing URL for an individual issue.

components/api_server/src/model/issue_tracker.py
fniessink: It's proximity search: https://confluence.atlassian.com/jirasoftwareserver/search-syntax-for-text-fields-939938747.html. I'll add the link as a comment.

And yes: we show the 20 most recent suggestions, based on the query typed by the users.

PR #12375, opened 2025-12-08
Filter metric types when adding metrics

### LINKED ISSUE #10527: Allow for ignoring metrics reported by the missing metrics metric
I'm always frustrated when the 'Missing Metrics' metric reports a metric as missing in one subject while it is present in another subject. Or when the 'Missing Metrics' metric reports a metric as missing that I consider optional.

Make it possible to ignore metrics that are being used in another subject or that are optional.

Add two source configuration parameters:

### LINKED ISSUE #12374: Allow for hiding metric types already used in subject or report
When adding a metric to a subject, allow for hiding metric types in the subject or in the whole report.

fniessink:
When adding metrics to subjects, allow for hiding metric types already used in the subject or in the whole report.

Closes #12374.

### DIFF
--- a/components/frontend/src/widgets/buttons/AddDropdownButton.jsx
+++ b/components/frontend/src/widgets/buttons/AddDropdownButton.jsx
@@ -7,6 +7,8 @@ import {
     MenuItem,
     MenuList,
     Popover,
+    Radio,
+    RadioGroup,
     TextField,
     Tooltip,
     Typography,
@@ -27,7 +29,7 @@ FilterCheckbox.propTypes = {
     setFilter: func,
 }
 
-function FilterCheckboxes({
+function Filters({
     itemType,
     allowHidingUnsupportedItems,
     showUnsupportedItems,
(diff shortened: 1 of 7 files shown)


components/frontend/src/widgets/buttons/AddDropdownButton.jsx
```diff
@@ -78,9 +102,11 @@ export function AddDropdownButton({ itemSubtypes, itemType, onClick, allItemSubt
     const [query, setQuery] = useState("") // Search query to filter item subtypes
     const [showUnsupportedItems, setShowUnsupportedItems] = useState(false) // Show only supported itemSubTypes or also unsupported itemSubTypes?
     const [hideUsedItems, setHideUsedItems] = useState(false) // Hide itemSubTypes already used?
+    const [hideUsedItemsScope, setHideUsedItemsScope] = useState("report") // Hide itemSubTypes already used in report or subject?
     let items = showUnsupportedItems ? allItemSubtypes : itemSubtypes
+    let usedItemKeys = hideUsedItemsScope === "report" ? usedItemSubtypeKeysInReport : usedItemSubtypeKeysInSubject
```
wkoot, line 107: I don't see exactly how `items` is updated when the radio is switched between 'report' and 'subject'. Is that re-evaluated automatically on each action?

wkoot (approved):

components/frontend/src/widgets/buttons/AddDropdownButton.jsx
fniessink: That's the magic of React. When the user switches between report and subject, the state is updated and the component rerendered.

PR #12386, opened 2025-12-09
Rename 'Tests' and 'Test suites'

### LINKED ISSUE #12322: Rename the 'tests' metric to 'test results'
## Tasks

- [x] Rename 'Tests' to 'Test results'.
- [x] Rename 'Test suites' to 'Test suite results'.
- [x] Changelog: Note new default names.

## Won't do

- Renaming the metric types from `tests` to `test_results` in the data model. 
- Renaming `test_suites` to `test_suite_results` in the data model.

fniessink:
Rename the 'Tests' metric to 'Test results' and the 'Test suites' metric to 'Test suite results' to better distinguish them from the 'Test cases' metric. 'Tests' and 'Test suites' metrics whose default name has not been overriden will automatically use the new default name.

Closes #12322.


components/shared_code/src/shared_data_model/sources/quality_time.py
```diff
@@ -53,8 +53,8 @@
     "Test branch coverage": "uncovered_branches",
     "Test line coverage": "uncovered_lines",
     "Test cases": "test_cases",
-    "Test suites": "test_suites",
-    "Tests": "tests",
+    "Test suite results": "test_suites",
```
wkoot, line 56: I presume we don't want to rename existing metrics with the default name, as we usually don't make a migration for that right?

fniessink: Even better: that isn't needed. Existing metrics that don't have an overridden name get the new name automatically from the data model.

PR #12387, opened 2025-12-09
Fix incorrect sprint names

### LINKED ISSUE #12328: Jira sources with entities show wrong sprint names
Image

fniessink:
When measuring issues with Jira as source, the names of sprints that issues belong to would not be correct.

Fixes #12328.

### DIFF
--- a/components/collector/src/source_collectors/jira/issues.py
+++ b/components/collector/src/source_collectors/jira/issues.py
@@ -12,7 +12,7 @@ from .base import JiraBase
 class JiraIssues(JiraBase):
     """Jira collector for issues."""
 
-    SPRINT_NAME_RE = re.compile(r",name=(.*),startDate=")
+    SPRINT_NAME_RE = re.compile(r",name=(.*),\w+=")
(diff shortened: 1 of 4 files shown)


components/collector/src/source_collectors/jira/issues.py
```diff
@@ -12,7 +12,7 @@
 class JiraIssues(JiraBase):
     """Jira collector for issues."""
 
-    SPRINT_NAME_RE = re.compile(r",name=(.*),startDate=")
+    SPRINT_NAME_RE = re.compile(r",name=(.*),rapidViewId=")
```
wkoot: Hmmm, is there not a some more robust way? I'm worried about getting mismatches between server and cloud at some point, the params don't need to adhere to an ordering right.. Perhaps something like `name=([^,]+)` - assuming that a comma is not allowed in sprint name? Or split the string either on `,` or `=` and select the right slice holding the name?

fniessink: Yeah, I'll see if I can make it more robust.

PR #12490, opened 2026-01-13
Singular and plural versions of metric units

### LINKED ISSUE #550: Provide singular and plural versions of the unit name
Tasks:
- Add singular and plural version of units to the data model.
- The user should be able to override both the singular and the plural version of the unit name.
- Use the singular version of the unit when the metric value is 1.
- Use the overridden version of the unit name in the measurement entity table headers.

Not to be changed:
- The metric type as displayed as header when the metric is expanded.

fniessink:
Add singular versions of the metric units to the data model and allow the user to override both the singular and plural forms via the UI.

Closes #550.


components/api_server/src/routes/metric.py
```diff
@@ -217,11 +217,12 @@ def create_issue_text(metric: Metric, measured_value: Value) -> tuple[str, str]:
     # for the text formatting notation used by Jira API version 2.
     source_urls = [f"[{url}|{url}]" for url in [get_source_url(source) for source in metric.sources] if url]
     percentage = "%" if metric.scale() == "percentage" else ""
-    issue_summary = f"Fix {measured_value}{percentage} {metric.unit} from {source_names}"
+    unit = metric.unit_singular if measured_value in ("1", "1.0") else metric.unit
```
wkoot: What about percentage, date/time values and strings such as version?

components/collector/src/source_collectors/quality_time/metrics.py
```diff
@@ -77,6 +77,7 @@ def __parse_entity(self, entity: Entity, metric, landing_url: URL) -> Entity:
         status_start = self.__metric_status_start(metric.get("status_start", ""))
         entity["status_start_date"] = status_start.isoformat() if status_start else ""
         entity["unit"] = metric.get("unit") or DATA_MODEL.metrics[metric["type"]].unit.value
+        entity["unit_singular"] = metric.get("unit_singular") or DATA_MODEL.metrics[metric["type"]].unit_singular.value
```
wkoot, line 80: Same here

components/frontend/src/subject/SubjectTableRow.test.jsx
wkoot: Shouldn't this then be altered to `+1` ?

components/frontend/src/subject/SubjectTableRow.test.jsx
```diff
@@ -135,8 +136,8 @@ it("works for informative metrics", async () => {
 it("takes the date order into account", async () => {
     history.push("?nr_dates=3&date_interval=1&date_order=ascending")
     const { container } = renderSubjectTableRow({ ascending: true })
-    expectText("+2")
-    expectLabelText("Metric type worsened from 10 to 12 things by +2 things")
```
wkoot, line 139: Do we still test the plural case `+2 things`?

components/frontend/src/source/SourceEntities.jsx
```diff
@@ -212,7 +212,7 @@ export function SourceEntities({ loading, measurements, metric, metricUuid, relo
     const metricEntities = dataModel.sources[sourceType]?.entities?.[metric.type]
 
     if (!metricEntities) {
-        const unit = dataModel.metrics[metric.type].unit || "entities"
+        const unit = getMetricUnit(metric, dataModel) || "entities"
```
wkoot, line 215: This seems awfully similar to a `formatMetricScale()` call with a value of 1

components/notifier/src/destinations/ms_teams.py
```diff
@@ -21,16 +21,26 @@ def metric_section(metric: MetricNotificationData, report_url: str) -> pymsteams
...
+        old_value = "unknown"
+    else:
+        old_value = f"{metric.old_metric_value}{unit_text(metric, metric.old_metric_value)}"
     old_value_text = " (unchanged)" if metric.new_metric_value == metric.old_metric_value else f", was {old_value}"
     section.addFact("Value:", f"{new_value}{old_value_text}")
     section.linkButton("View metric", f"{report_url}#{metric.metric_uuid}")
     return section
 
 
+def unit_text(metric: MetricNotificationData, value: str) -> str:
```
wkoot, line 38: This seems to be kindof duplicate with `components/api_server/src/routes/metric.py`

components/shared_code/src/shared_data_model/meta/metric.py
```diff
@@ -61,3 +62,12 @@ def set_default_scale(self) -> Self:
         """If the metric supports just one scale, make that scale the default scale."""
         self.default_scale = self.scales[0] if len(self.scales) == 1 else self.default_scale
         return self
+
+    @model_validator(mode="after")
+    def check_unit(self) -> Self:
+        """Check that if the singular unit is not none, neither is the plural unit, and vice versa."""
+        if (self.unit, self.unit_singular).count(Unit.NONE) == 1:
```
wkoot, line 74: What's the rationale for having both unit strings None? Perhaps good for in the docstring or in a comment

components/api_server/src/routes/metric.py
wkoot: In `components/frontend/src/utils.jsx` we're also checking on negative values, should the same happen here? Or perhaps check on `abs(val)`?

wkoot: Actually, since we're also explicitly dealing with floats, should the value not be in the closed interval `[-1,1]\0` (without zero)? I.e. `0,5 story point` (half a story point) instead of `0,5 story points`, but still `1,5 story points` (one and a half story points).

components/frontend/src/utils.jsx
```diff
@@ -393,9 +396,9 @@ export function formatMetricScale(metric, dataModel) {
     return scale === "percentage" ? "%" : ""
 }
 
-export function formatMetricScaleAndUnit(metric, dataModel) {
+export function formatMetricScaleAndUnit(metric, dataModel, value) {
```
wkoot, line 407: This file seems to randomly include or omit `propTypes` to functions, is there a reason for it? :))

components/api_server/src/routes/metric.py
fniessink: - Percentages are dealt with in the code you quoted, so not sure what you are asking.
- I'm pretty sure measurements are never date/time values (but days or minutes). But if they were, they would not have a unit.
- Versions don't have a unit.
- Negative values: I don't think we normally have negative measurement values, but can't hurt to add -1. The -1 and -1.0 in `utils.jsx` were added for delta's between measurement values.
- It seems native English speakers would say "zero dot five story pointS" (https://www.quora.com/Is-0-5-singular-or-plural) and even "one dot zero story pointS". I guess we should only use the singular unit with "1" and "-1". I'll remove 1.0 and -1.0 from the unity checks.

components/collector/src/source_collectors/quality_time/metrics.py
fniessink: Not sure what you mean with "same"?

components/frontend/src/subject/SubjectTableRow.test.jsx
fniessink: good catch!

components/frontend/src/subject/SubjectTableRow.test.jsx
fniessink: No, but there's the -4 things. So both the plural and singular unit form of the delta are tested.

components/collector/src/source_collectors/quality_time/metrics.py
wkoot: Sorry, was lazy. A reference to the comments placed on `components/api_server/src/routes/metric.py`

fniessink: Yeah, I got that. But I don't see how those comments are relevant for this code? Lines 79 and 80 are just getting the plural and singular version of the unit from the metric or the data model.

components/frontend/src/source/SourceEntities.jsx
fniessink: The calls look similar, but `getFormatScale(metric, dataModel)` returns the scale of the metric, not the unit.

components/collector/src/source_collectors/quality_time/metrics.py
wkoot: Indeed it is, I must have misinterpreted on first read

components/notifier/src/destinations/ms_teams.py
fniessink: Resolved by adding a `Metric.unit_for_value(self, value: Value) -> str` method and using that in both locations.

components/shared_code/src/shared_data_model/meta/metric.py
fniessink: Both unit and unit_singular are None for metrics that measure the version scale for example. I'll add a comment.

components/frontend/src/utils.jsx
fniessink: WIP :-) I've started to add `propTypes` to regular functions (in addition to React components), but haven't finished yet.

wkoot (approved):

PR #12491, opened 2026-01-13
Count issues linked to multiple metrics only once

### LINKED ISSUE #12483: The same issue linked to multiple metric is counted for each metric
**Describe the bug**
When a Jira issue is linked to multiple metrics, the issue summary card counts the issue for each metric.

**To Reproduce**
1. In a report without other linked issues, link the same Jira issue to two metrics
2. Note that the issue summary card says there are two issues

**Expected behaviour**
If the same issue is linked to multiple metrics, it should be counted only once.

- Quality-time: v5.48.1 (but has been present for a long time)

fniessink:
Count a Jira issue linked to multiple metrics only once in the issues dashboard card.

Fixes #12483.

### DIFF
--- a/components/frontend/src/dashboard/IssuesCard.test.jsx
+++ b/components/frontend/src/dashboard/IssuesCard.test.jsx
@@ -9,12 +9,17 @@ const report = {
             metrics: {
                 metric_uuid: {
                     issue_ids: ["ID-1"],
-                    issue_status: [{ status_category: "doing" }],
+                    issue_status: [{ issue_id: "ID-1", status_category: "doing" }],
                 },
                 another_metric_uuid: {
-                    issue_ids: ["ID-2", "ID-3"],
-                    issue_status: [{ connection_error: "oops" }],
+                    issue_ids: ["ID-1", "ID-2", "ID-3"],
+                    issue_status: [
+                        { issue_id: "ID-1", status_category: "doing" },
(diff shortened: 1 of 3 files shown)


components/frontend/src/dashboard/IssuesCard.test.jsx
```diff
@@ -9,12 +9,17 @@ const report = {
             metrics: {
                 metric_uuid: {
                     issue_ids: ["ID-1"],
-                    issue_status: [{ status_category: "doing" }],
+                    issue_status: [{ issue_id: "ID-1", status_category: "doing" }],
```
wkoot, line 12: So ID-1 is attached to two metrics and only counted as "Doing" once. Is it possible for the same issue ID having multiple statusses; i.e. once "doing" and once "done", when it happened to get resolved in between collector runs for the various metrics? Is that an edge case that we need to catch?

wkoot (approved):

components/frontend/src/dashboard/IssuesCard.test.jsx
fniessink: Yes, that's possible because the issue status is collected, and stored, per metric. To prevent this issue we could store issue status in one `issue_status` list per report instead of a list per metric.  Ticket: https://github.com/ICTU/quality-time/issues/12493

PR #12520, opened 2026-01-22
Fix PDF-export on Kubernetes

### LINKED ISSUE #12514: Exporting to PDF doesn't work when using the Kubernetes deployment
**Describe the bug**
Exporting to PDF doesn't work when using the Kubernetes deployment due to a missing service in the renderer Helm chart.

**To Reproduce**
1. Export a report to PDF.
2. Observe the notification of a 500 status.

**Your environment**
- Quality-time: v5.48.2

fniessink:
Add missing service to renderer Helm chart that was causing PDF-exports to fail on Kubernetes.

Fixes #12514.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -17,6 +17,7 @@ If your currently installed *Quality-time* version is not the penultimate versio
 - When creating an issue from Quality-time in Jira, make sure that the link back to Quality-time doesn't expand any items but only goes to the metric. Fixes [#12481](https://github.com/ICTU/quality-time/issues/12481).
 - Count a Jira issue linked to multiple metrics only once in the issues dashboard card. Fixes [#12483](https://github.com/ICTU/quality-time/issues/12483).
 - Remove restart policy from docker-compose.yml. Fixes [#12506](https://github.com/ICTU/quality-time/issues/12506).
+- Fix missing service in renderer Helm chart that was causing PDF-exports to fail on Kubernetes. Fixes [#12514](https://github.com/ICTU/quality-time/issues/12514).
 
 ## v5.48.2 - 2026-01-09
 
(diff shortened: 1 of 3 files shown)


.github/workflows/helm-chart.yml
```diff
@@ -42,4 +42,10 @@ jobs:
           helm upgrade --install --render-subchart-notes quality-time helm
           kubectl wait --all pods --timeout=2m --for=condition=Ready
           kubectl wait --all deployments --timeout=30s --for=condition=Available
-        timeout-minutes: 3
```
wkoot, line 45: I'd keep the explicit timeout, it helps a lot with failing deployments. The additional steps can be put in a separate task

tests/application_tests/src/test_report.py
```diff
@@ -51,10 +52,13 @@ def setUp(self):
         """Override to setup the driver."""
         firefox_options = webdriver.FirefoxOptions()
         firefox_options.add_argument("--headless")
-        self.driver = webdriver.Remote("http://selenium:4444", options=firefox_options)
+        selenium_url = f"http://{self.SELENIUM_HOSTNAME}:{self.SELENIUM_PORT}"
```
wkoot: Why not make `selenium_url` and `application_url` properties in `ApplicationTestCase`?

.github/workflows/helm-chart.yml
```diff
@@ -42,4 +42,10 @@ jobs:
...
+          export QUALITY_TIME_HOSTNAME=localhost
+          export QUALITY_TIME_PORT=8080
+          export SELENIUM_HOSTNAME=localhost
+          export SELENIUM_PORT=4444
+          kubectl port-forward service/quality-time-www ${APPLICATION_PORT}:${APPLICATION_PORT} &
+          docker run -d selenium/standalone-firefox -p ${SELENIUM_PORT}:${SELENIUM_PORT}
```
wkoot: I'm not sure you can run regular docker containers at this point, it may have been hijacked by minikube

fniessink: I think the problem is that the browser runs in a container and needs to use the host (port 8080) to get to the Quality-time webserver. I'll try using a browser on the Ubuntu host instead of in a container.

PR #12592, opened 2026-02-05
Allow for ignoring draft merge requests when using GitLab.

### LINKED ISSUE #12327: Allow for ignoring draft merge requests when using GitLab
I'm always frustrated when Quality-time shows open merge request, but they are drafts.

A possibility to configure the source (GitLab) to ignore drafts.

Image

as discussed with @wkoot

Sebastiaan127001:
Closes #12327

### DIFF
--- a/components/collector/src/source_collectors/gitlab/merge_requests.py
+++ b/components/collector/src/source_collectors/gitlab/merge_requests.py
@@ -46,6 +46,7 @@ query MRs($projectId: ID!) {{
         createdAt
         updatedAt
         mergedAt
+        draft
         {approved}
       }}
     }}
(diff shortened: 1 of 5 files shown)


components/collector/src/source_collectors/gitlab/merge_requests.py
```diff
@@ -143,6 +144,7 @@ def _create_entity(self, merge_request) -> Entity:
             merged=merge_request["mergedAt"],
             downvotes=str(merge_request["downvotes"]),
             upvotes=str(merge_request["upvotes"]),
+            draft=str(merge_request.get("draft", "False")),
```
wkoot: Why not lowercase here, instead of in the frontend? Also - if the source data is a bool, wouldn't it be cleaner to default to `False` instead of `"False"`?

wkoot (approved):

components/collector/src/source_collectors/gitlab/merge_requests.py
Sebastiaan127001: fixed

wkoot (approved):

Issue #12560, opened 2026-01-29
Adding a custom requests_ca_bundle to the api_server in the Helm chart


tijder:
**Is your feature request related to a problem? Please describe.**
I am now running Quality Time behind a proxy that intercepts all HTTP and HTTPS traffic. The HTTPS traffic is re-encrypted with its own certificate. This causes the API server to log all kinds of certificate issues.
Now I can solve the problem by mounting a volume with the correct certificate and then setting an env variable REQUEST_CA_BUNDLE to the path of the certificate.

**Describe the solution you'd like**
It would be nice if the Helm chart had an option to specify additional env, volumeMounts, and volumes for the api_server deployment.

**Describe alternatives you've considered**
To customize it myself in the forked version of the Helm chart.

**Additional context**
If you like, I can also make a pull request with the desired changes.

fniessink:
A PR would be appreciated!

Issue #12559, opened 2026-01-29
Helm chart deployen without security context


tijder:
**Is your feature request related to a problem? Please describe.**
I would like to deploy Quality-Time in an OpenShift environment. There, I am unable to adjust many security context settings. Currently, the www and database deployment have a customized security context in the chart. So I'm not able to deploy to OpenShift yet.
If I modify the chart and remove the security context, it still works. However, I then have to build the proxy image myself without the CAP_NET_BIND_SERVICE rule.

**Describe the solution you'd like**
I would like to be able to overwrite the security context in values.yaml. It would also be nice if there were an unprivileged proxy image or that the proxy is not needed and everything is handled by the ingress.

**Describe alternatives you've considered**
Modify the helm chart myself and build a custom proxy image.

**Additional context**
If you like, I can also make a pull request with the desired changes.

fniessink:
A PR would be appreciated!

PR #12671, opened 2026-02-17
Helm Chart Option for Use Without Proxy

tijder:
In this pull request, I implemented the necessary changes for #12559 and #12560.

- The www.deployProxy property has been added. When set to `false`, the proxy is bypassed, allowing direct access to the API and frontend 
- An option has been included to add extra volumes to the `api_server`.
- An option to modify the `securityContext` in the database has been introduced.

### DIFF
--- a/helm/templates/api_server.yaml
+++ b/helm/templates/api_server.yaml
@@ -61,6 +61,10 @@ spec:
                   optional: true
             - name: LDAP_SEARCH_FILTER  # override to make sure that double dollar signs are processed like in docker
               value: {{ .Values.api_server.env.LDAP_SEARCH_FILTER }}
+          {{- if .Values.api_server.extraVolumeMounts }}
+          volumeMounts:
+            {{- .Values.api_server.extraVolumeMounts | nindent 12 }}
+          {{- end }}
 {{- with .Values.api_server.resources }}
           resources: {{- toYaml . | nindent 12 }}
 {{- end }}
@@ -69,6 +73,9 @@ spec:
               drop:
                 - ALL
       restartPolicy: Always
+      {{- if .Values.api_server.extraVolumes }}
+      {{- .Values.api_server.extraVolumes | nindent 8 }}
+      {{- end }}
 ---
 apiVersion: v1
 kind: Service
--- a/helm/templates/api_server.yaml
+++ b/helm/templates/api_server.yaml
@@ -63,7 +63,7 @@ spec:
               value: {{ .Values.api_server.env.LDAP_SEARCH_FILTER }}
           {{- if .Values.api_server.extraVolumeMounts }}
           volumeMounts:
-            {{- .Values.api_server.extraVolumeMounts | nindent 12 }}
+            {{- toYaml .Values.api_server.extraVolumeMounts | nindent 12 }}
           {{- end }}
 {{- with .Values.api_server.resources }}
           resources: {{- toYaml . | nindent 12 }}
@@ -74,7 +74,8 @@ spec:
                 - ALL
       restartPolicy: Always
       {{- if .Values.api_server.extraVolumes }}
-      {{- .Values.api_server.extraVolumes | nindent 8 }}
+      volumes:
+        {{- toYaml .Values.api_server.extraVolumes | nindent 8 }}
       {{- end }}
 ---
 apiVersion: v1
--- a/helm/templates/api_server.yaml
+++ b/helm/templates/api_server.yaml
@@ -61,19 +61,15 @@ spec:
                   optional: true
             - name: LDAP_SEARCH_FILTER  # override to make sure that double dollar signs are processed like in docker
               value: {{ .Values.api_server.env.LDAP_SEARCH_FILTER }}
-          {{- if .Values.api_server.extraVolumeMounts }}
           securityContext:
             {{- toYaml .Values.api_server.securityContext | nindent 12 }}
+          {{- if .Values.api_server.extraVolumeMounts }}
           volumeMounts:
             {{- toYaml .Values.api_server.extraVolumeMounts | nindent 12 }}
           {{- end }}
 {{- with .Values.api_server.resources }}
           resources: {{- toYaml . | nindent 12 }}
 {{- end }}
-          securityContext:
-            capabilities:
-              drop:
-                - ALL
       restartPolicy: Always
       {{- if .Values.api_server.extraVolumes }}
       volumes:
--- a/helm/templates/api_server.yaml
+++ b/helm/templates/api_server.yaml
@@ -62,6 +62,8 @@ spec:
             - name: LDAP_SEARCH_FILTER  # override to make sure that double dollar signs are processed like in docker
               value: {{ .Values.api_server.env.LDAP_SEARCH_FILTER }}
           {{- if .Values.api_server.extraVolumeMounts }}
+          securityContext:
+            {{- toYaml .Values.api_server.securityContext | nindent 12 }}
           volumeMounts:
             {{- toYaml .Values.api_server.extraVolumeMounts | nindent 12 }}
           {{- end }}
--- a/helm/templates/database.yaml
+++ b/helm/templates/database.yaml
@@ -57,14 +57,7 @@ spec:
           resources: {{- toYaml . | nindent 12 }}
 {{- end }}
           securityContext:
-            capabilities:
-              add:
-                - CHOWN
-                - DAC_OVERRIDE
-                - SETGID
-                - SETUID
-              drop:
-                - ALL
+            {{- toYaml .Values.database.securityContext | nindent 12 }}
           volumeMounts:
             - mountPath: /data/db
               name: {{ .Release.Name }}-{{ template "database_name" . }}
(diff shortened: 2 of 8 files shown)


helm/values.yaml
```diff
@@ -42,5 +53,17 @@ renderer:
...
     repository: "ictu/quality-time_proxy"
+  ingress:
+    enabled: false
+    ingressClassName: ""
+    hostname: ""
+    pathType: "ImplementationSpecific"
+    tls: []
+    #  - secretName: chart-example-tls
+    #    hosts:
+    #      - chart-example.local
```
fniessink: It might be clearer to uncomment these and add comments what needs to be changed before deploying the chart.

helm/templates/www.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+{{- if not .Values.www.directlyToService }}
```
fniessink: I'm not sure I'm a fan of `directlyToService`, it raises the question: what/who directly to what service? How about `deployProxy` with a default value of `true`?

fniessink (commented):
I have asked @wkoot for a review as well, he's way more knowledgeable in Helm charts than I am.

helm/templates/www.yaml
tijder: I think it's a good suggestion. I have changed it.

helm/values.yaml
tijder: I see this happening in many more helmet charts. That's why I did it this way.
- [kube-prometheus-stack](https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack?modal=values&path=alertmanager.ingress.tls)
- [gitlab](https://artifacthub.io/packages/helm/gitlab/gitlab?modal=values&path=global.ingress.tls
)

But I can change it.

helm/templates/api_server.yaml
```diff
@@ -61,6 +61,10 @@ spec:
                   optional: true
             - name: LDAP_SEARCH_FILTER  # override to make sure that double dollar signs are processed like in docker
               value: {{ .Values.api_server.env.LDAP_SEARCH_FILTER }}
+          {{- if .Values.api_server.extraVolumeMounts }}
```
wkoot, line 66: Can you include the used options in `values.yaml` or document their usage? I'd be interested to see how it is applied exactly, now they're just empty lists. Also, why not make `api_server.extraVolumes` a list of dictionaries which contain both volume name and mountpoint instead of two separate lists?

helm/templates/database.yaml
```diff
@@ -57,14 +57,7 @@ spec:
...
           securityContext:
-            capabilities:
-              add:
-                - CHOWN
-                - DAC_OVERRIDE
-                - SETGID
-                - SETUID
-              drop:
-                - ALL
+            {{- toYaml .Values.database.securityContext | nindent 12 }}
```
wkoot, line 60: Perhaps surround the option with `{{- if .Values.database.securityContext }}`, make the whole thing optional (i.e. for local deployment)

helm/templates/www.yaml
```diff
@@ -72,8 +75,9 @@ spec:
       port: 8080
       targetPort: 8080
   sessionAffinity: None
+{{- end }}
 ---
-{{- if .Values.www.ingress -}}
+{{ if .Values.www.ingress.enabled -}}
```
wkoot: Is there a use case for having ingress disabled, but still wanting to specify `www.ingress` options?
And is there a reason to the different approach to trimming before/after in the various blocks?

helm/templates/www.yaml
```diff
@@ -82,27 +86,55 @@ metadata:
     app.kubernetes.io/name: {{ .Chart.Name }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: {{ template "www_name" . }}
+  {{- with .Values.www.ingress.annotations }}
   annotations:
-    {{- range $key, $value := .Values.www.ingress.annotations }}
-    {{ $key }}: {{ $value | quote }}
-    {{- end }}
+    {{- toYaml . | nindent 4 }}
```
wkoot: Does this include the `|quote` functionality, otherwise good to document either in values or readme

helm/templates/www.yaml
```diff
@@ -82,27 +86,55 @@ metadata:
...
+  {{- end }}
 spec:
-  ingressClassName: {{ .Values.www.ingress.ingressClassName }}
+  {{- if .Values.www.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.www.ingress.ingressClassName | quote }}
+  {{- end }}
   rules:
-    - host: {{ printf "%s" $.Values.www.ingress.hostname }}
+    {{- if .Values.www.ingress.hostname }}
```
wkoot: So if there is no hostname provided, then the `rules` block is empty - correct? Does that work?

helm/values.yaml
wkoot: If uncommented, then it probably needs to be applied in values-ci and readme as well.

helm/values.yaml
```diff
@@ -16,6 +18,15 @@ collector:
     repository: "ictu/quality-time_collector"
 
 database:
+  securityContext:
```
wkoot, line 36: Shouldn't the context override possibility also be applied to all other pods?

helm/templates/www.yaml
```diff
@@ -82,27 +86,55 @@ metadata:
...
+                name: {{ .Release.Name }}-{{ template "frontend_name" . }}
+                port:
+                  number: 5000
+        {{- end }}
+    {{- end }}
+  {{- if .Values.www.ingress.tls }}
   tls:
-{{ toYaml .Values.www.ingress.tls | indent 4 }}
-{{- end }}
+    {{- range .Values.www.ingress.tls }}
```
wkoot: Why does this one use `range` instead of `{{- with .Values.www.ingress.tls }}`?

helm/values.yaml
tijder: I reverted the tls changes and removed his lines

helm/templates/api_server.yaml
tijder: I included examples in the values.yaml. I think its better to have them seperated. That way you have more control over how it's used.

helm/templates/database.yaml
tijder: It's in the default yaml file (it can't be null). So if you don't want to use securityContext you can set the value:
```
securityContext: {}
```

helm/templates/www.yaml
tijder: Yes so you can set default values in the yaml file. That way it's easer to see which values can be set.
I reverted the trimming change.

helm/templates/www.yaml
tijder: I reverted this back

helm/templates/www.yaml
tijder: I put the if statement one line up, so before the rules.

helm/templates/www.yaml
tijder: I reverted this changes.

helm/values.yaml
tijder: I have made the securityContext now configurable in all the deployments.

fniessink (approved):

fniessink:
@tijder Thanks for the PR. LGTM. One thing is possibly missing and that's an entry in the changelog if you think one is needed (it's a technical change so I don't think so, but just checking to be sure).

helm/templates/api_server.yaml
```diff
@@ -61,6 +61,12 @@ spec:
...
+          securityContext:
```
wkoot, line 64: Is the securityContext meant to be inside the extraVolumeMounts branch?

tijder: Sorry that is indeed not meant there. I corrected the securityContext bloc.

tijder:
> @tijder Thanks for the PR. LGTM. One thing is possibly missing and that's an entry in the changelog if you think one is needed (it's a technical change so I don't think so, but just checking to be sure).

@fniessink thanks! I agree.
I see some pipelines are still waiting for the status to be reported. Is there something I need to do?

wkoot (approved):

fniessink:
> @fniessink thanks! I agree. I see some pipelines are still waiting for the status to be reported. Is there something I need to do?

Yes, you need to "allow changes to a pull request branch created from a fork", see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

tijder:
I did set the checkbox, but it didn't trigger any pipeline. Did I forgot something?

afbeelding

fniessink:
> I did set the checkbox, but it didn't trigger any pipeline. Did I forgot something?

I don't think so. However, I expected to be able to make changes to your brach (rebase it) which would allow me to trigger all CI-pipelines. I'm getting `error: Authentication error: Authentication required: You must have push access to verify locks` when trying to `push -f` the rebased branch. I'll have to investigate.

fniessink:
Ok, so I managed to add a commit to trigger CI and I can rebase your branch. However, still not all checks run. I'll have to investigate some more.

fniessink:
Closing this PR in favour of #12986.

PR #12715, opened 2026-03-03
Improve change failure rate metric docs

### LINKED ISSUE #10526: Improve documentation on change failure rate.
on this page: 
https://quality-time.readthedocs.io/en/latest/reference.html#change-failure-rate
the documentation explains that two sources are needed (except when using Azure).

on this page:
https://quality-time.readthedocs.io/en/latest/reference.html#change-failure-rate-from-jira
the explanation can be improved

1. elaborate the parameter that says:
Issue query in JQL (Jira Query Language). → it does not tell when the query should yield

2. When using Jira, you should use two sources, but this section only shows only one. What configuration is used for the other.
3. How to tell Quality-time which source is the nominator and which one is the denominator?

fniessink:
Fixes #10526.


Sebastiaan127001 (approved):
👍 I believe you have covered all the feedback points. One thing: My grammar checker, that I have configured to use British English, prefers spelling look-back period with a hyphen.

PR #12842, opened 2026-03-27
Improve documentation for exporting reports

### LINKED ISSUE #12801: Complete the documentation regarding how to export a report
Documentation on the page explaining how to export a report is incomplete.
It says how to convert a public_key.json into a regular public_key file using Python, but it does not say that one should save it as a file.

 https://quality-time.readthedocs.io/en/latest/api.html#exporting-reports

fniessink:
Document how to use the public key of the destination Quality-time instance when exporting a quality report.

Fixes #12801.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -150,7 +150,7 @@ Replace the username and password with the username and password you use to logi
 The exporting endpoint is available via `https://quality-time.source.org/api/v3/report/<report-uuid>/json?public_key=<public-key>`.
 The exporting endpoint returns JSON content only.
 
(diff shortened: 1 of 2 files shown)


Sebastiaan127001 (approved):
This was indeed this missing part in the documentation. Isn't it better to save the public key in a file and then add this to the curl command as argument?

fniessink:
> This was indeed this missing part in the documentation. Isn't it better to save the public key in a file and then add this to the curl command as argument?

No, because it's not an argument for curl, but a query parameter for the URL.

PR #12913, opened 2026-04-03
Add a parameter to exclude idle time from GitLab pipeline duration.

### LINKED ISSUE #12127: Expand the way CI pipeline duration is measured
I'm always frustrated when Quality-time reports a different CI pipeline duration when a pipeline had waiting times between jobs or restarts of jobs. The GitLab UI reports the net time and Quality-time reports the gross time.

The possibility to select the way Quality-time measures the pipeline duration in the source configuration.

Add a parameter "Include wait time" with possible values "yes" (default) and "no"

GitLab returns the following attributes that are relevant for computing wait time: `created_at`, `updated_at`, `started_at`, and `finished_at`, `duration`, and `queued_duration`.

The first four are timestamps with millisecond precision. The last two are minutes as whole numbers. The docs don't specify whether the duration includes the queued duration or not, but the [example](https://docs.gitlab.com/api/pipelines/#retrieve-a-single-pipeline) suggests it does not. It's also unclear how exactly the duration and queued duration are calculated from the `*_at` timestamps.

Conclusion: Quality-time will itself calculate the wait time by computing: `started_at - created_at`.

reported by @just-frank-it

Sebastiaan127001:
Closes #12127.

### DIFF
--- a/components/collector/src/source_collectors/gitlab/json_types.py
+++ b/components/collector/src/source_collectors/gitlab/json_types.py
@@ -39,18 +39,9 @@ class Pipeline(GitLabJSON):
     source: str
     created_at: str
     web_url: str
+    duration: int = 0
     updated_at: str = ""
     schedule_description: str = ""  # Pipeline schedule description for scheduled pipelines
-
-    @property
-    def start(self) -> datetime:
-        """Return the pipeline start time."""
-        return parse_datetime(self.created_at)
-
-    @property
-    def end(self) -> datetime:
-        """Return the pipeline end time."""
-        return parse_datetime(self.updated_at) if self.updated_at else datetime.now(tz=tzutc())
 
     @property
     def datetime(self) -> datetime:
@@ -58,9 +49,20 @@ class Pipeline(GitLabJSON):
         return parse_datetime(self.updated_at or self.created_at)
 
     @property
-    def duration(self) -> timedelta:
+    def bruto_duration(self) -> timedelta:
+        """Return the bruto duration of the pipeline, meaning the duration including idle time."""
+        start = parse_datetime(self.created_at)
+        end = parse_datetime(self.updated_at) if self.updated_at else datetime.now(tz=tzutc())
+        return end - start
+
+    @property
+    def netto_duration(self) -> timedelta:
+        """Return the netto duration of the pipeline, meaning the duration excluding idle time."""
+        return timedelta(minutes=self.duration)
+
+    def pipeline_duration(self, exclude_idle_time: bool) -> timedelta:
         """Return the duration of the pipeline."""
-        return self.end - self.start
+        return self.netto_duration if exclude_idle_time else self.bruto_duration
 
 
 @dataclass
--- a/components/collector/src/source_collectors/gitlab/pipeline_duration.py
+++ b/components/collector/src/source_collectors/gitlab/pipeline_duration.py
@@ -2,17 +2,27 @@
 
 from typing import TYPE_CHECKING
 
+from collector_utilities.date_time import minutes
 from collector_utilities.exceptions import CollectorError
 
 from .base import GitLabPipelineBase
 
 if TYPE_CHECKING:
     from collector_utilities.type import Value
-    from model import Entities, SourceResponses
+    from model import Entities, Entity, SourceResponses
+
+    from .json_types import Pipeline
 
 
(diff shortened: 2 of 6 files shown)


components/collector/src/source_collectors/gitlab/json_types.py
```diff
@@ -57,9 +58,10 @@ def datetime(self) -> datetime:
         """Return the datetime of the pipeline."""
         return parse_datetime(self.updated_at or self.created_at)
 
-    @property
-    def duration(self) -> timedelta:
+    def pipeline_duration(self, exclude_idle_time: bool) -> timedelta:
         """Return the duration of the pipeline."""
+        if exclude_idle_time:
+            return timedelta(minutes=self.duration)
         return self.end - self.start
```
fniessink:
```suggestion
        return timedelta(minutes=self.duration) if exclude_idle_time else self.end - self.start
```

components/collector/src/source_collectors/gitlab/pipeline_duration.py
```diff
@@ -2,18 +2,29 @@
...
+    from model import Entities, Entity, SourceResponses
+
+    from .json_types import Pipeline
 
 
 class GitLabPipelineDuration(GitLabPipelineBase):
     """GitLab CI-pipeline duration collector."""
 
+    def _create_entity(self, pipeline: Pipeline) -> Entity:
+        """Create an entity from a GitLab pipeline."""
```
fniessink:
```suggestion
        """Extend to also add the duration to the entity created from a GitLab pipeline."""
```

components/collector/src/source_collectors/gitlab/pipeline_duration.py
```diff
@@ -2,18 +2,29 @@
...
+        entity = super()._create_entity(pipeline)
+        entity["duration"] = str(
+            minutes(pipeline.pipeline_duration(self._parameter("pipeline_idle_time_exclude") == "yes"))
+        )
```
fniessink:
```suggestion
        exclude_idle_time = self._parameter("pipeline_idle_time_exclude") == "yes"
        entity["duration"] = str(minutes(pipeline.pipeline_duration(exclude_idle_time)))
```

docs/src/changelog.md
```diff
@@ -8,14 +8,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 If your currently installed *Quality-time* version is not the penultimate version, please first check the upgrade path in the [versioning policy](versioning.md) before upgrading.
 
-<!-- The line "## <square-bracket>Unreleased</square-bracket>" is replaced by the tools/release/release.py script with the new release version and release date. -->
+<!-- The line "## <square-bracket>Unreleased</square-bracket>" is replaced by the tools/release3#/release.py script with the new release version and release date. -->
```
fniessink: Remove this accidental edit please.

components/shared_code/src/shared_data_model/sources/gitlab.py
```diff
@@ -291,6 +291,12 @@
             default_value="slowest",
             metrics=["pipeline_duration"],
         ),
+        "pipeline_idle_time_exclude": SingleChoiceParameter(
```
fniessink: Please rename the parameter to "exclude_idle_time_from_pipeline_duration" for better descriptiveness.

fniessink (changes requested):

components/collector/src/source_collectors/gitlab/json_types.py
Sebastiaan127001: Please expound on the suggestion that you have made here.

fniessink: The ternary-if needs only one return statement and is shorter (three lines instead of one). The behaviour is the same.

It would read even better if it said: `return self.netto_duration if exclude_idle_time else self.bruto_duration`. 

That would require adding these two properties to the Pipeline class:

```python
    @property
    def bruto_duration(self) -> timedelta:
        """Return the bruto duration of the pipeline, meaning the duration including idle time."""
        return self.end - self.start

    @property
    def netto_duration(self) -> timedelta:
        """Return the netto duration of the pipeline, meaning the duration excluding idle time."""
        return timedelta(minutes=self.duration)
```

fniessink (approved):

PR #12932, opened 2026-04-08
Ignore notification destinations without webhook

### LINKED ISSUE #12906: Unconfigured MS Teams notification produces endless errors
**Describe the bug**
Unconfigured notification with MS Teams webhook produces endless errors in notifier

**To Reproduce**
1. Create a new MS Teams webhook notification
2. Do not configure it
3. Observe errors in notifier container

**Additional context**
Logging snippet:
```
requests.exceptions.MissingSchema:
Invalid URL '': No scheme supplied.
Perhaps you meant https://?
2026-04-02 12:32:17 ERROR Could not deliver notification ms_teams.py:62
File
"/home/notifier/destinations/ms_team
s.py", line 60, in send_notification
card.send()
```

fniessink:
Don't attempt to send notifications to notification destinations without a webhook.

Fixes #12906.

### DIFF
--- a/components/notifier/src/strategies/notification_strategy.py
+++ b/components/notifier/src/strategies/notification_strategy.py
@@ -38,6 +38,7 @@ class NotificationFinder:
                     [
                         Notification(report, destination.get("report_url", "").strip("/"), notable_metrics, destination)
                         for destination in destinations
+                        if destination["webhook"]
                     ],
                 )
         return notifications
--- a/components/notifier/tests/strategies/test_notification_strategy.py
+++ b/components/notifier/tests/strategies/test_notification_strategy.py
@@ -49,7 +49,11 @@ class StrategiesTests(unittest.TestCase):
                 "title": "Title",
                 "subjects": {SUBJECT_ID: {"metrics": {METRIC_ID: self.red_metric}, "type": "software"}},
                 "notification_destinations": {
-                    "destination_uuid": {"name": "destination", "report_url": "https://report"}
+                    "destination_uuid": {
+                        "name": "destination",
+                        "report_url": "https://report",
+                        "webhook": "https://xxxxx.webhook.office.com/xxxxxxxxx",
+                    }
                 },
             },
         ]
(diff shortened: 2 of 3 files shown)


components/notifier/src/strategies/notification_strategy.py
```diff
@@ -38,6 +38,7 @@ def get_notifications(
                     [
                         Notification(report, destination.get("report_url", "").strip("/"), notable_metrics, destination)
                         for destination in destinations
+                        if destination["webhook"]
```
wkoot, line 41: Although this problably works, we don't have any non-webhook notifications. Should a notification without webhook simply not be part of `notable_metrics` ?

components/notifier/tests/strategies/test_notification_strategy.py
```diff
@@ -178,7 +178,11 @@ def test_multiple_reports_with_same_destination(self):
             "webhook": "webhook",
             "subjects": {SUBJECT_ID: subject2},
             "notification_destinations": {
-                NOTIFICATION_DESTINATION_ID: {"report_url": "https://report2", "name": "destination2"},
+                NOTIFICATION_DESTINATION_ID: {
+                    "report_url": "https://report2",
+                    "name": "destination2",
+                    "webhook": "webhook",
```
wkoot: Is this the default value that all webhooks get? Did we design for non-webhook notifications, or is this just a superfluous field?

wkoot: Note that the default field input placeholder says "Webhook" and the comment says "Paste a Microsoft Teams webhook URL here."

components/notifier/src/strategies/notification_strategy.py
fniessink: This checks whether the user has actually entered a web hook URL in the destination and ignores the destination if it has no web hook URL. That was the cause of the problem: the notifier would try to send notifications to a destination even if the web hook URL was an empty string.

components/notifier/tests/strategies/test_notification_strategy.py
fniessink: No "webhook" is my lazy dummy web hook URL. I'll make it a proper example.

PR #13062, opened 2026-04-28
Update helm chart values and templates

### LINKED ISSUE #13061: Helm malformed yaml with deployProxy false but ingress enabled
Error:
```
$ helm template --debug --version v5.53.0 -f values.yaml qt oci://registry-1.docker.io/ictu/quality-time
[...]
Error: YAML parse error on quality-time/templates/www.yaml: error converting YAML to JSON: yaml: line 35: did not find expected key
```

The error appears to be in `| indent` of a block that's preceded by spaces

wkoot:
Closes #13061

### DIFF
--- a/helm/templates/database.yaml
+++ b/helm/templates/database.yaml
@@ -53,11 +53,10 @@ spec:
                   name: {{ .Values.secrets.dbCredential }}
                   key: DATABASE_PASSWORD
                   optional: true
-{{- with .Values.database.resources }}
+          {{- with .Values.database.resources }}
           resources: {{- toYaml . | nindent 12 }}
-{{- end }}
-          securityContext:
-            {{- toYaml .Values.database.securityContext | nindent 12 }}
+          {{- end }}
+          securityContext: {{- toYaml .Values.database.securityContext | nindent 12 }}
           volumeMounts:
             - mountPath: /data/db
               name: {{ .Release.Name }}-{{ template "database_name" . }}
@@ -82,10 +81,10 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.database.storageSize | default "1Gi" | quote }}
-{{- if .Values.database.storageClassName }}
-  storageClassName: {{ .Values.database.storageClassName | quote }}
-{{- end }}
+      storage: {{ .Values.database.storageSize }}
+  {{- if .Values.database.storageClassName }}
+  storageClassName: {{ .Values.database.storageClassName }}
+  {{- end }}
 ---
 apiVersion: v1
 kind: Service
@@ -107,6 +106,7 @@ spec:
       targetPort: 27017
   sessionAffinity: None
 ---
+{{- if .Values.database.env }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -116,6 +116,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: {{ template "database_name" . }}
 data:
-{{- range $key, $val := .Values.database.env }}
-  {{ $key }}: "{{ $val }}"
+  {{- range $key, $value := .Values.database.env }}
+  {{ $key }}: "{{ $value }}"
+  {{- end }}
 {{- end }}
--- a/helm/templates/frontend.yaml
+++ b/helm/templates/frontend.yaml
@@ -30,11 +30,10 @@ spec:
             - configMapRef:
                 name: {{ .Release.Name }}-{{ template "frontend_name" . }}-env
                 optional: true
-{{- with .Values.frontend.resources }}
+          {{- with .Values.frontend.resources }}
           resources: {{- toYaml . | nindent 12 }}
-{{- end }}
-          securityContext:
-            {{- toYaml .Values.notifier.securityContext | nindent 12 }}
+          {{- end }}
+          securityContext: {{- toYaml .Values.frontend.securityContext | nindent 12 }}
       restartPolicy: Always
 ---
 apiVersion: v1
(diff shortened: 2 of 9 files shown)


helm/values.yaml
wkoot, line 93: Open issue: when `deployProxy: false`, the renderer cannot produce output

fniessink: Have you opened a bug report for this issue?

helm/templates/database.yaml
```diff
@@ -82,10 +81,10 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.database.storageSize | default "1Gi" | quote }}
-{{- if .Values.database.storageClassName }}
-  storageClassName: {{ .Values.database.storageClassName | quote }}
-{{- end }}
+      storage: {{ .Values.database.storageSize }}
+  {{- if .Values.database.storageClassName }}
+  storageClassName: {{ .Values.database.storageClassName }}
```
fniessink, line 86: Why did you drop the `| quote`?

helm/templates/www.yaml
```diff
@@ -81,15 +80,15 @@ metadata:
     app.kubernetes.io/component: {{ template "www_name" . }}
   annotations:
     {{- range $key, $value := .Values.www.ingress.annotations }}
-    {{ $key }}: {{ $value | quote }}
+    {{ $key }}: "{{ $value }}"
     {{- end }}
 spec:
   {{- if .Values.www.ingress.ingressClassName }}
-  ingressClassName: {{ .Values.www.ingress.ingressClassName | quote }}
+  ingressClassName: {{ .Values.www.ingress.ingressClassName }}
```
fniessink, line 87: Why did you drop the `| quote`?

helm/templates/www.yaml
```diff
@@ -81,15 +80,15 @@ metadata:
...
   {{- end }}
   {{- if .Values.www.ingress.hostname }}
   rules:
-    - host: {{ .Values.www.ingress.hostname | quote }}
+    - host: {{ .Values.www.ingress.hostname }}
```
fniessink, line 91: Why did you drop the `| quote`?

helm/templates/database.yaml
```diff
@@ -82,10 +81,10 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: {{ .Values.database.storageSize | default "1Gi" | quote }}
-{{- if .Values.database.storageClassName }}
-  storageClassName: {{ .Values.database.storageClassName | quote }}
-{{- end }}
+      storage: {{ .Values.database.storageSize }}
```
fniessink, line 84: Why did you drop the `| quote`?

helm/templates/www.yaml
```diff
@@ -81,15 +80,15 @@ metadata:
     app.kubernetes.io/component: {{ template "www_name" . }}
   annotations:
     {{- range $key, $value := .Values.www.ingress.annotations }}
-    {{ $key }}: {{ $value | quote }}
+    {{ $key }}: "{{ $value }}"
```
fniessink, line 83: Why did you drop the `| quote`? The quotes around `{{ $value }}" won't escape quotes in the value, right?

helm/templates/frontend.yaml
```diff
@@ -30,11 +30,10 @@ spec:
...
                 name: {{ .Release.Name }}-{{ template "frontend_name" . }}-env
                 optional: true
-{{- with .Values.frontend.resources }}
+          {{- with .Values.frontend.resources }}
           resources: {{- toYaml . | nindent 12 }}
-{{- end }}
-          securityContext:
-            {{- toYaml .Values.notifier.securityContext | nindent 12 }}
+          {{- end }}
+          securityContext: {{- toYaml .Values.notifier.securityContext | nindent 12 }}
```
fniessink: Shouldn't this be: `.Values.`**`frontend`**`.securityContext`?

fniessink (changes requested):

helm/templates/database.yaml
wkoot: It was not needed to begin with, all storage classes are regular strings without special chars (not even space)

helm/templates/www.yaml
wkoot: It was not needed to begin with, all ingress classes are regular strings without special chars (not even space)

helm/templates/www.yaml
wkoot: Hostnames cannot contain special characters

helm/templates/database.yaml
wkoot: Storage sizes cannot contain special characters (i.e. 1Gi)

helm/templates/www.yaml
wkoot: Sadly, the `|quote` would not do so either. I've opted to remove it entirely because it isn't as expected, it just puts quotes around the string. The alternative is `quote str1 str2 str` which would simply quote each arg and separate them with spaces (if the separator were dynamic, we could at least use it for e.g. ":" concat in image string composure). I believe using explicit quotes is more clear than the filter which doesn't do what one would expect from it. See the implementation in https://github.com/Masterminds/sprig/blob/master/strings.go#L90

helm/templates/frontend.yaml
wkoot: That's a good find! I didn't check the original content

helm/values.yaml
wkoot: I was still hoping to do so here, but it is indeed too large and I'll split it off

helm/templates/frontend.yaml
wkoot: Verified all other cases of `.Values.` and this was the only one. I guess it was not noticed because the frontend and notifier have the same security context

fniessink (approved):

PR #13265, opened 2026-05-19
Add helm lint to `just check`

### LINKED ISSUE #13246: Add `helm lint` to `just check`
Add `helm lint --quiet --strict helm/` to `just check` in the root folder.

fniessink:
Closes #13246.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: quality-time
-version: v5.55.0-rc.0
+version: 5.55.0-rc.0
 appVersion: "v5.55.0-rc.0"
 description: Helm chart for Quality-time, an automated quality system for software development and maintenance
 type: application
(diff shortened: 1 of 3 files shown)


helm/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: quality-time
-version: v5.54.0
+version: 5.54.0
```
fniessink: Changed because `helm lint` complained:

```
==> Linting helm/
[WARNING] Chart.yaml: version 'v5.54.0' is not a valid SemVerV2

Error: 1 chart(s) linted, 1 chart(s) failed
```

Can this cause trouble downstream?

wkoot: I think it shouldn't cause big troubles, but perhaps worth a deployment note

wkoot: Does the version bump deal with the numeric version only, or expect the `v`-prefix?

wkoot (approved):

PR #13500, opened 2026-06-13
Prevent trailing newline in generated credentials

fniessink:
Use printf instead of echo so the base64-encoded secrets do not contain a trailing newline. MongoDB rejects usernames and passwords with control characters (SASLprep), which prevents the root user from being created and make all database authentication fail.


fniessink:
@wkoot FYI.

PR #13528, opened 2026-06-16
Harden database image

### LINKED ISSUE #13514: Run the database container as non-root and remove `gosu`
## Summary

All high-severity Trivy alerts on the database image come from the bundled
`gosu` binary. `gosu` is a statically-linked Go binary that the MongoDB
entrypoint uses only to drop privileges from root to the `mongodb` user. Its
36 open CVEs are all Go **standard-library** vulnerabilities; `gosu` is never
rebuilt upstream, so they cannot be patched by bumping anything.

By running the database container directly as the `mongodb` user (UID/GID 999)
we make `gosu` unnecessary, can delete it from the image, and can additionally
drop the Linux capabilities that only exist to support the root→gosu step.

This continues the production-image hardening (read-only root filesystem,
non-root components, reduced attack surface) already in progress.

## What this clears

- **All 36 `gosu` alerts**, including:
  - the only **critical**: `CVE-2025-68121`
  - ~14 of the 16 **high** alerts (`CVE-2025-61726`, `CVE-2025-61729`,
    `CVE-2026-25679`, `CVE-2026-32280/32281/32283`, `CVE-2026-33811/33814`,
    `CVE-2026-39820/39823/39825/39836`, `CVE-2026-42499/42504`, …)
- Bonus: lets us drop `CHOWN`, `DAC_OVERRIDE`, `SETGID`, `SETUID` from the
  database container's capabilities.

## Background / why it is safe

Verified against `mongo:8.3.2`:

- `mongodb` is **UID/GID 999**; `/data/db` and `/data/configdb` are **already
  owned by 999 in the image**.
- The entrypoint already chowns data to `mongodb` and runs `mongod` as 999 —
  it only uses `gosu` inside `if [ "$(id -u)" = '0' ]`. Starting as 999 makes
  it `exec mongod` directly, never touching `gosu`.
- Smoke-tested: `mongod` starts cleanly as UID 999 with a read-only root
  filesystem, `tmpfs` on `/tmp` and `/data/configdb`, and a fresh data volume
  (`mongod startup complete`, listening on `27017`).

Because the data directory is already 999-owned, **existing deployments need
no data migration**.

### 1. `components/database/Dockerfile`
- Remove the binary: add `rm -f /usr/local/bin/gosu` to the existing `RUN`.
- Run non-root by default: add `USER 999`.

### 2. `docker/docker-compose.yml` (database service)
- ~Add `user: "999:999"`.~
- Add `cap_drop: [ALL]` (consistent with the other hardened services).
- Ensure `/data/configdb` is writable under the read-only root filesystem
  (the image declares it a volume; confirm during testing whether an explicit
  tmpfs/empty mount is needed).

### 3. Helm (`helm/values.yaml` + `helm/templates/database.yaml`)
- `database.securityContext`:
  - `capabilities.drop: [ALL]` (remove the added `CHOWN/DAC_OVERRIDE/SETGID/SETUID`)
  - keep `readOnlyRootFilesystem: true`
  - add `runAsUser: 999`, `runAsGroup: 999`
- Pod-level security context:
  - `fsGroup: 999` — required so a freshly provisioned (empty, root-owned)
    PVC is chowned to 999 on first mount.
  - `fsGroupChangePolicy: OnRootMismatch` — skips the recursive chown when the
    volume root already matches (existing 999-owned PVCs), avoiding slow
    re-chowns on large volumes.

## Migration impact for operators

- **Standard deployments: seamless.** Data is already 999-owned and `mongod`
  was already running as 999 (via `gosu`); only *how* it becomes 999 changes.
- **Caveat:** operators running the database with a custom non-999 user, or
  with a data volume hand-chowned to a different UID, must align it to UID/GID
  999.

## Verification before merge

1. **Compose, fresh:** `docker compose up` → DB initializes (root user
   created) and serves; full app smoke test (server connects, report loads).
2. **Compose, upgrade:** start the current image on an existing `dbdata`
   volume, then swap to the rebuilt non-root image → existing data is read.
3. **Helm:** `helm template`/lint; install on kind/minikube → confirm
   `fsGroup` chown on a fresh PVC and that dropped caps don't break startup.
4. Confirm `gosu` is absent from the built image and re-run Trivy locally →
   the 36 alerts drop off.

## Residual alerts after this change

These are upstream base-image issues, unaffected by this change (candidates to
dismiss or leave for routine dependency bumps):

- ~40 Ubuntu OS-package CVEs in the mongo base image (mostly low/medium; fixed
  on the next mongo digest bump).
- 1 bundled `js-yaml` 3.13.1 in `mongosh` (`CVE-2025-64718`, medium).
- 1 LOW `rand` crate inside Chromium's `librav1e` in the renderer image
  (`GHSA-cq8v-f236-94qc`).

fniessink:
Run the MongoDB with user 999 so that `gosu` can be removed from the database container image to reduce its attack surface.

Closes #13514.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -38,4 +38,48 @@ kubectl get pods | grep quality-time-www
 
 ## Customizing
 
-Note that the `securityContext` of each component in the Helm chart values contains `readOnlyRootFilesystem: true`; when overriding the `securityContext` of a component, include `readOnlyRootFilesystem: true` in the override to keep the read-only root filesystem, as overrides replace the whole `securityContext`.
+The chart sets a `securityContext` on each component (and, for the database, also a `podSecurityContext`), including `readOnlyRootFilesystem: true`. Helm deep-merges your overrides onto these defaults, so you only need to specify the keys you want to change; the other defaults stay in place. To *remove* a default key, set it to `null` in your override — omitting it leaves the default unchanged. Note that lists, such as `capabilities.drop`, are replaced rather than merged when you override them.
+
+### Running the database as a different user
+
+By default the database container runs as the `mongodb` user (UID 999). Some platforms (such as OpenShift) require containers to run with a UID >= 1000 or assign an arbitrary high UID. The database image supports this: its data directories (`/data/db` and `/data/configdb`) are group-owned by the root group (GID 0) and group-writable, so the container can run as any non-root user that belongs to GID 0.
+
+To run the database as a fixed UID >= 1000, override the user in the `securityContext` and set a matching `fsGroup` in the `podSecurityContext` so the platform makes the persistent volume writable for that user:
+
+```yaml
+database:
+  securityContext:
+    capabilities:
+      drop:
+        - ALL
+    readOnlyRootFilesystem: true
+    runAsUser: 1000 # any value >= 1000
+    runAsGroup: 0
+  podSecurityContext:
+    fsGroup: 1000
+    fsGroupChangePolicy: OnRootMismatch
+```
+
+### Running the database with an arbitrary UID (OpenShift)
+
+On platforms that assign an arbitrary UID (such as OpenShift with the default `restricted-v2` SCC), don't pin the UID or group at all: the platform picks the UID, runs it with GID 0, and the data directories are already group-writable for GID 0. Because the chart ships `runAsUser: 999`, `runAsGroup: 999`, `fsGroup: 999`, and `fsGroupChangePolicy` as defaults, and Helm keeps unspecified defaults, you have to *remove* all four by setting them to `null`:
+
+```yaml
+database:
+  securityContext:
+    capabilities:
+      drop:
+        - ALL
+    readOnlyRootFilesystem: true
+    runAsUser: null
+    runAsGroup: null
+  podSecurityContext:
+    fsGroup: null
+    fsGroupChangePolicy: null
+```
+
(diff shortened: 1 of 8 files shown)


helm/README.md
```diff
@@ -39,3 +39,27 @@ kubectl get pods | grep quality-time-www
...
+database:
+  securityContext:
+    capabilities:
+      drop:
+        - ALL
+    readOnlyRootFilesystem: true
+    runAsUser: 1000 # any value >= 1000
+    runAsGroup: 0
+  podSecurityContext:
+    fsGroup: 1000
```
wkoot, line 59: What does this do on a platform like openshift, and why is this one not GID 0 but the UID? The documentation seems to suggest that fsGroup is a supplementary group ID

helm/values.yaml
```diff
@@ -40,16 +40,19 @@ collector:
...
-        - SETGID
-        - SETUID
       drop:
         - ALL
     readOnlyRootFilesystem: true
+    runAsUser: 999
+    runAsGroup: 999
+  podSecurityContext:
+    fsGroup: 999
+    fsGroupChangePolicy: OnRootMismatch
```
wkoot, line 57: This seems unexpected with regards to `readOnlyRootFilesystem`

helm/README.md
fniessink: It shouldn't be used on openShift. I've extended the helm/README.md to explain this.

helm/values.yaml
fniessink: I've extended the helm/README.md to explain this.

wkoot (approved):

PR #14685, opened 2026-09-18
Show Dependency-Track components' parent(s)

### LINKED ISSUE #8687: Display which parent dependency the current dependency is part of in Dependency-Track
## Summary

Add a 'Parent component' column to the measurement details of the 'security warnings' and 'dependencies' metrics when
Dependency-Track is the source. The column shows the top-level (root) dependency, or dependencies, of the project
through which the component is pulled in. The column is empty when the project depends on the component directly.

## Motivation

Dependency-Track reports a vulnerability against the component that actually contains it, which is often a transitive
dependency. To act on the warning, the user needs to know which of *their own* dependencies pulls that component in.

From the issue: Dependency-Track reports a vulnerability in Netty and says an update is possible, but Netty is a
transitive dependency of Quarkus, and Quarkus has no release that picks up the newer Netty. Without the parent
component, the user cannot tell that the warning is not actionable by bumping Netty.

## Scope

In scope:

- A new 'Parent component' entity attribute for the `security_warnings` entity of the Dependency-Track source.
- The same attribute for the `dependencies` entity, because the same metric shows outdated transitive dependencies and
  the data needed is already retrieved there (see [Collector changes](#collector-changes)).

Out of scope (see [Follow-up](#follow-up)):

- A parameter to filter entities by direct versus transitive dependency, as requested in the second issue comment.
- Showing the full dependency path (root → … → component) rather than only the root.
- Showing the parent component's version, latest version, or update status.

## Background: how Dependency-Track models the dependency graph

Based on the Dependency-Track sources (`master`, September 2026):

- `Project` and `Component` each have a `DIRECT_DEPENDENCIES` column that holds a JSON **string**, serialized as the
  `directDependencies` JSON property. It is `null` when the node has no direct dependencies.
  (`src/main/java/org/dependencytrack/model/{Project,Component}.java`)
- The string contains an array of component identities, written by
  `BomUploadProcessingTask.resolveDirectDependenciesJson()` from `ComponentIdentity.toJSON()`:

  ```json
  [
    {
      "uuid": "b2b1a5c2-...",
      "group": "io.quarkus",
      "name": "quarkus-netty",
      "version": "3.2.0",
      "purl": "pkg:maven/io.quarkus/quarkus-netty@3.2.0",
      "purlCoordinates": "pkg:maven/io.quarkus/quarkus-netty@3.2.0",
      "cpe": null,
      "swidTagId": null,
      "objectType": "COMPONENT"
    }
  ]
  ```

  Note that `objectType` can also be `SERVICE`; Dependency-Track does not store `directDependencies` for services.
- The project's `directDependencies` are the roots of the graph. The graph is edge-list-per-parent: to find the
  parents of a component you have to invert the relation. Dependency-Track itself does this with a `LIKE` query over
  `DIRECT_DEPENDENCIES` (`ComponentQueryManager.getParentDependenciesOfComponent()`).
- `GET /api/v1/component/project/{uuid}` returns the components of a project, each serialized with their
  `directDependencies`. This is the endpoint the design builds on.
- The findings API (`GET /api/v1/finding/project/{uuid}`, already used by the collector) returns no graph
  information: `Finding`'s component map is limited to `uuid`, `name`, `group`, `version`, `scope`, `purl`, `cpe` and
  `project`.

### Data model changes

In `components/shared_code/src/shared_data_model/sources/dependency_track.py`, add:

```python
PARENT_COMPONENT_ATTRIBUTE = EntityAttribute(
    name="Parent component",
    url="parent_component_landing_url",
    help="The direct dependency of the project through which this component is included. Empty if the project depends "
    "on this component directly. If the component is included through more than one dependency, all of them are "
    "listed.",
)
```

and insert it directly after the existing `EntityAttribute(name="Component", url="component_landing_url")` in both the
`dependencies` and the `security_warnings` entity. The generated attribute key is `parent_component`.

Rendering follows the generic entity table (`components/frontend/src/source/SourceEntityAttribute.jsx`); no frontend
change is needed. Note that `SourceEntityAttribute` supports at most one URL per attribute, which is why the
collector only sets `parent_component_landing_url` when there is exactly one parent (see below).

### Collector changes

Add a small graph helper to `components/collector/src/source_collectors/dependency_track/base.py`:

```python
class DependencyTrackComponentGraph:
    """The dependency graph of the components of one Dependency-Track project."""

    def __init__(self, components: list[DependencyTrackComponent]) -> None:
        self.__names = {component["uuid"]: component["name"] for component in components}
        self.__parents: dict[str, set[str]] = defaultdict(set)
        for component in components:
            for child in json.loads(component.get("directDependencies") or "[]"):
                self.__parents[child["uuid"]].add(component["uuid"])

    def root_components(self, uuid: str) -> dict[str, str]:
        """Return the UUIDs and names of the project's direct dependencies that include the component, transitively."""
        roots: set[str] = set()
        visited: set[str] = set()
        queue = [uuid]
        while queue:
            current = queue.pop()
            if current in visited:
                continue  # Guard against cycles in the graph
            visited.add(current)
            if parents := self.__parents.get(current):
                queue.extend(parents)
            elif current != uuid:  # A component without parents is a direct dependency of the project
                roots.add(current)
        # Ignore UUIDs that do not resolve to a component; those are services, which have no dependency graph
        named_roots = {root: self.__names[root] for root in roots if root in self.__names}
        return dict(sorted(named_roots.items(), key=itemgetter(1)))
```

One method returns both the UUIDs and the names so that the graph is walked once per entity, and so that the
column text and the link are derived from the same set of resolvable roots.

Roots are defined as *components without parents* rather than as the entries of the project's `directDependencies`.
Both are the same for a well-formed BOM, and the former needs no project-level data and stays correct when the
project's `directDependencies` is `null`.

Both collectors then set two extra entity fields:

```python
roots = graph.root_components(component_uuid)
entity["parent_component"] = ", ".join(roots.values())
if len(roots) == 1:
    entity["parent_component_landing_url"] = f"{landing_url}/components/{next(iter(roots))}"
```

`components/collector/src/source_collectors/dependency_track/security_warnings.py`:

- In `_get_source_responses()`, after `self._projects = await self._get_projects_by_uuid()`, retrieve
  `GET {api_url}/component/project/{uuid}` for every project and build one `DependencyTrackComponentGraph` per
  project, stored in `self._graphs: dict[str, DependencyTrackComponentGraph]`. Reuse
  `DependencyTrackBase._get_source_responses()` so the existing pagination applies. Keep these requests out of the
  returned `SourceResponses` so `_parse_entities()` keeps seeing only finding responses.
- In `_create_entity()`, look up the graph via `component["project"]` and add the two fields.

`components/collector/src/source_collectors/dependency_track/dependencies.py`:

- This collector already retrieves exactly the components needed, so it needs **no extra requests**. Restructure
  `_parse_entities()` to first collect the components of all response pages, group them by
  `component["project"]["uuid"]`, build one graph per project, and only then create the entities.
- Move the `DependencyTrackComponent` TypedDict to `json_types.py` and add
  `directDependencies: NotRequired[str]`, so `security_warnings.py` and `base.py` can share it. The findings API
  returns a *different* component shape (with `project` as a UUID string rather than an object), so the TypedDict
  that `security_warnings.py` currently also calls `DependencyTrackComponent` has to stay separate. Rename it to
  `DependencyTrackFindingComponent` to keep the two apart, and update the import in `test_security_warnings.py`.

### Edge cases

| Case | Behavior |
| --- | --- |
| Project depends on the component directly | Column empty, no link |
| Component is reachable through several roots (diamond) | All root names, comma-separated and sorted; no link |
| BOM without dependency graph (all `directDependencies` are `null`) | Every component looks like a root, so the column is empty everywhere — degrades silently, which is the honest answer |
| Cycle in the graph | `visited` set terminates the walk; roots reachable from the cycle are still reported |
| Root UUID refers to a service | Skipped, because Dependency-Track stores no graph for services, so it appears in neither the column nor the link |
| Component UUID from a finding is not in the component list | `root_components()` returns an empty dict, column empty |

## Impact on existing behavior

- **Measurement values do not change.** The new attribute is informational; `_include_entity()` is untouched.
- **Entity keys do not change** (`finding["matrix"]` for security warnings, `component["uuid"]` for dependencies), so
  existing entity statuses such as 'false positive' and 'won't fix' are preserved.
- **Old measurements** stored before the change have no `parent_component` field; but neither has the data model, so the
  frontend does not render the column.
- **Requests.** The dependencies metric issues no extra requests. The security warnings metric issues one extra
  paginated component request per project: `ceil(number_of_components / 100)` requests, on top of the existing
  finding requests. For a project with 2000 components that is 20 extra requests. This is the same volume the
  dependencies metric already issues against the same instances, so it is considered acceptable; if it turns out not
  to be, switch to [alternative A](#alternatives-considered).
- **Dependency-Track version.** The `directDependencies` field exists since the dependency graph was introduced in
  Dependency-Track v4.5. No new API permissions are needed: `view_portfolio` already covers
  `/api/v1/component/project/{uuid}`.

## Alternatives considered

**A. `GET /api/v1/component/project/{projectUuid}/dependencyGraph/{componentUuids}`.** Dependency-Track's own UI uses
this to expand the graph to a component. It takes `|`-separated component UUIDs and returns a map of UUID to a
trimmed component that carries a `dependencyGraph` set of child UUIDs, covering the paths from the project roots to
each requested component plus all project roots. Roots would be the returned nodes that are nobody's child.

Rejected as the primary approach because it needs batching logic to keep URLs within length limits (a UUID plus
separator is 37 characters), the response is not actually small (it always includes every project root and its direct
children), and the endpoint is less widely used than the plain component API. It remains the right fallback if the
extra requests for the security warnings metric prove too expensive.

**B. Walking `GET /api/v1/dependencyGraph/component/{uuid}/directDependencies` from the project roots downwards.**
Rejected: one request per graph node, which is an order of magnitude more requests than either other approach.

**C. Showing the direct parent rather than the root ancestor.** Rejected: the issue explicitly asks for the
"(root/ancestor) parent", and the direct parent of a deeply nested component is usually just as un-actionable as the
component itself.

## Testing

Unit tests in `components/collector/tests/source_collectors/dependency_track/`:

- `test_dependencies.py` and `test_security_warnings.py`: extend the fixtures with a `directDependencies` JSON string
  and assert the new entity fields. Cover: direct dependency (empty column), one transitive parent (name and link),
  a component reachable through two roots (two names, no link), a multi-level chain (the root is reported, not the
  intermediate parent), a cycle, a `null`/missing `directDependencies`, and a root with `objectType` `SERVICE`
  (neither a name nor a link, so that the column does not end up with a link on empty text).
- `test_security_warnings.py` currently drives the mocks with an ordered
  `get_request_json_side_effect=[projects, findings]`. The extra component requests change that order, so the
  `collect_findings()` helper in the test case needs a `components` argument and the ordering has to match the order
  in which `_get_source_responses()` issues the requests. This touches every test in the file.
- Add the component fixture to `base_test.py` so both test modules share it.

Run with `cd components/collector && just test`, and the data model tests with `cd components/shared_code && just test`.

## Documentation

- `docs/src/reference.md` is generated from the data model by `docs/src/create_reference_md.py`; regenerate it rather
  than editing it.
- Add an entry to the `### Added` section of `## [Unreleased]` in `docs/src/changelog.md`, in the house style, for
  example:

  > When measuring security warnings or dependencies with Dependency-Track as source, also show the parent component
  > in the measurement details, that is, the dependency of the project through which the component is included.
  > Closes [#8687](https://github.com/ICTU/quality-time/issues/8687).

## Implementation checklist

1. Move `DependencyTrackComponent` to `json_types.py`, add `directDependencies`, and rename the findings-shaped
   TypedDict in `security_warnings.py` to `DependencyTrackFindingComponent`.
2. Add `DependencyTrackComponentGraph` to `base.py`.
3. Add the `Parent component` attribute to both entities in the data model.
4. Wire up `dependencies.py` (group responses by project, build graph, set fields).
5. Wire up `security_warnings.py` (fetch components per project, build graph, set fields).
6. Update and extend the unit tests.
7. Regenerate the reference documentation and add the changelog entry.

## Follow-up

The second comment on the issue also asks to *filter* on whether a dependency is transitive. That is a separate,
larger change: it needs a new source parameter (for example 'Dependency types' with values `direct` and `transitive`)
and an extension of `DependencyTrackLatestVersionStatusBase._include_entity()`, and unlike the parent column it
*does* change measurement values. Worth splitting into its own issue once the parent column is in place, since the
graph data it needs is exactly what this change already retrieves.

fniessink:
When measuring 'security warnings' or 'dependencies' with Dependency-Track as source, the measurement details show a new 'Parent component' column with the dependency of the project through which the component is included. The column is empty when the project depends on the component directly.

Closes #8687.


components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -22,13 +56,25 @@ class DependencyTrackBase(TokenAuthenticationSourceCollector):
...
+        """Return the landing URL of the component with the specified UUID."""
+        return URL(f"{self.__landing_url()}/components/{uuid}")
+
+    def _landing_url_of_project(self, uuid: str) -> URL:
+        """Return the landing URL of the project with the specified UUID."""
+        return URL(f"{self.__landing_url()}/projects/{uuid}")
+
+    def __landing_url(self) -> str:
+        """Return the landing URL of the Dependency-Track instance, without trailing slash."""
+        return str(self._parameter("landing_url")).strip("/")
```
wkoot, line 76: Why does `URL` not handle duplicate slashes when concatenating url paths, i.e. like `pathlib.Path`?

components/collector/src/source_collectors/dependency_track/base.py
```diff
@@ -109,6 +155,14 @@ def _is_latest(project: DependencyTrackProject) -> bool:
 class DependencyTrackLatestVersionStatusBase(DependencyTrackBase):
     """Base class for Dependency-Track collectors that can be filtered by latest version status."""
 
+    def _parent_component_attributes(self, uuid: str, graph: DependencyTrackComponentGraph) -> dict[str, str]:
+        """Return the parent component entity attributes for the component with the specified UUID."""
+        roots = graph.root_components(uuid)
+        attributes = {"parent_component": ", ".join(roots.values())}
+        if len(roots) == 1:  # Entity attributes support at most one URL, so only link if there's exactly one parent
```
wkoot, line 162: So if there is more than one root, this functionality is disabled? Is that clear in the UI?

components/collector/src/source_collectors/dependency_track/security_warnings.py
```diff
@@ -44,9 +44,21 @@ async def _get_source_responses(self, *urls: URL) -> SourceResponses:
         """Extend to get the findings."""
         api_url = await self._api_url()
         self._projects = await self._get_projects_by_uuid()
+        # The findings API returns no dependency graph information, so retrieve the components of each project to
+        # be able to determine the parent components of the vulnerable components
+        self._graphs = {uuid: await self._get_component_graph(uuid) for uuid in self._projects}
```
wkoot, line 49: This seems a bit risky regarding memory consumption; do we really need the full graph?

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
```diff
@@ -59,23 +64,27 @@ def entities(self, description: str = "description", severity: str = "Unassigned
...
         projects: list[DependencyTrackProject] | None = None,
         findings: list[DependencyTrackFinding] | None = None,
+        components: list[DependencyTrackComponent] | None = None,
     ) -> MetricMeasurement:
-        """Collect a measurement for the given projects and findings, defaulting to the fixtures."""
+        """Collect a measurement for the given projects, components, and findings, defaulting to the fixtures."""
         return await self.collect_measurement(
             get_request_json_side_effect=[
                 self.projects() if projects is None else projects,
+                [self.component()] if components is None else components,
```
wkoot, line 87: This seems unexpected, but is it just a shorthand for testing?

components/shared_code/src/shared_data_model/sources/dependency_track.py
```diff
@@ -25,6 +25,16 @@
     EntityAttribute(name="Project", url="project_landing_url"),
     EntityAttribute(name="Project version"),
 ]
+COMPONENT_ATTRIBUTES = [
+    EntityAttribute(name="Component", url="component_landing_url"),
+    EntityAttribute(
+        name="Parent component",
+        url="parent_component_landing_url",
+        help="The direct dependency of the project through which this component is included. Empty if the project "
```
wkoot, line 33: Perhaps it is helpful to talk about a "top-level" or "root" dependency

components/collector/src/source_collectors/dependency_track/base.py
fniessink: Because `URL` is just a `str`: `URL = NewType("URL", str)`. To make it replace any repeated slashes it would have to become a class. Worth contemplating, but it would touch unrelated code. There's 11 places where we do `strip("/")` and 49 where `"/"` occurs, and not all of them are about URLs.

components/collector/src/source_collectors/dependency_track/base.py
fniessink: It's clear in the sense that if a dependency has multiple roots, the roots won't have URLs. Nothing breaks.It would be better if all roots could have their own URL of course. I'll open an issue for that.

components/collector/src/source_collectors/dependency_track/security_warnings.py
fniessink: It worked for the Eco SBoM and that's a big one.

components/collector/tests/source_collectors/dependency_track/test_security_warnings.py
fniessink: It's the step where the components are retrieved to be able to look up the root parents (see your previous comment).

components/shared_code/src/shared_data_model/sources/dependency_track.py
fniessink: Root component is better, yes. Applying.

